PR #78, opened 2026-05-07
docs: add installation guides, runtime troubleshooting, feature overview, and dep-check script

jtomek-strike48:
> [!NOTE]
> **The failing `Docker Build` check is NOT caused by this PR.** It's a pre-existing CI infrastructure issue where `Dockerfile:3` runs `cargo install cargo-chef` unlocked, so Cargo resolves the newly-published `cargo-platform@0.3.3` which requires rustc 1.91 — but the base image pins `FROM rust:1.88-bookworm`. Cargo itself prints the recommended fix: *"Try re-running `cargo install` with `--locked`"*. That one-line fix is in #79. Once #79 merges, re-running Docker Build on this PR will go green. All 17 other checks pass, including every Rust check (Format, Clippy, Check, Test, Build amd64/arm64, Check macOS, Analyze rust) and every security/compliance check (Gitleaks, Dependency Audit, SBOM, CodeQL).

## Summary

Adds the docs index and companion guides for installing and running Pick, plus a standalone dependency validator script that pairs with `install.sh` for users who skip the installer.

- `check-deps.sh` (executable) — prerequisite validator, bash with `set -euo pipefail`.
- `docs/README.md` — documentation index.
- `docs/INSTALLATION.md` — full installation guide (410 lines).
- `docs/QUICK_INSTALL.md` — 5-minute quickstart (88 lines).
- `docs/RUNTIME_TROUBLESHOOTING.md` — common runtime errors and fixes (331 lines).
- `docs/PICK_CUSTOMER_FEATURES.md` — feature overview (462 lines).

## Notes

- These files were created in earlier sessions but never staged. No code changes, no behavior changes.
- Content scanned for tenant/customer names and internal URLs — none found.
- `COMPETITIVE_ANALYSIS_TRIAGE.md` in the working tree is intentionally **not** included; its own header marks it `DO NOT COMMIT` per the repo's Agent Notes Policy.

## Test plan

- [ ] (reviewer) Read `docs/README.md` and confirm the document linking is coherent.
- [ ] (reviewer) `bash check-deps.sh` on a clean machine and verify the validator runs.
- [ ] (reviewer) Confirm `docs/PICK_CUSTOMER_FEATURES.md` is appropriate for a public repo (it's customer-facing, not internal-only).

PR #77, opened 2026-05-07
fix(config): preserve URL scheme in normalize_host so desktop auto-detects WebSocket transport

jtomek-strike48:
> [!NOTE]
> **The failing `Docker Build` check is NOT caused by this PR.** It's a pre-existing CI infrastructure issue where `Dockerfile:3` runs `cargo install cargo-chef` unlocked, so Cargo resolves the newly-published `cargo-platform@0.3.3` which requires rustc 1.91 — but the base image pins `FROM rust:1.88-bookworm`. Cargo itself prints the recommended fix: *"Try re-running `cargo install` with `--locked`"*. That one-line fix is in #79. Once #79 merges, re-running Docker Build on this PR will go green. All 17 other checks pass.

## Summary

- `normalize_host` previously stripped the URL scheme. Headless apps read `STRIKE48_HOST` directly and the scheme survived, but the desktop UI routed user input through `normalize_host` and lost it. Without a scheme, the SDK's `parse_url` fell back to **gRPC**.
- Strike48 production is Cloudflare-fronted on `:443`. Cloudflare terminates TLS cleanly but does not proxy arbitrary gRPC/HTTP2, so tonic surfaces the closed stream as `TLS connection failed: transport error` — after a successful TLS 1.3 handshake. That is why the symptom looked like a TLS bug but headless worked with identical credentials.
- `normalize_host` now preserves any provided scheme and, when no scheme is given and the host ends with `:443`, infers `wss://` (the Cloudflare convention).
- `load_settings` re-runs `normalize_host` on persisted `last_config.host` so existing users with a bare hostname saved by older versions get migrated transparently on first launch.
- The `ConfigForm` placeholder is updated from `grpc://localhost:50061` to `wss://strike48.example.com:443` so the correct scheme is discoverable.

## Files changed

- `crates/core/src/config.rs` — preserve scheme; infer `wss://` for bare `:443` hosts; 8 new unit tests (preservation, inference, case-insensitivity, trimming, error cases).
- `crates/core/src/settings.rs` — re-run `normalize_host` on persisted host and atomically save when the value changes.
- `crates/ui/src/components/config_form.rs` — updated placeholder.

## Test plan

- [x] `cargo test -p pentest-core --lib config::` → 8/8 pass.
- [x] `cargo clippy -p pentest-core --all-targets --tests -- -D warnings` → clean.
- [x] `cargo check -p pentest-desktop` → clean.
- [x] `cargo check -p pentest-headless` → clean.
[...]

PR #73, opened 2026-05-07
feat: add macOS and Windows code signing to release workflow

taylor-wood-devo:
## Summary
- Add macOS code signing (Developer ID Application) with hardened runtime to all macOS binaries
- Add Apple notarization via `xcrun notarytool` so Gatekeeper accepts downloads without warnings
- Add Windows code signing via Azure Trusted Signing so SmartScreen accepts downloads without warnings
- Add `entitlements.plist` for WebView (unsigned executable memory) and network client access
- Applied to `pentest-connector` (desktop) and `pentest-agent` (headless) across macOS (x86_64 + aarch64) and Windows (x86_64)

### Org-level (already configured)
- `APPLE_CERTIFICATE_P12`, `APPLE_CERTIFICATE_PASSWORD`, `APPLE_TEAM_ID`, `APPLE_ID`
- `AZURE_TENANT_ID`, `AZURE_CLIENT_ID`, `AZURE_CLIENT_SECRET`, `AZURE_SIGNING_ENDPOINT`, `AZURE_SIGNING_ACCOUNT`, `AZURE_CERT_PROFILE`

### Repo-level (pick)
| Secret | Description |
|--------|-------------|
| `APPLE_APP_PASSWORD` | App-specific password for notarization |

## Test plan
- [ ] Verify `APPLE_APP_PASSWORD` is set as a repo secret on pick
- [ ] Tag a release and confirm macOS signing/notarization passes
- [ ] Tag a release and confirm Windows Azure Trusted Signing passes
- [ ] Download macOS binary and verify no Gatekeeper warning
- [ ] Download Windows binary and verify signature in Properties > Digital Signatures

PR #79, opened 2026-05-07
fix(ci): use --locked when installing cargo-chef to unblock Docker Build

jtomek-strike48:
## Summary

\`Dockerfile:3\` ran \`cargo install cargo-chef\` unlocked. Cargo then resolved the newest compatible version of every transitive dependency at build time. \`cargo-platform\` 0.3.3 was recently published requiring rustc 1.91, but the base image pins \`FROM rust:1.88-bookworm\`. Every Docker Build job started failing with:

\`\`\`
cargo-platform@0.3.3 requires rustc 1.91
rustc 1.88.0 is not supported
\`\`\`

Adding \`--locked\` pins cargo-chef's own lockfile so the install is reproducible against the pinned toolchain. This mirrors what \`Dockerfile.scratch:16\` and \`Dockerfile.scratch:41\` already do on the same command.

## Files changed

- \`Dockerfile\`: one-line change on line 3.

## Impact

Unblocks Docker Build on every open PR (including #77 and #78) once merged.

## Test plan

- [ ] (reviewer) CI's Docker Build job should go green on this PR.
- [ ] (reviewer) After merge, rebase/re-run #77 and #78 so they pick up the fix.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,6 @@
 # ── Stage 1: Chef ──────────────────────────────────────────────
 FROM rust:1.88-bookworm AS chef
-RUN cargo install cargo-chef
+RUN cargo install cargo-chef --locked
 WORKDIR /app
 
 # ── Stage 2: Planner ──────────────────────────────────────────


taylor-wood-devo:
for the last commit with `--locked` on this branch, I hit this in https://github.com/Strike48-public/pick/pull/73 and bumped the default image vs locking deps
- if the image bump is fine I think pursue that, but if we fight more issues with new image we can just `--locked` and use this PR and i'll rebase #73 if that works with you

jtomek-strike48:
Scope note for reviewers: this PR is intentionally the minimum change to unblock CI (exactly one line, matching what `Dockerfile.scratch` already does on the same command). The broader cleanup identified during review — bumping the Rust base image off 1.88 and pinning cargo-chef to a specific version — is tracked as a follow-up in #80 so it doesn't get lost.

All 17 required checks are green, including the previously-failing Docker Build itself (9m16s pass). Once this merges, re-running Docker Build on #77 and #78 will clear their only failing check.

jtomek-strike48:
Agreed — Plan A is the better call. Your base image bump to `rust:1.91-bookworm` in #73 fixes today's Docker Build failure *and* closes half of the follow-up I filed in #80 (item 1: "bump Rust base image"), so it's strictly more valuable than my `--locked` workaround.

Closing #79 as superseded. Once #73 merges, #77 and #78 will pick up the fix on re-run. I'll also update #80 to drop item 1 (the base-image bump) since #73 handles it, and keep item 2 (pin cargo-chef version) as the remaining follow-up.

If something regresses with the newer compiler post-merge and we need the `--locked` fallback, happy to open a fresh one-liner.

jtomek-strike48:
Superseded by #73 (base image bump to rust:1.91-bookworm fixes the same issue and closes half of #80).

PR #39, opened 2026-03-28
feat: CyberChef integration with drag-drop UI and JWT auto-cleanup

jtomek-strike48:
## Summary

Implements comprehensive CyberChef integration with interactive UI and automatic JWT token expiration handling.

### CyberChef Backend (Commit 1)
- New `pentest-cyberchef` crate with tool and recipe library
- 20 pre-built recipes for common pentest operations:
  - Encoding/decoding: base64, hex, URL
  - Hashing: MD5, SHA-1, SHA-256
  - Cryptography: XOR brute force, ROT13
  - Data extraction: URLs, IPs, emails, domains
  - Compression: gzip, zlib
  - Analysis: magic detection, JWT decode
- Rust-native executor for fast, dependency-free execution
- Full test coverage (14 tests passing)

### CyberChef UI (Commit 2)
- Full drag-and-drop interface for building recipe chains
- Slot-based drop zones for smooth, flicker-free drag experience
- Enable/disable operations with visual toggle
- Auto Bake mode for real-time execution
- Resizable input/output panels
- Recipe chain execution with proper error handling
- Comprehensive unit tests (18 tests) and E2E tests (21 tests)
- Added to sidebar navigation

### JWT Token Auto-Cleanup (Commit 2)
- Validates JWT tokens on settings load
- Automatically clears expired tokens (5 min expiry + 30s buffer)
- Prevents "Registration failed" errors on restart
- Falls back to OTT registration when token expired
- No manual intervention required

### Files Changed
- **Backend:** 11 files, 963 insertions (new cyberchef crate)
- **UI & Auth:** 14 files, 2280 insertions (CyberChef UI + JWT validation)
- **Docs:** 3 new documentation files (testing guide, E2E setup, token fix)

## Test Plan

- [ ] Verify CyberChef appears in sidebar navigation
- [ ] Test drag-and-drop operations from left panel to recipe chain
- [ ] Test reordering operations within the recipe chain
- [ ] Test enable/disable toggle on operations
- [ ] Test remove operations (drag out or click ×)
- [ ] Test Auto Bake mode (ON: real-time execution, OFF: manual Bake button)
- [ ] Test recipe execution with multiple chained operations
- [ ] Test input/output panel resizing
- [ ] Verify JWT token auto-cleanup on restart (check logs for "Cleared expired/invalid auth token")
- [ ] Run unit tests: `cargo test cyberchef_sortable_tests`
- [ ] Run E2E tests: `cd e2e && npm test` (if Playwright configured)

[...]

PR #97, opened 2026-05-15
ci: disable macOS code signing by default

taylor-wood-devo:
## Summary

- Adds a `macos_signing` boolean input (default: **false**) to the existing `workflow_dispatch` trigger
- Gates all Apple certificate import, `codesign`, and `notarytool` notarization steps behind `inputs.macos_signing == true`
- macOS builds for both x86_64 and aarch64 still produce unsigned artifacts (desktop + headless)
- Windows and Linux builds are unaffected

## Context

The macOS developer account is temporarily suspended, causing all signing/notarization steps to fail on release runs.

## What's gated behind the flag

| Step | `build-desktop-macos` | `build-headless-macos` |
|------|----------------------|----------------------|
| Import Apple certificate | gated | gated |
| `codesign` binary | gated | gated |
| `notarytool` submit + wait | gated | gated |

## Re-enabling signing

When the account is restored, either:
1. Run the release workflow via **workflow_dispatch** with `macos_signing: true`
2. Flip the default back to `true` in the workflow file

## Test plan

- [ ] Trigger a workflow_dispatch release with `macos_signing: false` — verify macOS builds succeed without signing
- [ ] Verify unsigned macOS artifacts are produced for both architectures (desktop + headless)
- [ ] Verify Windows/Linux builds are unaffected

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -9,6 +9,11 @@ on:
       version:
         description: 'Version to release (e.g., 0.1.0)'
         required: true
+      macos_signing:
+        description: 'Enable macOS code signing & notarization (requires active Apple developer account)'
+        required: true
+        default: false
+        type: boolean
 
 permissions:
   contents: read
@@ -109,6 +114,7 @@ jobs:
         run: cargo build --package pentest-desktop --release --target ${{ matrix.target }} --features pentest-platform/desktop-pcap
 
       - name: Import Apple certificate
+        if: inputs.macos_signing == true
         env:
           APPLE_CERTIFICATE_P12: ${{ secrets.APPLE_CERTIFICATE_P12 }}
           APPLE_CERTIFICATE_PASSWORD: ${{ secrets.APPLE_CERTIFICATE_PASSWORD }}
@@ -125,6 +131,7 @@ jobs:
           security list-keychains -d user -s $KEYCHAIN_PATH login.keychain-db
 
       - name: Sign macOS binary
+        if: inputs.macos_signing == true
         env:
           APPLE_TEAM_ID: ${{ secrets.APPLE_TEAM_ID }}
         run: |
@@ -134,6 +141,7 @@ jobs:
             target/${{ matrix.target }}/release/pentest-connector
 
       - name: Notarize macOS binary
+        if: inputs.macos_signing == true
         env:
           APPLE_ID: ${{ secrets.APPLE_ID }}
           APPLE_APP_PASSWORD: ${{ secrets.APPLE_APP_PASSWORD }}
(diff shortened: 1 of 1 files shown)


jtomek-strike48 (approved):
## ✓ APPROVED

**Status:** Safe to merge - no critical issues found.

### Summary

This PR safely gates macOS signing steps behind a `macos_signing` workflow input (default: false), unblocking releases during the Apple developer account suspension. The implementation is clean and correct.

### Why this is safer than KubeStudio PR #39

Unlike the KubeStudio version, Pick's release workflow uses **workflow_dispatch only** (no tag push trigger). This means `inputs.macos_signing` is **always defined**, avoiding the undefined/null handling edge case found in KubeStudio PR #39.

### Current status

- **CI:** 19/20 checks passing (1 skipped is expected)
- **Changes:** Minimal and focused (11 additions, 0 deletions)
- **Impact:** macOS builds remain functional (unsigned)
- **Reversibility:** Easy to re-enable signing

### Optional improvements

I'm leaving some non-blocking suggestions in separate comments that would improve robustness and maintainability. These are the same recommendations I made for KubeStudio PR #39 - feel free to apply them here if you'd like, but they're not required for merge.

**Verdict:** Approve and merge when ready.

jtomek-strike48:
## Optional: Consider using truthiness instead of `== true`

**Severity:** LOW (defensive programming, not critical for workflow_dispatch-only triggers)

While `inputs.macos_signing == true` works correctly here (since workflow_dispatch always defines inputs), using truthiness is more idiomatic for boolean conditionals:

```yaml
# Current (works fine)
if: inputs.macos_signing == true

# Slightly more idiomatic
if: inputs.macos_signing
```

**Why this doesn't matter as much for Pick:**

Unlike KubeStudio PR #39 (which has tag push triggers where inputs are undefined), Pick's workflow only uses `workflow_dispatch`, so `inputs.macos_signing` is always explicitly `true` or `false`, never `null`.

**Recommendation:** Optional - only change if you prefer the shorter syntax.

jtomek-strike48:
## Optional: Add pre-flight secret validation

**Severity:** LOW (nice to have)

When `macos_signing: true` is set but required secrets are missing or expired, the certificate import step will fail with cryptic errors. Consider adding validation before attempting to use secrets.

See my comment on [KubeStudio PR #39](https://github.com/Strike48-public/kubestudio/pull/39#issuecomment-4463292325) for a complete pre-flight validation step that:
- Checks all 5 required secrets before building
- Provides clear error messages about which secrets are missing
- Fails fast (saves 20+ minutes of wasted build time)

**Recommended location:** Before "Import Apple certificate" in both `build-desktop-macos` and `build-headless-macos` jobs.

**Benefit:** Operators know exactly what to fix instead of debugging cryptic keychain errors.

jtomek-strike48:
## Optional: Document temporary nature with inline comment

**Severity:** LOW (maintainability)

The `default: false` setting is a temporary workaround, but nothing in the workflow file indicates this. Future maintainers may wonder why signing is disabled by default.

**Suggested addition:**

```yaml
# TEMPORARY: macOS code signing disabled by default during Apple developer account suspension.
# TODO: Change default to true once account is restored (estimated: Q2 2026).
# Impact: Unsigned macOS builds trigger Gatekeeper warnings but remain functional.
macos_signing:
  description: 'Enable macOS code signing & notarization (requires active Apple developer account)'
  required: true
  default: false  # TEMPORARY - change to true when account restored
  type: boolean
```

See my detailed comment on [KubeStudio PR #39](https://github.com/Strike48-public/kubestudio/pull/39#issuecomment-4463293093) for the full rationale.

**Benefit:** Makes it clear this isn't the intended long-term state and provides timeline for restoration.

PR #83, opened 2026-05-07
fix(desktop): don't auto-open browser on connect, show Pending Approval callout

jtomek-strike48:
> [!CAUTION]
> **Do not merge — blocked by #85.** Post-merge testing revealed this PR regresses the Chat panel and Device Info quick action on desktop: after disabling browser-OAuth by default, the desktop webview has no way to obtain the Matrix session token required for GraphQL calls, so ChatPanel can't list agents and nothing gets auto-selected. Headless is unaffected. Needs investigation into alternative token-acquisition paths before shipping. See #85 for full analysis.

> [!IMPORTANT]
> **This is a stacked PR on top of #77.** Please review and merge #77 first. GitHub will auto-retarget this PR to `main` when #77 merges. The diff below shows only the changes on top of #77.

## Summary

Two linked UX problems made post-registration approval on the desktop confusing or invisible:

1. Clicking **Connect** opened a system browser tab on every attempt. That tab was trying to pre-fetch a Matrix chat token (separate from the connector JWT), but the path is flaky (tracked in #81) and the chat token isn't required for the connector to function. **The browser now stays closed by default**, gated behind `PICK_ENABLE_BROWSER_AUTH=1` for anyone who still needs it. StrikeHub mode's IPC auth path is unchanged.

2. The desktop flipped to the **Dashboard as soon as the wss handshake succeeded** — even though the server response carried no JWT (authorization was still pending admin approval). Users saw a non-functional Dashboard with "Pending approval" buried in the Recent Activity feed. **The status now stays `Connecting` until a JWT actually exists**, and the connecting screen renders a **bordered amber callout card** telling the user exactly what to do.

## What the callout looks like

```
┏━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┓
┃  [shield icon]                          ┃
┃  Pending approval                       ┃
┃                                         ┃
┃  Open Prospector Studio and approve     ┃
┃  this connector to continue.            ┃
┗━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┛
```

- Icon in a circular amber bubble (uses the existing `--warning` token — warm oklch amber that's already defined across all 8 themes)
- Title in warning-amber, hint in muted-foreground
- Gentle outward box-shadow pulse (2.4s period)

## Files changed

- `crates/ui/src/components/connecting_screen.rs` — replace single text line with structured callout; rename `Awaiting approval` -> `Pending approval`.
[...]

Issue #81, opened 2026-05-07
Browser OAuth callback binds to random port when 4000 and 5173 are busy, breaking CORS


jtomek-strike48:
## Observed behavior

During an end-to-end test of PR #77, the browser-based OAuth login flow in `crates/core/src/matrix/auth.rs` emitted this warning:

```
[BROWSER_AUTH] Ports 4000 and 5173 busy, using random port (CORS may fail)
[BROWSER_AUTH] Callback server listening on port 34749
```

The callback server then timed out after 120s waiting for the access token:

```
[BROWSER_AUTH] Timeout waiting for token
browser OAuth failed: matrix: Login timed out — no token received within 120 seconds
```

## Why it happens

`crates/core/src/matrix/auth.rs` (around line 314) tries to bind to `127.0.0.1:4000`, then `127.0.0.1:5173`, then falls back to `127.0.0.1:0` (OS-assigned random port). The Matrix server's CORS allow-list only includes the two fixed ports, so when the OS-assigned port is used, the cross-origin `fetch` to `/auth/refresh` from the callback HTML is blocked by CORS and no token ever arrives.

The fallback path already logs a warning acknowledging this will fail, then proceeds anyway.

## Impact

- Deployments that authenticate via the pre-approval OTT (`STRIKE48_REGISTRATION_TOKEN`) are **not** affected — the browser flow is only a fallback and the OTT path succeeds before the timeout matters.
- Deployments **without** a registration token, where the user logs in via the browser, will silently hang for 120s and then error out every time ports 4000 and 5173 are occupied.
- Busy-port situations are common on developer workstations (dev servers, Vite, Vue CLI all default to these ports).

## Suggested fixes

Pick one or more:

1. **Add more ports to the fallback list** and to the Matrix server's CORS allow-list (e.g., 4001-4010, 5174-5183). Low code complexity, but requires a coordinated change on the Strike48 server side.
2. **Emit a clear user-facing error** when both 4000 and 5173 are busy, telling the user which processes to stop. Better UX than a silent 120s hang.
3. **Do not fall through to a random port**. If the fixed ports are busy, fail fast with actionable guidance.
4. **Use a system-browser-friendly custom URI scheme** (like the Android path already does) to avoid the fixed-port requirement entirely on desktop. Larger change.

Option 2 combined with option 3 is the lowest-risk immediate improvement; option 1 is the proper long-term fix.

## Reproduction

[...]

PR #108, opened 2026-05-18
fix: add lazy browser-OAuth to ChatPanel for desktop auth

### LINKED ISSUE #85: Desktop Chat/Device Info regression: ChatPanel has no way to get a Matrix session token when browser-auth is off
## Observed behavior

After PR #83 disabled the browser-OAuth fallback by default on desktop, clicking "Device Info" on the Dashboard (a quick action that opens the Chat panel with a pre-filled prompt) shows **"Select an agent to begin"** instead of auto-selecting the pentest-connector agent and sending the prompt. Chat panel is effectively non-functional on desktop unless the user opts back into browser-OAuth via `PICK_ENABLE_BROWSER_AUTH=1`.

Headless mode is not affected because in headless, the workspace UI is rendered via the LiveView server over Prospector Studio's proxy, which injects the session token via a `__st` query parameter on the WebSocket URL.

## Why this is happening (preliminary, needs verification)

The ChatPanel requires a **session-backed Matrix chat token** (for GraphQL calls to `/api/v1alpha`). This is separate from the connector JWT (which authenticates the gRPC/wss stream). Per the comment in `crates/ui/src/liveview_connector/mod.rs:258`:

> Only `MatrixTokenObtained` carries a session-backed token valid for GraphQL.
> `CredentialsUpdated` carries the connector JWT (gRPC only, no DB session).

The chat token is populated via `crate::session::set_auth_token(...)` in exactly two places in the codebase:

1. `crates/ui/src/liveview_connector/auth.rs:30` — set after browser-OAuth returns in `try_fetch_matrix_token_fallback`. **PR #83 gated this path behind `PICK_ENABLE_BROWSER_AUTH=1`, so it no longer fires by default.**
2. `crates/ui/src/liveview_connector/mod.rs:894` — set when `handle_ws_open` sees `?__st=<token>` on an incoming WebSocket URL. This only fires when Strike48's proxy forwards a LiveView WebSocket session, which is the headless / Prospector-Studio-hosted workflow, not the desktop webview workflow.

Net: with #83 merged as-is, the desktop webview has no path to obtain a chat token, so the ChatPanel can't list agents, so the pentest-connector agent is never auto-selected, so quick actions that open Chat with a pre-filled prompt look like they do nothing.

## Verification needed

Before fixing, three questions should be answered concretely:

1. **Does the desktop webview ever trigger `handle_ws_open` on the local liveview server?** If yes, does the query string include `__st`? The desktop app does `start_liveview_server(...)` on launch, but I don't know whether wry/webkit2gtk actually opens a WebSocket through that path.
2. **Is there any mechanism for the desktop webview to acquire a Matrix session token without opening the system browser?** e.g., could the connector JWT be exchanged for a Matrix session via some server-side endpoint? Could the chat client be adapted to use the connector JWT directly for GraphQL (probably not — the comment explicitly says the connector JWT is gRPC-only, no DB session)?
3. **If a browser tab is ultimately necessary, what is the least-intrusive trigger?** Candidates:
   - Open on first Chat tab click (lazy)
   - Open only when user explicitly signs in via a button in the Chat panel
   - Keep #83's default off, but change Device Info / Network Scan quick actions to NOT open Chat — use a direct-tool-invocation path instead

## Impact

- Device Info, Network Scan, WiFi Scan, and any other Dashboard quick action that seeds the Chat panel appear broken to users on desktop who haven't set `PICK_ENABLE_BROWSER_AUTH=1`.
- Direct use of the Chat panel itself (clicking the Chat tab, typing a question) also fails for the same reason.
- Tools invoked directly through the Tools page work fine — this is purely a Chat/Agent-driven workflow regression.

## Proposed scope

1. **Investigation first** (30-60 min): trace the two bullets above, confirm hypotheses, document what "desktop needs" vs. "headless needs" look like.
2. **Then** choose from:
   - (a) Lazy browser-OAuth: trigger `try_fetch_matrix_token_fallback` when the Chat tab is first visited (not on Connect).
   - (b) Reuse connector JWT for chat: if technically possible, avoid the dual-token problem entirely.
   - (c) Move quick actions off Chat: invoke tools directly, bypassing the agent-chat flow.
   - (d) Document the limitation: desktop users must set `PICK_ENABLE_BROWSER_AUTH=1` for chat/agent features. Least work, worst UX.

## Related

- Introduced by PR #83 (`fix/desktop-no-auto-browser-oauth`).
- Interacts with issue #81 (browser-OAuth path is still flaky even when enabled).
- ChatPanel auth path: `crates/ui/src/components/chat_panel/mod.rs:197` gates agent fetch on `effective_token` being non-empty.

## Blocks

- **PR #83 should not be merged until this is resolved** or explicitly accepted as a tradeoff.

## Recommended agent

**backend-architect** or **ui-builder** depending on which fix path is chosen. Investigation is code-reading-heavy; start with `code-explorer`.

jtomek-strike48:
## Summary

Fixes #85 by implementing lazy browser-based OAuth authentication that triggers only when the Chat panel is opened without a Matrix session token. This restores Chat/Device Info functionality on desktop after PR #83 disabled automatic browser-OAuth on connect.

## Problem

After PR #83:
- Browser-OAuth gated behind `PICK_ENABLE_BROWSER_AUTH=1` env var
- Desktop webview doesn't receive `__st` token (no Strike48 proxy injection)
- Connector JWT cannot be exchanged for Matrix session token (no API exists)
- **Result:** Chat panel completely broken on desktop

## Investigation Results

Confirmed via code analysis:
1. **Desktop webview doesn't get `__st`** - it connects directly to local LiveView server without proxy
2. **No JWT exchange API** - connector JWT is gRPC-only, cannot become GraphQL session token
3. **Solution:** Lazy browser-OAuth (trigger on Chat panel visit)

## Solution

Implement lazy authentication that triggers only when user opens Chat:

```rust
// ChatPanel checks for empty token when becoming visible
if props.visible && effective_token.is_empty() && !browser_auth_attempted() {
    browser_auth_attempted.set(true);
    // Spawn async browser OAuth flow
    fetch_matrix_token_browser(&api_url).await
}
```

## Benefits

- ✅ Chat works on desktop without `PICK_ENABLE_BROWSER_AUTH=1`
- ✅ No browser popup on connect (only when Chat is clicked)
- ✅ Device Info / Network Scan quick actions work again
- ✅ Leverages existing browser OAuth (no new code paths)
- ✅ Minimal UX friction (auth happens when needed)

## Implementation Details

**Added to `crates/ui/src/components/chat_panel/mod.rs`:**
- `browser_auth_attempted` signal to track auth state
- Conditional check when panel becomes visible
- Async spawn of `fetch_matrix_token_browser()`
- Session store update on success
- User-visible logs for auth status

**Triggers when ALL true:**
- Chat panel visible
- `effective_token` is empty
- `api_url` is not empty
- Haven't attempted auth yet
- Agents not loaded

## Test Plan

- [ ] Launch desktop app without `PICK_ENABLE_BROWSER_AUTH` set
- [ ] Connect to Strike48 (no browser popup on connect)
- [ ] Click Chat tab
- [ ] Verify browser opens for OAuth
- [ ] Complete OAuth flow
- [ ] Verify Chat panel loads agents successfully
- [ ] Click Device Info quick action
- [ ] Verify it opens Chat with pre-filled prompt

## Related

- Fixes #85 (P1 bug)
- Unblocks PR #83 from merging
- Related to #81 (browser OAuth port issues)

## Validation

- ✅ `cargo check --all-targets` passes
- ✅ `cargo clippy --all-targets -- -D warnings` passes  
- ✅ `cargo fmt --all` passes
- ✅ No new dependencies added
- ✅ Leverages existing `fetch_matrix_token_browser()` code

### DIFF
--- a/crates/ui/src/components/chat_panel/mod.rs
+++ b/crates/ui/src/components/chat_panel/mod.rs
@@ -191,6 +191,51 @@ pub fn ChatPanel(props: ChatPanelProps) -> Element {
     }
 
     // -----------------------------------------------------------------------
+    // Lazy browser-OAuth: trigger on first Chat panel visit when token is missing
+    // -----------------------------------------------------------------------
+
+    // Track whether we've attempted browser auth
+    let mut browser_auth_attempted = use_signal(|| false);
+
+    // If Chat panel is visible, we don't have a token, and we haven't tried browser auth yet
+    if props.visible
+        && effective_token.is_empty()
+        && !api_url.is_empty()
+        && !browser_auth_attempted()
+        && !agents_loaded()
+    {
+        browser_auth_attempted.set(true);
+        let api_url_clone = api_url.clone();
+
+        crate::liveview_server::push_terminal_line(TerminalLine::info(
+            "[chat] No auth token — opening browser for authentication...",
+        ));
+
+        spawn(async move {
+            match pentest_core::matrix::fetch_matrix_token_browser(&api_url_clone).await {
+                Ok(token) => {
+                    tracing::info!(
+                        "[chat] Browser auth succeeded, token length: {}",
+                        token.len()
+                    );
+                    crate::liveview_server::set_matrix_credentials(&api_url_clone, &token);
+                    crate::session::set_auth_token(&token);
+                    crate::liveview_server::push_terminal_line(TerminalLine::success(
+                        "[chat] Authentication successful — chat ready",
+                    ));
+                }
+                Err(e) => {
+                    tracing::warn!("[chat] Browser auth failed: {}", e);
+                    crate::liveview_server::push_terminal_line(TerminalLine::error(format!(
+                        "[chat] Authentication failed: {}",
+                        e
+                    )));
+                }
+            }
+        });
+    }
+
+    // -----------------------------------------------------------------------
     // Fetch agents when we have a token
     // -----------------------------------------------------------------------
 


taylor-wood-devo (approved):
LGTM — no regression on the OTT pre-approval path (verified end-to-end: fresh creds → OTT → JWT → chat creates conversation and streams response). The new lazy-OAuth block is guarded behind effective_token.is_empty() so it stays dormant when any auth path already produced a token; covers the narrow edge case where no saved creds + no bridge token. Minor: render-time spawn slightly unidiomatic (vs use_effect) but guard prevents loops.

PR #109, opened 2026-05-18
fix: add retry logic with exponential backoff for OTT redemption

### LINKED ISSUE #84: Connector silently wedges on failed OTT redemption (no retry, no error surfaced)
## Observed behavior

When the server-side OTT redemption at `/api/connectors/register-with-ott` returns a non-2xx response, the connector logs the error once and then sits idle with no retry, no reconnect, and no clear signal to the user that the connector is dead in the water.

Observed on a Strike48 tenant. The first connection attempt for instance `pick-headless-*` hit a **500 Internal Server Error** on OTT redemption; the second attempt with a fresh instance_id succeeded on the same endpoint, suggesting the server retained corrupt state from the first attempt. That's a server-side issue we can't fix here, but on the client side Pick should handle a failed OTT redemption gracefully.

## Log excerpt (sanitized)

```
INFO  CredentialsIssued received, processing OTT exchange...
INFO  [step] ExchangingToken
INFO  [log] Processing approval credentials...
DEBUG OTT registration: api_url=https://<host>/, register_url=/api/connectors/register-with-ott
DEBUG OTT registration response status: 500 Internal Server Error (attempt 1)
ERROR Failed to register public key with OTT: Registration error: Post-approval OTT registration failed: {"error":"Registration failed"}
INFO  [log] OTT registration failed: Registration error: Post-approval OTT registration failed: {"error":"Registration failed"}
(no further activity — connector holds wss stream idle forever)
```

Code path: `crates/ui/src/liveview_connector/auth.rs::handle_credentials_issued` — the error branch at the bottom only logs; no status change, no retry, no reconnect.

## Impact

- Headless agents deployed against an unreliable Strike48 instance may wedge silently and never recover without human intervention.
- Desktop users see a confusing state: the UI is still on "Exchanging credentials..." but nothing is happening. The #83 fix keeps the connecting screen visible, but the callout now misleads because the approval has already happened — the failure is downstream.
- The "attempt 1" in the log line implies a retry, but no retry actually happens for HTTP 500s in this path.

## Suggested fixes

Pick one or more. These compose:

1. **Retry the OTT redemption with backoff.** 2-3 attempts with 1s/2s/4s jitter. Transient server errors would self-heal without user intervention. Applies only to 5xx responses and network errors; 4xx should not retry because they typically indicate a permanent client-side mismatch.
2. **Surface the failure in the UI/logs with actionable wording.** If the retry budget is exhausted, change the connecting step to `Error` with a message like *"Server couldn't redeem the approval token. Remove this connector from Prospector Studio and reconnect to retry."* The StatusChanged(Error) path is already wired up for other failures.
3. **Fall back to `WaitingForApproval` state.** If the server might re-issue `CredentialsIssued` (e.g., admin clicks Approve again), we could return to waiting instead of errroring out. Behavior depends on whether Strike48 supports re-issuing OTTs.
4. **Tear down the stream and reconnect from scratch.** If the server state is corrupt, a fresh register-with-OTT cycle might succeed. Equivalent to the "restart with new instance_id" workaround we hit manually.

Option 1 + option 2 is the smallest-risk baseline. Option 4 is more invasive but addresses the root-cause behavior we observed.

## Reproduction

Not deterministic (depends on server-side state). To reproduce synthetically:

1. Stub the POST `/api/connectors/register-with-ott` endpoint to return 500.
2. Launch `pentest-agent` pointed at the stubbed server; approve in Prospector Studio.
3. Observe: connector logs the error once; `pgrep pentest-agent` shows it alive; stream is idle; no retry attempted.

## Related

- Discovered while validating PR #77 and PR #83 against a live Strike48 tenant.
- Same silent-wedge pattern as issue #82 (stale credentials startup stall). Both issues point at the broader theme: connector error-recovery paths are incomplete.

## Recommended agent

**backend-architect** — this is a state machine / retry logic change in `crates/ui/src/liveview_connector/auth.rs` and `mod.rs`.

jtomek-strike48:
Fixes #84

## Summary

Connector previously silently wedged when OTT redemption failed (network transient errors, server busy, etc.). This PR adds retry logic with exponential backoff to handle transient failures gracefully.

**Changes:**
- Retry loop with MAX_RETRIES = 3
- Exponential backoff: 1s, 2s, 4s delays between attempts
- User-visible terminal logging for each retry attempt
- On success: exits loop and continues normal flow
- On failure: returns to WaitingForApproval state with clear error message directing user to re-approve in Prospector Studio

## Test Plan

- [ ] Manual test: Trigger OTT redemption failure (network disconnect, invalid OTT)
- [ ] Verify retry attempts with exponential backoff delays
- [ ] Verify user-visible terminal messages during retries
- [ ] Verify returns to WaitingForApproval on exhaustion
- [ ] Verify normal flow when redemption succeeds

## Related

Part of bug-fixing sprint to knock out all P1/P2/P3 issues.

### DIFF
--- a/crates/ui/src/liveview_connector/auth.rs
+++ b/crates/ui/src/liveview_connector/auth.rs
@@ -108,66 +108,109 @@ impl LiveViewConnector {
             Some(self.config.instance_id.clone()),
         );
 
-        // Register public key with OTT
-        match ott_provider
-            .register_public_key_with_ott_data(
-                &creds.ott,
-                ott_api_url,
-                &creds.register_url,
-                &connector_type,
-                Some(&self.config.instance_id),
-            )
-            .await
-        {
-            Ok(response) => {
+        // Register public key with OTT (with retry for transient errors)
+        const MAX_RETRIES: u32 = 3;
+        let mut last_error = None;
+
+        for attempt in 1..=MAX_RETRIES {
+            if attempt > 1 {
+                let backoff_ms = 1000 * (1 << (attempt - 2)); // 1s, 2s, 4s
                 tracing::info!(
-                    "Registered public key with OTT. Client ID: {}",
-                    response.client_id
+                    "OTT registration attempt {} failed, retrying in {}ms",
+                    attempt - 1,
+                    backoff_ms
                 );
-                self.send_event(ConnectorEvent::Log(TerminalLine::success(format!(
-                    "Key registered: {}",
-                    response.client_id
+                self.send_event(ConnectorEvent::Log(TerminalLine::info(format!(
+                    "Retrying OTT registration (attempt {}/{}) in {}ms...",
+                    attempt, MAX_RETRIES, backoff_ms
                 ))));
+                tokio::time::sleep(tokio::time::Duration::from_millis(backoff_ms)).await;
+            }
 
-                // Get JWT using private_key_jwt
-                match ott_provider.get_token().await {
-                    Ok(jwt_token) => {
-                        tracing::info!("Obtained JWT, will reconnect with JWT authentication");
-                        self.send_event(ConnectorEvent::Log(TerminalLine::success(
-                            "JWT obtained, reconnecting...",
-                        )));
+            match ott_provider
+                .register_public_key_with_ott_data(
+                    &creds.ott,
+                    ott_api_url,
+                    &creds.register_url,
+                    &connector_type,
+                    Some(&self.config.instance_id),
+                )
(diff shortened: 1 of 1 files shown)


taylor-wood-devo (approved):
LGTM — retry logic is well-bounded, logs are clear, exhaustion path correct. Optional nit (don't block on it): code comment mentions '1s, 2s, 4s' but only two delays actually fire since loop tops at 3 attempts.

PR #110, opened 2026-05-18
fix: reduce startup stall on stale credentials from 90s to 5s

### LINKED ISSUE #82: Startup stalls ~90s on stale credentials pointing at unreachable auth URL
## Observed behavior

On launch, the connector can hang for ~60-90s before reaching the Connect UI when a previously-saved credentials file in `~/.strike48/credentials/` points at an auth URL that no longer resolves or is unreachable.

Representative log excerpt from an end-to-end session:

```
[ConnectAndRun] No saved auth_token — trying pre-approval auth
Looking for saved credentials at: ~/.strike48/credentials/pentest-connector_<uuid>.json
Credentials file found, loading...
Loaded saved credentials ... client_id=matrix:connector:local:non-prod:pentest-connector:<uuid>
Loaded saved credentials, will use JWT authentication
Token request attempt 1/3 to https://auth.strike48.test/realms/non-prod/protocol/openid-connect/token
[30s timeout]
Token request failed (attempt 1/3), retrying in 500ms...
Token request attempt 2/3 to https://auth.strike48.test/realms/non-prod/protocol/openid-connect/token
[30s timeout]
...
```

In the session that surfaced this, the stale file pointed at a former local-dev realm (`auth.strike48.test` / `non-prod`) that the workstation could no longer resolve. The SDK's token refresh made three attempts with 30s timeouts each before finally giving up.

## Why it happens

`crates/ui/src/liveview_connector/mod.rs` cleans up stale credentials files in two places:

1. In the reconnect loop, if `ott.get_token().await` returns an error (line ~499). Good.
2. In the post-registration handler, if the server replies with an auth error (line ~770 area). Good.

What neither path covers: **the token-refresh attempt that runs before the first registration hits the wire.** If that refresh fails for a network/DNS/connect reason (not an "auth" keyword in the error), the code retries through the SDK's internal 3x30s loop before returning an error. During that window, the UI is blocked in the async task.

Additionally, the credentials file's `keycloak_url` is trusted without a liveness check, so an unreachable host can't be short-circuited early.

## Impact

- Users upgrading from an older dev build to production, or switching between realms, see a startup hang of up to ~90s with no UI feedback.
- The hang is silent from the user's perspective — the "Connecting..." screen sits there with no indication that the problem is a stale credential rather than network latency.
- Once the 90s elapses, the code does the right thing and falls through to OTT registration, so the bug is "just" a UX stall rather than a functional failure.

## Suggested fixes

Pick one or more:

1. **Reduce SDK token-refresh retry budget on startup.** 3x30s is appropriate for an already-registered connector that should self-heal, but on the *initial* refresh (when we have no idea whether the credentials are still valid), 1x5s is better. Fall through to OTT registration faster.
2. **Add a pre-flight reachability check** on the credentials file's `keycloak_url` (e.g., `reqwest::Client::builder().connect_timeout(Duration::from_secs(3))` on a simple GET). If it fails, treat the credentials as stale and delete them immediately.
3. **Show progress in the UI** during the token refresh. "Checking saved credentials..." with a Cancel button would at least make the hang visible and interruptible.
4. **Validate the credentials file's URL scheme/host against the currently-configured host** on load. If they don't match a recognizable pattern, treat as stale.

Option 1 combined with option 3 is the smallest-risk immediate improvement. Option 2 is more invasive but eliminates the stall entirely.

## Reproduction

1. Save a working credentials file at `~/.strike48/credentials/pentest-connector_<device-id>.json` with a valid `keycloak_url`.
2. Edit the file and change `keycloak_url` to an unreachable host (e.g., `https://auth.strike48.test/...`).
3. Launch the connector.
4. Observe ~90s hang on the connecting screen before it falls through to OTT-approval flow.

## Related

- Discovered while validating PR #77. The credentials wipe workaround was to `rm -f ~/.strike48/credentials/*.json && rm -f ~/.config/pentest-connector/settings.json`.

## Recommended agent

**backend-architect** for option 1/2, **ui-builder** for option 3.

jtomek-strike48:
Fixes #82

## Summary

Connector startup previously stalled for ~90 seconds when saved credentials pointed at an unreachable auth URL (e.g., switching from local-dev to production), because the SDK's token refresh made 3 attempts with 30s timeouts each.

This PR adds a **5-second timeout on the first connection attempt** to fail fast on unreachable auth URLs.

**Changes:**
- Wrap first `ott.get_token()` call with `tokio::time::timeout(Duration::from_secs(5))`
- On timeout: clean up stale credentials file, clear token, set `ott_provider = None`
- Connector falls through to OTT-approval flow (no 90s stall)
- On subsequent reconnects (`connection_failures > 0`): use full SDK retry budget (credentials are known-good)

## Test Plan

- [ ] Manual test: Create stale credentials file with unreachable auth URL
- [ ] Verify connector starts in <10s (previously ~90s)
- [ ] Verify stale credentials file is removed
- [ ] Verify falls through to OTT-approval flow
- [ ] Verify normal flow when credentials are valid (no timeout)

## Related

Part of bug-fixing sprint (all P1/P2/P3 issues). Follows PR #109 (OTT retry logic).

### DIFF
--- a/crates/ui/src/liveview_connector/auth.rs
+++ b/crates/ui/src/liveview_connector/auth.rs
@@ -108,66 +108,109 @@ impl LiveViewConnector {
             Some(self.config.instance_id.clone()),
         );
 
-        // Register public key with OTT
-        match ott_provider
-            .register_public_key_with_ott_data(
-                &creds.ott,
-                ott_api_url,
-                &creds.register_url,
-                &connector_type,
-                Some(&self.config.instance_id),
-            )
-            .await
-        {
-            Ok(response) => {
+        // Register public key with OTT (with retry for transient errors)
+        const MAX_RETRIES: u32 = 3;
+        let mut last_error = None;
+
+        for attempt in 1..=MAX_RETRIES {
+            if attempt > 1 {
+                let backoff_ms = 1000 * (1 << (attempt - 2)); // 1s, 2s, 4s
                 tracing::info!(
-                    "Registered public key with OTT. Client ID: {}",
-                    response.client_id
+                    "OTT registration attempt {} failed, retrying in {}ms",
+                    attempt - 1,
+                    backoff_ms
                 );
-                self.send_event(ConnectorEvent::Log(TerminalLine::success(format!(
-                    "Key registered: {}",
-                    response.client_id
+                self.send_event(ConnectorEvent::Log(TerminalLine::info(format!(
+                    "Retrying OTT registration (attempt {}/{}) in {}ms...",
+                    attempt, MAX_RETRIES, backoff_ms
                 ))));
+                tokio::time::sleep(tokio::time::Duration::from_millis(backoff_ms)).await;
+            }
 
-                // Get JWT using private_key_jwt
-                match ott_provider.get_token().await {
-                    Ok(jwt_token) => {
-                        tracing::info!("Obtained JWT, will reconnect with JWT authentication");
-                        self.send_event(ConnectorEvent::Log(TerminalLine::success(
-                            "JWT obtained, reconnecting...",
-                        )));
+            match ott_provider
+                .register_public_key_with_ott_data(
+                    &creds.ott,
+                    ott_api_url,
+                    &creds.register_url,
+                    &connector_type,
+                    Some(&self.config.instance_id),
+                )
+                .await
+            {
+                Ok(response) => {
+                    tracing::info!(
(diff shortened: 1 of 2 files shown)


taylor-wood-devo (approved):
LGTM — 5s timeout only on first connection attempt is the right asymmetry; reconnects keep full SDK retry budget. Clean stale-credentials cleanup. Minor edge case: users on >5s latency networks could see false-positive credential deletion, but acceptable trade-off for the 90s→5s startup improvement.

PR #111, opened 2026-05-18
fix: fail fast with clear error when OAuth callback ports are busy

### LINKED ISSUE #81: Browser OAuth callback binds to random port when 4000 and 5173 are busy, breaking CORS

jtomek-strike48:
Fixes #81

## Summary

Browser OAuth callback previously fell back to a random OS-assigned port when both 4000 and 5173 were busy, then **silently timed out after 120 seconds** because the Matrix server's CORS whitelist only includes the fixed ports.

This PR removes the random-port fallback and **fails fast with a clear, actionable error message**.

**Changes:**
- Remove fallback to `127.0.0.1:0` (random port)
- Return immediate error with guidance:
  - Which ports are required (4000 or 5173)
  - How to check what's using them (`lsof -i :4000`, `lsof -i :5173`)
  - What to stop (dev servers: Vite, Vue CLI, etc.)
- Fail fast (immediate error) instead of 120s silent timeout

**Before:** 120s hang → "Login timed out" (no guidance)  
**After:** Immediate error with actionable steps

## Test Plan

- [ ] Manual test: Bind ports 4000 and 5173 (`python3 -m http.server 4000`, `python3 -m http.server 5173`)
- [ ] Attempt browser OAuth login
- [ ] Verify immediate clear error message (no 120s hang)
- [ ] Verify error message includes `lsof` guidance
- [ ] Stop blocking processes, verify OAuth succeeds

## Related

Part of bug-fixing sprint (all P1/P2/P3 issues). Follows PR #110 (startup timeout).

### DIFF
--- a/crates/core/src/matrix/auth.rs
+++ b/crates/core/src/matrix/auth.rs
@@ -323,15 +323,18 @@ pub async fn fetch_matrix_token_browser(matrix_url: &str) -> crate::error::Resul
                     l
                 }
                 Err(_) => {
-                    tracing::warn!("[BROWSER_AUTH] Ports 4000 and 5173 busy, using random port (CORS may fail)");
-                    tokio::net::TcpListener::bind("127.0.0.1:0")
-                        .await
-                        .map_err(|e| {
-                            crate::error::Error::Matrix(format!(
-                                "Failed to bind callback server: {}",
-                                e
-                            ))
-                        })?
+                    // Ports 4000 and 5173 are required for CORS (Matrix server whitelist).
+                    // Using a random port would cause the callback to timeout due to CORS rejection.
+                    tracing::error!(
+                        "[BROWSER_AUTH] Ports 4000 and 5173 are busy. Browser OAuth requires one of these ports. \
+                         Stop the process using these ports (likely dev servers: `lsof -i :4000` and `lsof -i :5173`)."
+                    );
+                    return Err(crate::error::Error::Matrix(
+                        "matrix: Browser OAuth requires ports 4000 or 5173 to be available. \
+                         Both ports are currently in use. Stop any dev servers (Vite, Vue CLI, etc.) \
+                         or other processes using these ports and try again. \
+                         To check: `lsof -i :4000` and `lsof -i :5173`".to_string()
+                    ));
                 }
             }
         }
--- a/crates/ui/src/liveview_connector/auth.rs
+++ b/crates/ui/src/liveview_connector/auth.rs
@@ -108,66 +108,109 @@ impl LiveViewConnector {
             Some(self.config.instance_id.clone()),
         );
 
-        // Register public key with OTT
-        match ott_provider
-            .register_public_key_with_ott_data(
-                &creds.ott,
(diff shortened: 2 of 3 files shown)


taylor-wood-devo (approved):
LGTM — happy path verified clean; PR is invisible when ports are free. Replaces a 120s silent failure with an immediate actionable error including lsof guidance. Tracing escalation from warn→error appropriate.

Issue #125, opened 2026-05-26
Mobile: "Capture" label clipped to "Captur" on narrow screens


jtomek-strike48:
## Symptom

On the mobile (Pixel 10) UI, a label or button that should read **"Capture"** displays as **"Captur"** — the trailing "e" is clipped.

## Root cause (suspected)

CSS overflow / text-truncation in a narrow container. All Rust source occurrences spell it correctly, so this is presentational, not a typo. Likely candidates:
- Tools page button label
- Status bar / dashboard label
- Sidebar / nav entry

## To investigate

- [ ] Identify which component renders the clipped label (screenshot needed)
- [ ] Check container width constraints, `text-overflow`, `white-space: nowrap`
- [ ] Verify behavior across breakpoints (tablet vs phone)

## Reproduce

1. Run mobile APK on Pixel 10 (Android, ~411dp width)
2. Navigate to wherever the Capture label appears
3. Observe truncation

## Recommended agent

ui-builder

Issue #124, opened 2026-05-26
Mobile: Auth Token required error shown immediately after connect, before OAuth completes


jtomek-strike48:
## Symptom

After PR #TBD (connect-screen error visibility) lands, the connect path now succeeds end-to-end on mobile:

- TCP + TLS + WebSocket handshake completes
- Strike48 registration succeeds (`Registered: matrix:use-non-c-discoball:pentest-connector:...`)

But the user immediately sees the error: **"Auth Token is required"**

## Root cause

The connector's downstream code path tries to make Matrix GraphQL calls (chat / agent / conversation list) before the browser-based OAuth flow has completed. The check that fires is:

```rust
// crates/core/src/matrix/client.rs:75-80
fn require_auth(&self) -> crate::error::Result<()> {
    if self.auth_token.is_none() {
        return Err(crate::error::Error::Matrix("Auth token required".into()));
    }
    Ok(())
}
```

The trigger is `[FetchTokenFallback] Trying browser-based OAuth login` followed by Matrix queries firing before the user has actually completed the OAuth flow in their browser.

## Reproduce

1. Install mobile APK with no saved auth token
2. Connect to a Strike48 host that requires OAuth (e.g. `wss://discoball.strike48.engineering` + tenant `use-non-c-discoball`)
3. Observe: connector registers successfully, then "Auth Token is required" surfaces

## Expected

Either:
- (a) Wait for OAuth completion before triggering Matrix calls, OR
- (b) Show a clear "Sign in to continue" state instead of an error
- (c) Disable / hide the Matrix-dependent UI until a token is present

## Notes

This is a UX gap, not a regression. It existed before — users just couldn't get this far because the connect flow was silently failing.

## Recommended agent

ui-builder (UI state for unauthenticated mode) + api-architect (Matrix call gating)

PR #126, opened 2026-05-26
fix: surface connect-screen errors and infer transport defaults

jtomek-strike48:
## Summary

Fixes the Connect button silent-failure on mobile. Typing `wss://discoball.strike48.engineering` (no port) or a bare hostname previously gave zero feedback — `normalize_host` rejected the input, but the error went into a terminal buffer the Connect screen never renders. Connection failures had the same fate.

This PR makes connect-phase errors visible AND infers sensible transport defaults so users typically don't hit the validation path at all.

## Changes

**`crates/core/src/config.rs`**
- New `NormalizedHost { value, inferred_scheme, inferred_port }` return type. Carries enough context for the UI to disclose what was inferred.
- Scheme-aware port inference table: `wss/https→443`, `ws/http→80`, `grpc→50051`, `grpcs→443`. Bare hosts default to `wss://...:443` (Strike48-on-Cloudflare common case).
- Inference policy documented inline (option C: visible inference) with the why and a clear revert path.

**`crates/ui/src/components/config_form.rs`**
- New `external_error: Option<String>` prop. Parent pipes connect-phase errors into the existing red banner.
- Live `inference_hint` rendered under the host field as `Will connect as wss://host:443` — only when inference actually changed the input. Explicit URLs stay quiet.

**`crates/ui/src/connector_app.rs`**
- New `connect_error` signal. Both `normalize_host` validation failures and `connect_and_run` transport failures populate it.
- Errors continue to log to `terminal_lines` for Dashboard debugging — banner adds visible parity.

**`crates/core/src/settings.rs`**
- Migration call site updated for the new return shape.

## Why option C (visible inference)?

Documented inline in `config.rs::normalize_host`. Short version:

1. The dominant Strike48 case is Cloudflare-fronted WebSocket on :443 — forcing users to type the full URL is friction.
2. Once gRPC ships, the same bare host could mean either transport. Showing the resolved URL lets users verify intent up-front instead of debugging mysterious routing later.
3. Trustworthy tooling (`cargo`, `gh`, `kubectl`) reports what it resolved. Pentest users distrust magic — match that contract.

To revert: drop the `hint()` rendering in `ConfigForm`, or change the no-scheme/no-port branch in `normalize_host` to return `Err`.

## Test plan

- [x] `cargo fmt --all`
- [x] `cargo check --all-targets`
- [x] `cargo clippy --all-targets -- -D warnings`
- [x] `cargo test --lib --bins` — 296 tests pass, 0 failures
- [x] 14 new `normalize_host` cases: positive coverage for every scheme, idempotence, whitespace trim, case-insensitive scheme, malformed input rejection
- [x] **Verified end-to-end on Pixel 10**: saved bare host migrates to `wss://...:443` on app start, transport flips from gRPC to WebSocket, TCP+TLS+WebSocket handshake completes, and Strike48 registration succeeds against `use-non-c-discoball`

## Discovered follow-ups (not in this PR)

While testing this fix on a real device we surfaced two pre-existing issues that were previously hidden behind the silent-connect failure:

- #124 — Mobile: \"Auth Token required\" error shown immediately after connect, before OAuth completes
- #125 — Mobile: \"Capture\" label clipped to \"Captur\" on narrow screens

Neither is a regression of this PR. Both deserve their own focused fixes.

## Build note

The justfile `build-android` recipe omits `--target`, which causes `dx build` to skip arm64 cross-compile and silently package stale arm64 `.so` libraries from previous builds. Worth a separate PR to add explicit target handling — caught while testing this change.

### DIFF
--- a/crates/core/src/config.rs
+++ b/crates/core/src/config.rs
@@ -114,6 +114,50 @@ impl Default for ConnectorConfig {
     }
 }
 
+/// Outcome of [`ConnectorConfig::normalize_host`].
+///
+/// Carries the canonical URL plus a record of which parts (scheme, port) were
+/// supplied by inference rather than typed by the user. Callers display the
+/// inference via [`Self::hint`] so users can verify the resolved transport
+/// before connecting — see the doc on `normalize_host` for the rationale.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct NormalizedHost {
+    /// Canonical URL with scheme and explicit port.
+    pub value: String,
+    /// `Some(scheme)` if `normalize_host` supplied a scheme the user did not type.
+    pub inferred_scheme: Option<&'static str>,
+    /// `Some(port)` if `normalize_host` supplied a port the user did not type.
+    pub inferred_port: Option<u16>,
+}
+
+impl NormalizedHost {
+    /// `true` if any defaulting occurred (scheme or port was supplied).
+    pub fn was_inferred(&self) -> bool {
+        self.inferred_scheme.is_some() || self.inferred_port.is_some()
+    }
+
+    /// User-facing line describing what we resolved, e.g.
+    /// `Will connect as wss://discoball.strike48.engineering:443`.
+    /// Returns `None` when the user typed everything explicitly.
+    pub fn hint(&self) -> Option<String> {
+        if self.was_inferred() {
+            Some(format!("Will connect as {}", self.value))
+        } else {
+            None
+        }
+    }
+}
+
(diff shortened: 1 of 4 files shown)


taylor-wood-devo (approved):
Rebased on main to pick up the Kali apt CDN fix from #133, CI is fully green on the new SHA (Docker amd64/arm64 both pass). Manually exercised the Connect screen with the inference and error-surfacing cases from the test plan; all behaved as documented. Inference hint only renders when something was actually inferred, and connect-phase errors now land in the red banner instead of going silent. LGTM.

PR #129, opened 2026-05-26
Add support for nix

josh-devo:
mostly just stuff i had to change to make it work on nix though i do want to call out that it downgrades proot right now by a minor version and that seems to be because 5.4.0 binaries don't exist in the release


josh-devo:
closing this for now, will reopen if i run into issues basically

PR #132, opened 2026-05-27
fix: route chat panel off dioxus-liveview broken event converters (#130)

### LINKED ISSUE #130: Panic in dioxus-liveview when loading conversation from history
## Repro

1. Launch desktop against remote studio:
   ```
   MATRIX_API_URL=https://studio.strike48.test:443 \
   STRIKEHUB_SOCKET=/tmp/pick-wait \
   cargo run --package pentest-desktop
   ```
2. Chat loads agents (ListAgents → 200), ListConversations → 200, ChatPanel auto-selects pentest-connector agent.
3. Click a previous conversation in history.
4. GetConversation returns 200, then process aborts.

Observed on branch `fix/issue-82-startup-timeout` (PR #110) but the panic site is in an external crate, so it almost certainly reproduces on `main`.

## Crash

```
thread 'main' (2943550) panicked at /Users/.../dioxus-liveview-0.7.3/src/events.rs:91:54:
called \`Option::unwrap()\` on a \`None\` value
note: run with \`RUST_BACKTRACE=1\` environment variable to display a backtrace
```

## Last log lines before crash

```
22:24:01.935  DEBUG pentest_ui::liveview_connector: Unhandled message variant
22:24:02.632  INFO  [gql] POST .../api/v1alpha (token_len=460 query=query ListConversations(...))
22:24:02.683  INFO  [gql] response status=200 OK
22:24:04.899  DEBUG [Matrix] get_conversation id=Q29udmVyc2F0aW9uOjAxOWU2NjVmLWM1ZDctNzg0Yy1iMjQ3LWJlNWMyZWMwYjE5Yw==
22:24:04.899  INFO  [gql] POST .../api/v1alpha (token_len=460 query=query GetConversation($id: ID!) {)
22:24:04.956  INFO  [gql] response status=200 OK
<panic>
```

The `Unhandled message variant` debug line may be related — pick is receiving a message variant it doesn't handle, possibly the same one whose missing field causes dioxus-liveview to unwrap None.

## Investigation needed

- Inspect `dioxus-liveview-0.7.3/src/events.rs:91` to identify which field is being unwrapped.
- Determine whether this is:
  1. An upstream dioxus-liveview bug — file with dioxus repo, pin to safer version or patch.
  2. A pick bug — we're sending an event/message without an expected field, fixable in our event proxy code (`crates/ui/src/liveview_connector/tools.rs` or `mod.rs`).
- Capture full backtrace with `RUST_BACKTRACE=1` on next repro for clearer signal.

## Impact

Blocks loading historical conversations on desktop. Workaround: avoid clicking conversation history items; start new conversations only.

## Environment

- Desktop: pentest-desktop
- macOS Darwin 25.3.0
- dioxus-liveview = 0.7.3 (per workspace Cargo.toml: `dioxus-liveview = { version = "0.7", features = ["axum"] }`)
- Studio: studio.strike48.test:443 (local cluster)

taylor-wood-devo:
Fixes #130.

## Summary

dioxus-liveview-0.7.x's `SerializedHtmlEventConverter` (`events.rs`) contains the same `.unwrap()` bug across `convert_mounted_data`, `convert_form_data`, and `convert_keyboard_data`: each downcasts a `PlatformEventData` payload and panics if it isn't the expected type. The unwrap reliably trips once a conversation has accumulated some amount of DOM state — historically reproduced on history reload, but also on typing/sending after a few back-and-forth turns.

There is no clean upstream fix yet (same code on `main` and 0.7.9). The pragmatic workaround is to stop sending the affected event types down dioxus's wire-event path for the chat panel. This PR does that for the three converters we hit.

## Changes

Two commits:

1. **`fix: avoid dioxus-liveview onmounted panic on history reload`** — removes the per-bubble `onmounted` listener in `chat_panel/render.rs` and moves the `triggerChartPostProcess()` call into the existing `MessageList` `use_effect`. Same chart rendering, fewer listeners, no panic on history reload.

2. **`fix: route chat input off dioxus-liveview broken event converters`** — removes `onsubmit` / `onkeydown` / `oninput` from the chat textarea. Auto-resize, Enter-to-send, and Send-button clicks are now handled by a single document-level delegated listener installed by `installChatSendBridge` in `utils.js`. Submissions return to Rust over a long-lived `document::eval` channel (`dioxus.send(text)` → `eval.recv::<String>()`).

## Why a JS bridge over a narrower workaround

A narrower fix (drop only `oninput`, keep `onsubmit` and `onkeydown`) was tried and confirmed insufficient — typing no longer panicked but Enter-to-send still tripped `convert_form_data` on the form submit. `onsubmit` and `onkeydown` route through the same broken `.unwrap()` paths, so leaving them in place just defers the panic. Routing everything through the eval channel removes the whole class of triggers in one place.

## Side-benefits

- Zero WebSocket chatter per keystroke. Auto-resize and Enter detection are pure browser; no server round trip for typing.
- Send is one round trip (JS → eval channel → `on_send.call`) instead of one per keystroke plus one for submit.
- Works identically inside Strike48 studio iframes and across all OS/arch combinations (only uses `document.addEventListener` and the dioxus eval channel — no platform-specific paths).
- Listener install is idempotent (`window.__chatSendBridgeInstalled` guard); dispatch callback is rebound on every install so a re-mounted `ChatInput` correctly retargets to its new `dioxus.send`.

## Test plan

Local validation in `pentest-desktop` against `studio.strike48.test:443`:

- [x] Load a fresh conversation, type a single character — no panic (previously panicked in `convert_mounted_data` then later in `convert_form_data`).
- [x] Send a message via Enter — `[Matrix] send_message ... msg_len=N` matches typed text; agent streams response; polling reaches `IDLE done=true has_agent_msg=true`.
- [x] Load a previously-existing conversation from history (the original #130 repro) — no panic, messages render.
- [x] Continue chatting in a loaded historical conversation (the input-panic repro) — no panic, full send/stream/idle cycle completes.
- [ ] Send a message via Send-button click instead of Enter.
- [ ] Multi-line message via Shift+Enter — textarea grows up to 200px.
- [ ] Exercise inside Strike48 studio iframe.

## Rebased on main

Rebased onto `main` to pick up the Kali apt mirror CI fix from #133, so `Build amd64` / `Build arm64` should turn green.

## Upstream follow-up (not in this PR)

File an issue with DioxusLabs/dioxus with a `RUST_BACKTRACE=1` capture against `events.rs:62` / `:78` / `:91`. The shared `.unwrap()` pattern should at minimum be a graceful return.

### DIFF
--- a/crates/ui/src/assets/utils.js
+++ b/crates/ui/src/assets/utils.js
@@ -117,6 +117,66 @@
     };
 
     /**
+     * Install document-level delegated listeners for the chat textarea:
+     *  - `input` events auto-resize the textarea between 40px and 200px,
+     *  - `keydown` Enter (no Shift) dispatches a send,
+     *  - `click` on a `.chat-send-btn` dispatches a send.
+     *
+     * `sendCallback` is invoked with the textarea text on each send trigger.
+     * It is wired to `dioxus.send` by Rust so submissions return over the
+     * eval channel instead of through dioxus's broken `convert_form_data`
+     * path (#130). The callback is overwritten on every call so a re-mounted
+     * ChatInput points the listeners at the new eval's `dioxus.send`.
+     *
+     * Idempotent for the listener install; the callback rebind is intentional.
+     */
+    window.installChatSendBridge = function(sendCallback) {
+        window.__chatSendDispatch = sendCallback;
+        if (window.__chatSendBridgeInstalled) return;
+        window.__chatSendBridgeInstalled = true;
+
+        function fireSend() {
+            var el = document.querySelector('.chat-textarea');
+            if (!el || el.disabled) return;
+            var text = el.value;
+            if (!text || !text.trim()) return;
+            if (typeof window.__chatSendDispatch === 'function') {
+                window.__chatSendDispatch(text);
+            }
+        }
+
+        // Auto-resize on every keystroke.
+        document.addEventListener('input', function(e) {
+            var t = e.target;
+            if (t && t.classList && t.classList.contains('chat-textarea')) {
+                t.style.height = 'auto';
+                t.style.height = Math.min(Math.max(t.scrollHeight, 40), 200) + 'px';
+            }
+        });
+
+        // Enter (no Shift) inside the chat textarea sends.
+        document.addEventListener('keydown', function(e) {
+            var t = e.target;
+            if (t && t.classList && t.classList.contains('chat-textarea')
+                && e.key === 'Enter' && !e.shiftKey) {
+                e.preventDefault();
+                fireSend();
+            }
+        });
+
+        // Send button click sends. preventDefault stops the native form-submit
+        // path (button is type="submit") so we don't trigger any dioxus
+        // onsubmit listener — there shouldn't be one, but be defensive.
+        document.addEventListener('click', function(e) {
+            var t = e.target;
+            if (t && t.classList && t.classList.contains('chat-send-btn')) {
+                e.preventDefault();
+                fireSend();
+            }
+        });
+    };
+
+    /**
      * Trigger chart post-processing (mermaid + echarts) on the next animation frame.
      * Calls window.__processChatCharts if it has been defined by chart_processor.js.
      */
--- a/crates/ui/src/components/chat_panel/input.rs
+++ b/crates/ui/src/components/chat_panel/input.rs
@@ -55,11 +55,8 @@ pub fn ChatInput(props: ChatInputProps) -> Element {
                 await new Promise(function() {});
                 "#,
             );
-            loop {
-                match eval.recv::<String>().await {
-                    Ok(text) => on_send.call(text),
-                    Err(_) => break,
-                }
+            while let Ok(text) = eval.recv::<String>().await {
+                on_send.call(text);
             }
         });
     });
--- a/crates/ui/src/components/chat_panel/input.rs
+++ b/crates/ui/src/components/chat_panel/input.rs
@@ -1,6 +1,15 @@
-//! Chat input area: auto-resizing textarea, send button, and keyboard handler.
+//! Chat input area: auto-resizing textarea + Send button.
+//!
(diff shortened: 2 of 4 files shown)


jtomek-strike48 (approved):
## Code Review: Approved ✅

### Summary

Comprehensive code analysis confirms the JavaScript bridge implementation correctly addresses issue #130 (dioxus-liveview panic). The remaining untested scenarios (Send button, Shift+Enter multi-line, iframe) have HIGH confidence of passing based on code logic analysis.

### Key Findings

**✅ Send Button Click (Test 5)**
- Uses identical `fireSend()` path as Enter key (already validated)
- Proper event delegation with defensive checks
- `preventDefault()` blocks form submission

**✅ Multi-line Shift+Enter (Test 6)**
- Conditional logic: `e.key === 'Enter' && !e.shiftKey`
- Shift+Enter does NOT match, allowing natural newline insertion
- Auto-resize fires on input event, measures scrollHeight, clamps 40-200px

**⚠️ Studio Iframe (Test 7)**
- Uses standard DOM APIs (no platform-specific paths)
- Should work based on architecture, but needs real environment validation
- Recommend testing post-merge

### Code Quality

- ✅ Defensive programming throughout
- ✅ Idempotent bridge installation
- ✅ Proper error handling
- ✅ Clippy-clean (warning fixed in commit 3)
- ✅ All 22 CI checks passing

### Architecture

The JS bridge approach is sound:
- Eliminates entire class of dioxus event converter panics
- Reduces WebSocket chatter (no round trip per keystroke)
- Platform-agnostic implementation
- Clean separation of concerns

### Post-Merge Validation

- [ ] Test in studio iframe environment
- [ ] Verify textarea clearing behavior
- [ ] Monitor production for eval channel errors
- [ ] File upstream dioxus issue

### Recommendation

**MERGE** - High confidence based on code analysis. Remaining unknowns are low-risk and can be addressed post-merge if needed.

---

Reviewed by: Claude Code (comprehensive code analysis + logic verification)

PR #191, opened 2026-07-02
fix(chat): revert send path to native events for Windows/WebView2 (#189)

### LINKED ISSUE #130: Panic in dioxus-liveview when loading conversation from history

### LINKED ISSUE #189: Chat send does not work on Windows MSI (StrikeHub-hosted): JS bridge dioxus.send reverse channel unreliable in WebView2 LiveView
## Symptom

When Pick is loaded inside StrikeHub on Windows (MSI install), the chat panel renders and other apps in Matrix Studio behave normally, but **pressing Enter (or clicking Send) in Pick chat does nothing** — the message never leaves the input. Same StrikeHub-hosted flow works on Linux and macOS. KubeStudio's chat panel (also hosted in StrikeHub) works on Windows.

Reported: 2026-07-01.

## Root-cause hypothesis (high confidence, not empirically confirmed on Windows)

Pick's chat input routes send events through a JavaScript bridge that hands text back to Rust via \`document::eval\`'s reverse channel (\`dioxus.send(text)\` → \`eval.recv::<String>().await\`). See \`crates/ui/src/components/chat_panel/input.rs:55-74\` and \`crates/ui/src/assets/utils.js:133-177\`.

Multiple places in the StrikeHub codebase already document \`document::eval\` in WebView2 + LiveView mode as unreliable:

- \`strikehub/crates/sh-core/src/ws_relay.rs:197-198\` — \`__st\` token appended to the connector WebSocket URL specifically to \"avoid the fallback polling path that fails on Windows WebView2.\"
- \`strikehub/crates/sh-ui/src/app.rs:688-698\` — \`MATRIX_AUTH_TOKEN\` passed via env var because \"the \`document::eval()\` polling path fails on Windows WebView2.\"
- \`pick/crates/ui/src/components/chat_panel/mod.rs:568-572\` — comment above the \`spawn(document::eval(...))\` in \`send_message\`: \"On Windows WebView2 in LiveView mode, \`document::eval\` responses can hang indefinitely.\"

Every other consumer of that channel has been worked around. Pick chat send is the only remaining path relying on the JS→Rust eval channel, and it happens to be the one that breaks for the reported user.

## Why KubeStudio isn't affected

KubeStudio's chat input uses native Dioxus events (\`oninput\`, \`onkeydown\`, \`onclick\`) — see \`kubestudio/crates/ks-ui/src/components/chat_panel.rs:1068-1085\`. Those flow through Dioxus's ordinary event pipeline, not the eval reverse channel.

Pick had to move off native events in commit \`f0e5455\` (\"route chat panel off dioxus-liveview broken event converters, #130\") because dioxus-liveview 0.7's \`convert_form_data\` / \`convert_keyboard_data\` panics once chat DOM state accumulates. KubeStudio uses dioxus-liveview 0.6 and doesn't hit that panic. Net: the earlier fix traded one bug for another, and the second bug is Windows-only.

## Reproduction

1. Install the latest StrikeHub MSI on Windows.
2. Launch StrikeHub, sign in, connect to Pick.
3. Open the chat panel, select an agent.
4. Type a message and press Enter (or click Send). Nothing happens.
5. Repeat on Linux or macOS — works.

## Suggested devtools probe to confirm the failure mode

In WebView2 devtools inside the Pick iframe:

- \`typeof window.installChatSendBridge\` — expected \`function\`; if \`undefined\`, the \`document::eval(UTILS_JS)\` injection at \`chat_panel/mod.rs:167-171\` never landed.
- \`window.__chatSendBridgeInstalled\` — expected \`true\` after mount.
- Add a \`console.log('fire', text)\` in \`utils.js\` \`fireSend()\` (throwaway) and press Enter. If the log fires but Pick never receives, that isolates the failure to the \`dioxus.send\` reverse channel.

## Proposed fix

Replace the \`dioxus.send\` reverse channel with a plain HTTP POST from the JS bridge to a small Axum route on Pick's LiveView server. Server pushes text into a \`tokio::sync::broadcast\` channel that \`ChatInput\` subscribes to. This avoids both the original dioxus-liveview event-converter panic **and** the WebView2 eval-response hang.

Fix branch: \`fix/chat-send-windows-webview2-eval\`.

taylor-wood-devo:
Closes #189.

## Symptom

Pick's chat panel accepts typing but pressing Enter (or clicking Send) delivers nothing on the Windows MSI build of StrikeHub. Linux and macOS work. KubeStudio's chat panel (also hosted inside StrikeHub) works on Windows.

## Root cause

The rerouting in #132 sends keystrokes through `document::eval`'s reverse channel (`dioxus.send(text)` → `eval.recv::<String>().await`). That channel does not deliver in the WebView2 + LiveView + StrikeHub-iframe combination. Three other places in the StrikeHub stack already document this exact class of failure and route around it:

- `strikehub/crates/sh-core/src/ws_relay.rs:198` — `__st` token appended to the connector WebSocket URL to avoid the polling fallback that fails on Windows WebView2.
- `strikehub/crates/sh-ui/src/app.rs:691` — `MATRIX_AUTH_TOKEN` passed via env because the `document::eval()` polling path fails on Windows WebView2.
- `pick/crates/ui/src/components/chat_panel/mod.rs:568` — post-send UI evals are fire-and-forget because `document::eval` responses hang indefinitely on Windows WebView2 in LiveView mode.

Chat send was the one remaining consumer of that channel that hadn't been worked around.

## What #132 actually did, and what this reverts

#132 bundled two changes:

1. Removed per-bubble `onmounted` listeners in `messages.rs`.
2. Rerouted `onsubmit`/`onkeydown`/`oninput`/`onclick` through a JavaScript bridge (`installChatSendBridge` in `utils.js`) using the eval reverse channel.

Change (1) is what actually fixed #130. The panic at `dioxus-liveview-0.7.x/src/events.rs:91:54` is in `convert_mounted_data` (verified against #130's stack trace and the `messages.rs` diff in f0e5455), not `convert_form_data`/`convert_keyboard_data` as `input.rs` claimed. Change (2) was a defensive precaution that mistook the crash site and traded #130 for a Windows-only bug.

A `dioxus-liveview` version bump does not help here: 0.7.3 and 0.7.9 sources are byte-identical (`diff -r` returns empty). KubeStudio isn't spared by its 0.7.9 pin — it's spared because it never registered per-bubble `onmounted`, so it never reaches events.rs:91.

This PR reverts only the send-path rerouting from #132, matching KubeStudio's native-event pattern (`oninput` + `onkeydown` + `onclick`), and leaves the `messages.rs` `onmounted` removal intact so #130 stays fixed. Also removes the now-unused `installChatSendBridge` helper from `utils.js`.

## Verification

Empirically confirmed on Windows MSI:

1. Bumped `PICK_REF` on a strikehub test branch to this fix's SHA.
2. Windows MSI job in strikehub's `Release` workflow (`dry_run=true`) built pentest-agent from this branch and produced `strikehub-windows-x86_64.msi`.
3. Installed on a Windows machine. Because `sh-core/connector_fetch.rs` prefers the cached fetched binary over the bundled sibling, manually swapped `~/.strike48/strikehub/bin/pentest-agent.exe` with the MSI-bundled fix (SHA `88a67f71…84805`) while leaving the `pentest-agent.version` marker at `v0.1.6` so the background update check stays quiet.
4. Pick chat in StrikeHub sends messages via Enter and via Send button. Chat previously did nothing.

## Test plan for review

- [x] Native events fire on Windows MSI (confirmed 2026-07-02).
- [x] `messages.rs` retains `onmounted` removal so #130 does not regress on history reload.
- [x] `installChatSendBridge` no longer referenced anywhere in pick.
- [ ] Reviewer sanity check: Shift+Enter still inserts a newline (not send).
- [ ] Reviewer sanity check: history reload does not panic on macOS or Linux.

## Deployment note

Once merged and tagged (e.g. `v0.1.7`), existing StrikeHub installs on all platforms pick up the fixed `pentest-agent` on next launch via `connector_fetch` (`strikehub b20d29b`). No StrikeHub release needed.

### DIFF
--- a/crates/ui/src/assets/utils.js
+++ b/crates/ui/src/assets/utils.js
@@ -117,66 +117,6 @@
     };
 
     /**
-     * Install document-level delegated listeners for the chat textarea:
-     *  - `input` events auto-resize the textarea between 40px and 200px,
-     *  - `keydown` Enter (no Shift) dispatches a send,
-     *  - `click` on a `.chat-send-btn` dispatches a send.
-     *
-     * `sendCallback` is invoked with the textarea text on each send trigger.
-     * It is wired to `dioxus.send` by Rust so submissions return over the
-     * eval channel instead of through dioxus's broken `convert_form_data`
-     * path (#130). The callback is overwritten on every call so a re-mounted
-     * ChatInput points the listeners at the new eval's `dioxus.send`.
-     *
-     * Idempotent for the listener install; the callback rebind is intentional.
-     */
-    window.installChatSendBridge = function(sendCallback) {
-        window.__chatSendDispatch = sendCallback;
-        if (window.__chatSendBridgeInstalled) return;
-        window.__chatSendBridgeInstalled = true;
-
-        function fireSend() {
-            var el = document.querySelector('.chat-textarea');
-            if (!el || el.disabled) return;
-            var text = el.value;
-            if (!text || !text.trim()) return;
-            if (typeof window.__chatSendDispatch === 'function') {
-                window.__chatSendDispatch(text);
-            }
-        }
-
-        // Auto-resize on every keystroke.
-        document.addEventListener('input', function(e) {
-            var t = e.target;
-            if (t && t.classList && t.classList.contains('chat-textarea')) {
-                t.style.height = 'auto';
-                t.style.height = Math.min(Math.max(t.scrollHeight, 40), 200) + 'px';
-            }
-        });
-
-        // Enter (no Shift) inside the chat textarea sends.
-        document.addEventListener('keydown', function(e) {
-            var t = e.target;
-            if (t && t.classList && t.classList.contains('chat-textarea')
-                && e.key === 'Enter' && !e.shiftKey) {
-                e.preventDefault();
-                fireSend();
-            }
-        });
-
-        // Send button click sends. preventDefault stops the native form-submit
-        // path (button is type="submit") so we don't trigger any dioxus
-        // onsubmit listener — there shouldn't be one, but be defensive.
-        document.addEventListener('click', function(e) {
-            var t = e.target;
-            if (t && t.classList && t.classList.contains('chat-send-btn')) {
-                e.preventDefault();
-                fireSend();
-            }
-        });
-    };
-
-    /**
      * Trigger chart post-processing (mermaid + echarts) on the next animation frame.
      * Calls window.__processChatCharts if it has been defined by chart_processor.js.
      */
--- a/crates/ui/src/components/chat_panel/input.rs
+++ b/crates/ui/src/components/chat_panel/input.rs
@@ -1,14 +1,13 @@
 //! Chat input area: auto-resizing textarea + Send button.
 //!
-//! All keystrokes, Enter-to-send, and Send-button clicks are handled by a
-//! delegated JavaScript listener (`installChatSendBridge` in utils.js). Sends
-//! are funneled back to Rust via a long-lived `document::eval` channel.
-//!
-//! This bypasses dioxus's `onsubmit` / `onkeydown` / `oninput` event path,
-//! which trips the `.unwrap()` in dioxus-liveview-0.7.x's `convert_form_data`
-//! and `convert_keyboard_data` once the conversation accumulates DOM state
-//! (#130). A pleasant side effect is that we no longer pay a WebSocket round
-//! trip per keystroke for auto-resize.
+//! Uses native Dioxus events (`oninput`, `onkeydown`, `onclick`) for the send
+//! path, matching KubeStudio's pattern. This deliberately reverts the JS-bridge
+//! rerouting from #132 while keeping the per-bubble `onmounted` removal in
+//! `messages.rs` — that removal is what actually fixed #130 (panic in
+//! `dioxus-liveview` 0.7.x `convert_mounted_data`). Sending through
+//! `document::eval`'s reverse channel (`dioxus.send`) doesn't deliver on
+//! Windows/WebView2 when Pick is hosted inside StrikeHub's iframe, which
+//! silently broke chat send on the Windows MSI (see issue #189).
 
 use dioxus::prelude::*;
 
@@ -26,9 +25,7 @@ pub struct ChatInputProps {
 /// Auto-resizing textarea + Send button with Enter-to-submit behaviour.
 ///
 /// The textarea grows from a minimum of 40px up to 200px as the user types.
-/// Plain Enter sends the message; Shift+Enter inserts a newline. Both auto-
-/// resize and send dispatch happen entirely on the client; the only Rust
-/// involvement is receiving the submitted text via the eval channel below.
+/// Plain Enter sends the message; Shift+Enter inserts a newline.
 #[component]
 pub fn ChatInput(props: ChatInputProps) -> Element {
     let is_sending = props.is_sending;
@@ -36,7 +33,9 @@ pub fn ChatInput(props: ChatInputProps) -> Element {
     let disabled = is_sending() || agent_thinking();
     let on_send = props.on_send;
 
-    // Re-focus the textarea when agent finishes (disabled → enabled)
+    let mut input_text = use_signal(String::new);
+
+    // Re-focus the textarea when agent finishes (disabled → enabled).
     use_effect(move || {
         if !disabled {
             spawn(async move {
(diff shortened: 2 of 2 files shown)


cchao-devo (approved):
Review summary — PR #191 (revert chat send to native events for Windows/WebView2)
  
  Verdict: LGTM / Approve — with one honest caveat on what I could verify, and a nudge on the two reviewer sanity-checks.

  Verification caveat (be transparent with the author)

  I reviewed on macOS (aarch64) via a worktree. I could NOT do a full cargo check -p pentest-ui — the ui crate fails to build on macOS due to ~115 errors in the pentest-tools dependency, but I confirmed this 
  fails identically on clean main, so it's a pre-existing macOS-build limitation, not introduced by #191. My review of the Rust is therefore by inspection + API-pattern cross-check, not a local compile. (The
  author validated empirically on the actual Windows MSI, which is the platform that matters here.)
  
  What I verified

  - ✅ #130 fix preserved — messages.rs is untouched by this PR (empty diff-stat); it retains no per-bubble onmounted (only an explanatory comment remains). So the revert does not reintroduce #130.
  - ✅ installChatSendBridge fully removed — zero references anywhere in the crate; no dead JS bridge left behind.
  - ✅ All Dioxus event APIs are idiomatic + codebase-consistent — Event<KeyboardData>/FormData/MouseData, evt.key(), evt.modifiers().shift(), evt.value(), Key::Enter, prevent_default() all match existing usage
  in keyboard_shortcuts.rs, konami_code.rs, cyberchef_page.rs. Key is in scope via dioxus::prelude::*.
  - ✅ Send-path logic is correct: oninput tracks the signal; onkeydown sends on Enter && !shift (so Shift+Enter → newline, one of the sanity checks — handled correctly); button changed type="submit" →
  type="button" with onclick; onsubmit neutralized with prevent_default() to stop iframe navigation. The trim + empty-guard + clear-textarea logic is consistent across keydown and click.

  The load-bearing assumption (worth a nod, not a blocker)

  The whole revert rests on the author's root-cause reattribution: #130's panic was in convert_mounted_data (fixed by the messages.rs onmounted removal), not convert_form_data/convert_keyboard_data as #132
  believed — so restoring native oninput/onkeydown is safe. The evidence is strong (stack trace → events.rs:91, diff -r ruling out the version bump, KubeStudio using native events fine on Windows). It's
  well-argued and I have no reason to doubt it — but it is the assumption that, if wrong, reintroduces #130. That's exactly why the author's "history reload does not panic on macOS/Linux" sanity-check matters.
  
  My recommendation to the author

  Both unchecked boxes in the PR are the right gates and worth explicitly confirming before merge:
  - Shift+Enter inserts newline — I confirmed this is coded correctly (!evt.modifiers().shift()), but worth a live click.
  - History reload doesn't panic on macOS/Linux — this is the one that guards the root-cause assumption; worth an actual reload test on Linux/macOS since #130 was a reload-triggered crash.

  Bottom line: approve. Exceptionally well-investigated revert, #130 fix demonstrably preserved, clean removal of the dead bridge. My only asterisk is that I couldn't locally compile the ui crate (pre-existing
  macOS limitation, not this PR) — so I'd lean on the author's Windows-MSI validation + the two live sanity checks.

Issue #183, opened 2026-07-01
EPIC: Bring Linux and Windows Pick to parity (OS-aware capability gating)


jtomek-strike48:
# EPIC: Bring Linux and Windows Pick to parity (OS-aware capability gating; no silent-empty Windows stubs)

**Issue Type:** Epic
**Component:** Pick agent (`crates/core` platform enum + tool registry, `crates/tools`, `crates/platform` desktop backends)
**Effort:** Large (epic; multiple children)
**Priority:** P1
**Source:** Operator requirement - "bring Linux and Windows Pick to parity." Surfaced alongside a Windows wifi bug (red-team-038) and a Windows fabrication incident (red-team-036).
**Target repo:** `Strike48-public/pick`
**Children:** red-team-038 (Windows wifi detection bug). Further children to be split from the audit below.

> **Reality check (verified against code 2026-07-01 via trace):**
> - **There is NO per-platform (Linux vs Windows) capability gating.** All desktop OSes collapse to a single `Platform::Desktop` (`crates/core/src/tools.rs:12-20,37-44`, `Platform::current()`). `supported_platforms()` can only say "Desktop", never "Linux-only". Tool registration in `crates/tools/src/lib.rs` is unconditional except `TrafficCaptureTool` (gated on `is_pcap_available()`, `lib.rs:228` - a dependency check, not an OS check).
> - Consequence: **Windows advertises the full Linux tool surface to the agent**, and divergence surfaces only at runtime - mostly as **silent empty/success** or generic exec errors, never as a typed "unsupported on this platform."
> - Real OS branching lives in `crates/platform` behind trait objects. Windows silent-empty stubs: network interfaces return empty `Vec` (`crates/platform/src/desktop/system.rs:144-147`, `not(target_os="linux")`); DNS-resolver enrichment returns empty (`crates/tools/src/safety_check/threat_intel.rs:119`, "deferred"); nmap raw-socket privilege hardcoded `false` (`crates/tools/src/external/nmap.rs:788`).
> - Linux-only binaries invoked with no Windows fallback: `which` (should be `where`) in `crates/tools/src/catalog.rs:131` + installers; `iw`/aircrack suite (`system.rs:151`, `wifi_scan_detailed.rs`, `wifi_attack.rs`); `pth-winexe` (`lateral_movement.rs:200`); hardcoded `bash`/`/bin/bash` (`webwright/mod.rs:587,853`, `sidecar.rs:152`).
[...]

PR #201, opened 2026-07-04
fix(platform): probe binaries with where on Windows (#200, Child C of #183)

### LINKED ISSUE #200: Child C: OS-aware binary probing (which -> where) so external tools are detected on Windows
**Parent epic:** #183. **Child:** C.

## Problem

Binary-availability probing shells out to the Linux `which` command. On a Windows host `which` does not exist, so every probe returns non-zero and **every external tool reads as Missing/Unknown** in the catalog (surfaced in `tools_page.rs` / `settings_page.rs`). Windows has no `which`; the PATHEXT-aware equivalent is `where` (a real `where.exe` in System32; exit code 0 when found, matching the existing `exit_code == 0` checks).

## Scope (traced)

**12 host probes to fix:**
- 11 sites route through `CommandExec::execute_command("which", ...)` — all interpret `exit_code == 0` as installed: `catalog.rs:136`, `external/install.rs:45,99`, `webwright/install.rs:56`, `external/burp.rs:101`, `external/zap.rs:189`, `installers/{zap.rs:31,webwright.rs:62,bloodhound.rs:32,package.rs:79,metasploit.rs:30}`.
- 1 raw `tokio::process::Command::new("which")` bypassing the trait: `autopwn/crack.rs:497` (probes `hashcat`).

**Leave alone (sandbox-internal, Linux-only binaries):** `sandbox/bwrap.rs:43`, `sandbox/proot.rs:54`. And `pty_shell.rs:959` is a test assertion, not a probe.

## Design

The `which`/`where` choice depends on **where the command runs**, not just the host OS:
- Host-direct (sandbox disabled or fallback) on Windows -> `where`
- Host-direct on Linux/macOS -> `which`
- Inside the sandbox (WSL2 on Windows, proot/bwrap on Linux) -> always `which` (Linux env), correct today

So the fix must land at the **host-execution boundary**, not at the call sites (a naive `cfg!(windows)` helper would wrongly send `where` into the Linux WSL sandbox). Plan:

1. Add a single source of truth in `crates/platform` for the host probe command name (`which` on Unix, `where` on Windows).
2. In `crates/platform/src/desktop/command.rs::execute_command_direct` (host path only — the sandbox path sends `full_cmd` to the Linux sandbox manager and is untouched), translate the `which` probe to the host-appropriate command.
3. Fix the raw `crack.rs:497` probe to use the same source of truth.
4. Add unit coverage for the probe-name mapping (per-OS). Note: the filter-out negative path only runs on non-Linux CI — see #199.

## Acceptance criteria

- [ ] On Windows host, external-tool availability is detected via `where` (tools no longer all read Missing).
- [ ] Linux/macOS behavior unchanged (`which`).
- [ ] Sandbox-internal probing unchanged.
- [ ] Test covers the per-OS probe-command mapping.

jtomek-strike48:
Implements **Child C** of the Windows/Linux parity epic (#183). Closes #200.

## Problem

Binary-availability probing shells out to the Linux `which` command. On a Windows host `which` does not exist, so every probe returns non-zero and **every external tool reads as Missing/Unknown** in the catalog (surfaced in the Settings Tools panel). Windows ships `where.exe` — PATHEXT-aware, same exit-code contract (0 = found).

## Fix

- Add `common::host_which_command()` (in the always-compiled `common` module): returns `where` on Windows, `which` elsewhere — a single source of truth.
- Apply it at the **host-execution boundary** in `execute_command_direct`, which fixes all 11 probe sites that route through `CommandExec::execute_command` at once.
- Fix the one raw `Command::new("which")` probe in `autopwn/crack.rs` (probes `hashcat`) that bypasses the trait.

## Why the boundary matters

The `which`/`where` choice depends on *where the command runs*, not just the host OS. `execute_command_direct` is only reached on host paths (sandbox disabled, or fallback after sandbox failure). The sandbox *success* path sends commands into a Linux environment (WSL2 on Windows, proot/bwrap on Linux) via a different code path (`manager.execute()`), where `which` is correct — so it is deliberately untouched. Sandbox-internal `which` probes (`bwrap`/`proot` backend detection) are also left as-is.

All 11 probe callers check only `exit_code == 0` (none parse stdout), so the swap is behavior-preserving on Linux/macOS.

## Review

rust-reviewer pre-push pass: APPROVE, no CRITICAL/HIGH; independently verified the host-vs-sandbox boundary and all three design constraints. Its MEDIUM doc-clarity notes are applied in 36961f3 (fallback-path docstring; string-match + exit-code-only assumption).

## Test plan

- [x] `cargo fmt --all` clean
- [x] `cargo test -p pentest-platform` — new `host_which_command_matches_target_os` passes
- [ ] CI green on push (Dependency Audit will fail on the pre-existing transitive `quick-xml` RUSTSEC-2026-0194/-0195, unrelated to this PR; ignore lands via #190)

### DIFF
--- a/crates/platform/src/common.rs
+++ b/crates/platform/src/common.rs
@@ -9,6 +9,25 @@ use std::sync::Arc;
 use std::time::Duration;
 use tokio::sync::Semaphore;
 
+/// The command that reports whether a binary is on the host's PATH.
+///
+/// Unix (Linux/macOS/BSD) uses `which`; Windows has no `which` but ships
+/// `where.exe`, which is PATHEXT-aware and, like `which`, exits 0 when the
+/// binary is found and non-zero otherwise. Using the wrong one makes every
+/// external tool read as "Missing" on Windows. See GitHub issue #183.
+///
+/// This describes the command for the **host** only. Commands sent into the
+/// sandbox (WSL2 on Windows, proot/bwrap on Linux) always run in a Linux
+/// environment where `which` is correct, so the sandbox path must not use this.
+#[must_use]
+pub const fn host_which_command() -> &'static str {
+    if cfg!(target_os = "windows") {
+        "where"
+    } else {
+        "which"
+    }
+}
+
 /// Resolve `host:port` to a [`SocketAddr`], trying a direct parse first and
 /// falling back to DNS resolution via [`ToSocketAddrs`].
 fn resolve_addr(addr: &str, port: u16) -> SocketAddr {
@@ -172,4 +191,21 @@ pub fn parse_ip_neigh(output: &str) -> Vec<ArpEntry> {
             })
         })
         .collect()
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn host_which_command_matches_target_os() {
+        // The host binary-probe command must be `where` on Windows (no `which`
+        // there) and `which` on every Unix-like target. See #183.
+        let cmd = host_which_command();
+        if cfg!(target_os = "windows") {
+            assert_eq!(cmd, "where");
+        } else {
+            assert_eq!(cmd, "which");
+        }
+    }
 }
--- a/crates/platform/src/desktop/command.rs
+++ b/crates/platform/src/desktop/command.rs
@@ -130,12 +130,13 @@ pub async fn execute_command_in_dir(
 
 /// Direct command execution (used as fallback or for internal operations).
 ///
-/// This runs on the **host**, so the Linux-only `which` binary-probe is
+/// This always runs on the **host**, so the Linux-only `which` binary-probe is
 /// rewritten to the host-appropriate command (`where` on Windows) via
-/// [`crate::common::host_which_command`]. The sandbox path in
-/// [`execute_command_in_dir`] does NOT come through here — it sends the command
-/// into a Linux environment (WSL2/proot/bwrap) where `which` is correct — so the
-/// rewrite is confined to host execution. See GitHub issue #183.
+/// [`crate::common::host_which_command`]. The sandbox *success* path does not
+/// come through here — it sends the command into a Linux environment
+/// (WSL2/proot/bwrap) where `which` is correct. A sandbox *failure* does fall
+/// back to this function (see [`execute_command_in_dir`]), but that fallback
+/// executes on the host, so the rewrite is still correct. See GitHub issue #183.
 pub(crate) async fn execute_command_direct(
     cmd: &str,
     args: &[&str],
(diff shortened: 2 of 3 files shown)


cchao-devo (approved):
Review summary — PR #201 (probe binaries with where on Windows)
  
  Verdict: LGTM / Approve. Clean, correctly-scoped fix with the right architectural instinct (single source of truth at the host-execution boundary). Verified locally.

  What I verified (not just read)

  - ✅ Compiles on macOS (pentest-platform + pentest-tools), new test host_which_command_matches_target_os passes, clippy clean with -D warnings.
  - ✅ All 11 trait-routed probe sites pass the bare literal "which" (catalog, burp, install×2, zap×2, package, webwright×2, metasploit, bloodhound) — so the cmd == "which" interception in execute_command_direct
  catches every one. No probe uses a path/variant (/usr/bin/which, a variable) that the string-match would silently miss.
  - ✅ The crack.rs raw Command::new("which") (the one that bypasses the trait) is correctly fixed directly via host_which_command().
  - ✅ Sandbox which probes correctly left alone — bwrap.rs/proot.rs run inside the Linux sandbox env where which is right; untouching them matches the stated host-vs-sandbox design.

  The design is right

  The key insight — that which/where depends on where the command runs, not the host OS — is correct and well-reasoned. Intercepting at execute_command_direct (the host boundary) rather than at each of the 11
  call sites is the right DRY choice, and deliberately not touching the sandbox-success path (which sends into Linux/WSL2) is the subtle-but-correct decision. The docstrings added in the follow-up commit explain
  this clearly.
  
  Behavior-preserving on Linux/macOS confirmed: host_which_command() returns "which" on both, so the swap is a no-op there; only Windows sees where.

  Minor / optional (non-blocking)

  - The string-match cmd == "which" is an implicit contract: it works because all callers pass the literal, but nothing enforces that a future probe site does. Very low risk (the pattern is consistent and the
  comment documents it), but if you wanted belt-and-suspenders, a tiny helper like probe_binary(name) on the exec trait would make the contract explicit rather than relying on a magic-string intercept. Purely
  optional — the current approach is fine and minimal.
  - where prints multiple paths / different stdout format than which — the PR correctly notes all callers check only exit_code == 0 (none parse stdout), so this is safe. I confirmed that's true across the 11
  sites. Good that it's called out.

  Note on claims
  
  The "rust-reviewer APPROVE" and the exit-code-contract / host-vs-sandbox boundary claims all hold under independent check. The Dependency-Audit CI failure (quick-xml RUSTSEC) is pre-existing/unrelated (#190).

  Bottom line: approve. No blocking issues; one optional design note (explicit probe helper vs. magic-string intercept).

Issue #185, opened 2026-07-01
Bug: Windows wifi detection ignores selected adapter and fails silently


jtomek-strike48:
# Bug: Windows wifi detection ignores selected adapter and fails silently (empty success)

**Issue Type:** Bug
**Component:** Pick agent (`crates/platform` desktop system backend, `crates/tools` wifi tool, `crates/ui` settings)
**Effort:** Medium
**Priority:** P2
**Source:** Operator report - wifi detection broken on Windows.
**Target repo:** `Strike48-public/pick`
**Parent epic:** red-team-037 (Linux/Windows parity). This is a concrete instance of that epic's silent-empty pattern.
**Related:** red-team-036 (fail-closed guardrail) - the silent empty here is exactly the failure that lets the agent fabricate "networks in range."

> **Reality check (verified against code 2026-07-01 via trace):**
> - **Pick has no first-party Windows wifi code.** All OS branching in `crates/platform/src/desktop/system.rs` is `#[cfg(target_os="linux")]` vs `#[cfg(not(...))]`. Windows falls through to the generic path.
> - `get_wifi_networks(interface)` (`system.rs:394`): the interface-aware branch `scan_specific_interface` (`system.rs:152`, shells `iw`/`nmcli`/`ip`, parses `iw scan` text) is `#[cfg(target_os="linux")]` and is only *called* under a Linux cfg (`system.rs:396`). **On Windows it is compiled out entirely.** So Windows always takes the auto-detect branch: `wifi_scan::scan()` (`system.rs:402-427`), which enumerates via the `win32-wlan` Native WiFi backend.
> - **The user-selected adapter is silently ignored on Windows.** `crates/tools/src/wifi_scan.rs:39` loads `settings.wifi_adapter` and logs it (`:42`) but it only feeds the Linux `scan_specific_interface`; on Windows there is no code path that honors it. The `win32-wlan` backend also just uses `interfaces.first()` - so only the first enumerated WLAN interface is ever scanned (wrong device for USB pentest adapters).
[...]

PR #203, opened 2026-07-04
fix(platform): recognize Windows WiFi adapter names case-insensitively (#185, part C)

jtomek-strike48:
Fixes **Part C** of #185 (Child E of the Windows/Linux parity epic #183).

## Bug

`check_wifi_connection_status` matched adapter names with a case-sensitive `contains("wi-fi")` / `contains("wireless")`. Windows names its primary adapter **"Wi-Fi"** (capital W) — which matched none of the branches (`to_lowercase().contains("wifi")` also missed it, because "wi-fi" has a hyphen). Result on Windows: `total_adapters=0`, `connected_via_wifi=false`, `safe_to_scan=true` — a false negative while actually connected over WiFi. That is exactly the silent-empty / false-safe hazard EPIC #183 targets.

## Fix

- Promote the `is_wifi_name` closure to a free function (now unit-testable).
- Lowercase the name once and match all patterns against the lowercased form, so "Wi-Fi", "Wi-Fi 2", and "Wireless Network Connection" are recognized. Behavior-preserving for every previously-matching name (`wlan*`, `wlp*`, `wl*`, `en0`, `wifi`, `*mon`).
- Fix a latent `let _ = interface;` (param is `_interface`) in the compiled-out `not(feature="wifi_scan")` arm of `get_wifi_networks`.

## Scope note (why only Part C)

There is no Windows runner in CI (only ubuntu + macOS), so `#[cfg(target_os="windows")]` code is compiled/run by neither local dev nor CI. Part C is pure, cross-platform-testable string logic, so it ships now. Parts **A** (honor the operator-selected adapter on Windows) and **B** (distinguish scan-failure from genuinely-empty) require a new `netsh`/Native WiFi scanner — the `wifi_scan` crate v0.7.0 exposes no interface selection and hardcodes `interfaces.first()`. Writing that unvalidated on a protected branch would risk shipping bugs, so A/B are deferred to a Windows-hardware session and tracked on #185.

## Review

rust-reviewer pre-push pass: **CLEAN, no findings**. Verified behavior preservation, no false positives (tested Ethernet/eth0/lo/docker0/vEthernet/Bluetooth/Local Area Connection/en1), doc-comment placement, and the `_interface` fix.

## Test plan

- [x] `cargo fmt --all` clean
- [x] `cargo test -p pentest-platform` — 3 new `is_wifi_name` tests pass (Windows/Linux/macOS/negative)
- [ ] CI green on push (Dependency Audit will fail on the pre-existing transitive `quick-xml` RUSTSEC-2026-0194/-0195, unrelated to this PR; ignore lands via #190)

### DIFF
--- a/crates/platform/src/desktop/system.rs
+++ b/crates/platform/src/desktop/system.rs
@@ -428,11 +428,31 @@ pub async fn get_wifi_networks(_interface: Option<String>) -> Result<Vec<WifiNet
 
     #[cfg(not(feature = "wifi_scan"))]
     {
-        let _ = interface; // Suppress unused warning
+        let _ = _interface; // Suppress unused warning (param is `_interface`)
         Err(Error::PlatformNotSupported(
             "WiFi scanning requires the 'wifi_scan' feature".into(),
         ))
     }
+}
+
+/// Whether a network-interface name looks like a WiFi adapter.
+///
+/// Matches Linux (`wlan0`, `wlp2s0`, monitor-mode `wlan0mon`), macOS (`en0`),
+/// and Windows (`Wi-Fi`, `Wireless Network Connection`) naming. All matching is
+/// done case-insensitively: Windows names the primary adapter "Wi-Fi" (capital
+/// W), which the previous case-sensitive `contains("wi-fi")` missed, causing
+/// `check_wifi_connection_status` to report 0 adapters / `safe_to_scan = true`
+/// on Windows even while connected over WiFi. See GitHub issue #185.
+fn is_wifi_name(name: &str) -> bool {
+    let n = name.to_lowercase();
+    n.starts_with("wlan")
+        || n.starts_with("wlp")
+        || n.starts_with("wl")
+        || n.starts_with("en0") // macOS WiFi (usually)
+        || n.contains("wi-fi") // Windows "Wi-Fi"
+        || n.contains("wireless") // Windows "Wireless Network Connection"
+        || n.contains("wifi")
+        || n.ends_with("mon") // monitor-mode interfaces (e.g. wlan0mon)
 }
 
 /// Check WiFi connection status for scan safety assessment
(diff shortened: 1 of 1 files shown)


cchao-devo (approved):
Verdict: LGTM / Approve — the fix is correct and the false-safe bug it closes is real and important. One nuance to raise on the "behavior-preserving" claim, and one pre-existing heuristic worth noting (neither
  blocking).

  Verified locally (macOS)
  
  - ✅ 3 new is_wifi_name tests pass, clippy clean -D warnings (this is pure cross-platform string logic, so macOS fully exercises it — unlike the ui crate).
  - ✅ The core fix is right: Windows "Wi-Fi" (capital W) previously matched nothing → 0 adapters, safe_to_scan=true while actually on WiFi. That's a textbook false-safe (the epic's target hazard) — lowercasing
  once and matching against that fixes it. Good, well-motivated catch.
  - ✅ Promoting the closure to a free fn makes it unit-testable; the _interface unused-var fix in the not(feature="wifi_scan") arm is a legit tidy-up.

  One nuance on "behavior-preserving for every previously-matching name" (MEDIUM — worth a note, not a block)

  The claim is that the change only adds case-insensitive matching. But lowercasing the whole name before the starts_with checks can create new matches that the old case-sensitive code rejected — i.e. it's not
  purely behavior-preserving, it's behavior-expanding. Example: a name like "WLAN AutoConfig" (or any adapter/service name starting with capital "WLAN") — old code: starts_with("wlan") on the raw string → no 
  match; new code: to_lowercase() → "wlan autoconfig" → starts_with("wlan") → now matches.
  
  For this function's purpose (detecting you're on WiFi → fail-safe toward "don't scan"), expanding the match set is the safe direction — a false positive means "we think you're on WiFi, be cautious," which errs
  correctly. So it's not a bug. But the PR body says "behavior-preserving for every previously-matching name," which is technically true but understates it — it also newly matches some previously-unmatched names.
  Worth the author rewording to "expands matching case-insensitively; the only directional change is more names now recognized as WiFi, which is the fail-safe direction." Minor accuracy point.
  
  Pre-existing heuristic worth noting (not introduced here)

  starts_with("en0") treats macOS en0 as WiFi — but on many Macs with Ethernet/Thunderbolt, en0 is Ethernet and WiFi is en1. The negative test even asserts en1 is not WiFi, which is wrong on that hardware. This
  is a pre-existing heuristic (#203 just moves it verbatim into the testable fn), so not a blocker for this PR — but since #203 is now codifying it in tests, it's a good moment to flag: the en0/en1 assumption is
  fragile and might deserve a follow-up (proper interface-type detection rather than name-prefix guessing). Track on #185/#183 if worth it.
  
  Scope note is sound

  Deferring Parts A/B (operator-selected adapter, scan-failure-vs-empty — both need a netsh/Native WiFi scanner the wifi_scan crate can't do) to a Windows-hardware session is the right call; shipping unvalidated
  Windows-only code on a protected branch would be worse. Part C being pure testable logic justifies shipping now.

  Bottom line: approve. Correct fix for a real false-safe bug, well-tested. Two optional notes: (1) reword "behavior-preserving" → "case-insensitively expanding, in the fail-safe direction"; (2) the en0-is-WiFi
  heuristic is fragile and now test-codified — candidate for a follow-up.

PR #196, opened 2026-07-04
feat(tools): OS-aware capability gating for desktop tools (#195, Child A of #183)

jtomek-strike48:
Implements **Child A** of the Windows/Linux parity epic (#183): an OS-aware capability model so a Linux-only tool is not advertised to the Strike48 agent on a Windows host. Closes #195.

## Problem

All desktop OSes collapse into a single `Platform::Desktop`, so the registry and agent cannot express "this tool is Linux-only." A Linux-only tool (aircrack-ng / hashcat backed) was therefore advertised on Windows, where it can only fail at runtime — or worse, return a plausible empty. `is_supported()` was only exercised in tests, so the advertise path filtered nothing.

Parity here means **honesty, not universal support**: the agent must know per-host which capabilities are real.

## Approach (additive; Linux behavior preserved exactly)

Rather than splitting the `Platform` enum (touches ~50 call sites and the serialized wire surface), add an orthogonal desktop-OS axis:

- `enum DesktopOs { Linux, MacOS, Windows, Other }` + `DesktopOs::current()` (returns `None` on wasm/Android/iOS).
- `ToolSchema.supported_os` (serde default = every desktop OS) + `.os(...)` builder.
- `PentestTool::supported_os()` trait method defaulting to every desktop OS.
- `is_supported()` additionally checks the current `DesktopOs` when on desktop.
- `ToolRegistry::supported_schemas()` / `supported_names()` — host-filtered.
- Advertise paths switched to the filtered variants: `connector.rs` (build_task_types/build_metadata), `liveview_connector` (registration metadata **and** the session tool-name list that feeds the agent's auto-approved `tool_configs`).
- First real consumers: `autopwn` crack and `wifi_scan_detailed` declared `[DesktopOs::Linux]`.

## Wire safety

`supported_os` is **not** serialized onto the Matrix wire (`to_json_schema()` is hand-built and omits it), and pre-existing serialized schemas still deserialize (serde default). The catalog dependency-collection path stays intentionally unfiltered (install/inventory surface across all OSes) — documented in code.

## Review

Reviewed with the rust-reviewer agent before opening. Findings addressed in commit 9d80469: HIGH (session tool-name list now uses `supported_names()`), MEDIUM (catalog unfiltered-by-design comment), LOW (BSD/`Other` test branch).

## Test plan

- [x] `cargo fmt --all` clean
- [x] `cargo clippy --workspace --locked --features pentest-platform/desktop-pcap -- -D warnings` clean
[...]

PR #205, opened 2026-07-04
fix: gate Linux-only WiFi attacks and Pass-the-Hash on non-Linux (#204, Child D of #183)

### LINKED ISSUE #204: Child D: gate Linux-only pentest binaries on non-Linux (WiFi attacks, Pass-the-Hash)
**Parent epic:** #183. **Child:** D (slice D1 — runtime honesty for Linux-only binaries).

## Problem

Two tool paths shell out to Linux-only binaries with no honest failure on non-Linux hosts:

- **WiFi attacks** (`crates/platform/src/desktop/wifi_attack.rs`, `WifiAttackOps for DesktopPlatform`): the whole impl (13 methods: `iw`/`airmon-ng`/`airodump-ng`/`aireplay-ng`/`aircrack-ng`) was compiled on all desktop OSes and, on Windows/macOS, failed with generic `ToolExecution("... is it installed?")` or silently no-op'd. These need raw-radio access (monitor mode, injection) that neither macOS nor Windows exposes, and — unlike offline `.cap` cracking — they cannot route through WSL2 (no USB radio passthrough).
- **Pass-the-Hash** (`crates/tools/src/lateral_movement.rs`, `try_pass_the_hash`): when `pth-winexe` (Linux passing-the-hash toolkit) could not be spawned, it returned `Ok(success: false, "pth-winexe not available")` — a **silent success** indistinguishable from a real auth rejection. The call site also swallowed errors with `if let Ok(result)`.

## Fix (slice D1 — buildable + validated on Linux/macOS CI today, no Child A dependency)

- `wifi_attack.rs`: gate the real impl `#[cfg(target_os = "linux")]` (byte-for-byte unchanged) and add a `#[cfg(not(target_os = "linux"))]` stub impl where every method returns `Error::PlatformNotSupported` with a clear "requires Linux raw radio; not available under WSL" message. Add `#[cfg(not(target_os="linux"))]` unit tests (run by macOS CI).
- `lateral_movement.rs`: on spawn failure return a typed `PlatformNotSupported` (distinguished from a genuine `success:false` auth rejection), and change the call site to propagate with `?` (execute_timed turns it into a failed ToolResult) instead of swallowing it.

Validated both cfg arms compile (Linux directly; non-Linux stub + tests via a temporary cfg-swap since no Windows/macOS target is installed locally). CI's `Check (macOS)` compiles and runs the stub path.

## Deferred (tracked, not in this slice)

- **D2 (needs Child A #196):** add `supported_os=[Linux]` advertise-gating to the remaining Linux-only tools (`autopwn/capture`, `external/wireless/aircrackng`, `wifi_scan`, `lateral_movement`).
- **D3 (needs Windows+WSL hardware):** validate that offline `aircrack-ng .cap` cracking and webwright actually route through WSL2; codify "offline crypto -> WSL, live-radio -> PlatformNotSupported."

## Acceptance criteria

- [ ] WiFi-attack ops on non-Linux return typed `PlatformNotSupported`, not generic/silent failure.
- [ ] Pass-the-Hash with a missing binary fails loudly (typed error), not `Ok(success:false)`.
- [ ] Linux behavior byte-for-byte unchanged.
- [ ] Tests cover the non-Linux honest-failure path.

jtomek-strike48:
Implements **Child D** (slice D1) of the Windows/Linux parity epic #183. Closes #204.

## Problem

Two tool paths shell out to Linux-only binaries and, on non-Linux hosts, failed dishonestly:

- **WiFi attacks** (`wifi_attack.rs`, `WifiAttackOps`): the whole `iw`/`airmon-ng`/`airodump`/`aireplay`/`aircrack-ng` impl compiled on all desktop OSes and, on Windows/macOS, produced generic `ToolExecution("... is it installed?")` errors or silent no-ops.
- **Pass-the-Hash** (`lateral_movement.rs`): a missing `pth-winexe` returned `Ok(success: false, "not available")` — a **silent success** indistinguishable from a real auth rejection, further hidden by an `if let Ok(...)` that swallowed errors.

These need raw-radio access (WiFi) or the Linux PTH toolkit; unlike offline `.cap` cracking, they cannot route through WSL2 (no USB radio passthrough), so the honest outcome is a typed error.

## Fix

- `wifi_attack.rs`: gate the real impl `#[cfg(target_os = "linux")]` (byte-for-byte unchanged) and add a `#[cfg(not(target_os = "linux"))]` stub impl where all 13 methods return `Error::PlatformNotSupported` via a shared `unsupported()`. `#[cfg(not(target_os="linux"))]` unit tests assert the contract (run by `Check (macOS)`).
- `lateral_movement.rs`: `try_pass_the_hash` returns `PlatformNotSupported` when `pth-winexe` can't be spawned (distinct from a genuine `success:false` auth failure); the call site propagates with `?` (inside `execute_timed`, so it fails just this technique honestly) instead of swallowing it.

Linux behavior is byte-for-byte unchanged.

## Validation note

There is no Windows runner in CI and no macOS/Windows target installed locally, so I validated **both cfg arms** by a temporary cfg-swap that compiled+tested the non-Linux stub (and its 2 tests) on the Linux host, then reverted. `Check (macOS)` compiles and runs the stub path in CI.

## Review

rust-reviewer pre-push pass: **CLEAN, ready to merge**. Independently verified cfg exclusivity (exactly one impl per target), all 13 stub signatures vs the trait, import gating, and that the `?` change doesn't abort other technique branches.

## Deferred (tracked on #183)

- **D2** (needs #196 merged): `supported_os=[Linux]` advertise-gating for the remaining Linux-only tools.
- **D3** (needs Windows+WSL hardware): validate offline `aircrack-ng`/webwright routing through WSL2.

## Test plan

- [x] `cargo fmt --all` clean
- [x] Linux build + both cfg arms validated; 2 non-Linux stub tests pass (via cfg-swap; run natively on macOS CI)
- [ ] CI green on push (Dependency Audit fails on pre-existing transitive `quick-xml` RUSTSEC-2026-0194/-0195, unrelated; clears via #190)

### DIFF
--- a/crates/platform/src/desktop/wifi_attack.rs
+++ b/crates/platform/src/desktop/wifi_attack.rs
@@ -1,14 +1,27 @@
 //! WiFi attack operations for desktop Linux (aircrack-ng suite)
+//!
+//! These operations drive the Linux-only `iw` / `airmon-ng` / aircrack-ng
+//! suite and require raw-radio access (monitor mode, packet injection). They
+//! are not available on macOS or Windows — and, unlike offline `.cap`
+//! cracking, they cannot be routed through the WSL2 sandbox because WSL has no
+//! USB WiFi radio passthrough. On non-Linux targets every method fails loudly
+//! with [`Error::PlatformNotSupported`] rather than a generic "is it
+//! installed?" error or a silent no-op. See GitHub issue #183.
 
 use crate::traits::*;
 use async_trait::async_trait;
 use pentest_core::error::{Error, Result};
-use std::process::Stdio;
-use std::time::Duration;
-use tokio::process::Command;
 
 use super::DesktopPlatform;
 
+#[cfg(target_os = "linux")]
+use std::process::Stdio;
+#[cfg(target_os = "linux")]
+use std::time::Duration;
+#[cfg(target_os = "linux")]
+use tokio::process::Command;
+
+#[cfg(target_os = "linux")]
 #[async_trait]
 impl WifiAttackOps for DesktopPlatform {
     async fn enable_monitor_mode(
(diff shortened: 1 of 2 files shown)


cchao-devo (approved):
Review summary — PR #205 (gate Linux-only WiFi attacks + Pass-the-Hash on non-Linux)
  
  Verdict: LGTM / Approve. Clean, well-scoped, correctly fixes a real silent-failure. Verified by local compilation on the actual non-Linux (macOS) target — the arm CI can't easily exercise.

  What I verified (not just read)

  - ✅ Non-Linux stub arm compiles natively on macOS (cargo check -p pentest-platform, aarch64-apple-darwin) — this is the check CI (Linux) can't do and the author could only validate via a temporary cfg-swap. It
  genuinely compiles.
  - ✅ Both stub tests pass on macOS (crack_wep/enable_monitor_mode → PlatformNotSupported).
  - ✅ clippy clean with -D warnings on the platform crate, macOS arm.
  - ✅ tools crate compiles with the PTH change; Error::PlatformNotSupported variant exists (core/src/error.rs:21).
  - ✅ Trait/stub parity: WifiAttackOps = 13 methods, stub impls exactly 13 (the "15" is +2 test fns).
  - ✅ "Linux impl byte-for-byte unchanged" — confirmed; the only - lines are use relocations gaining #[cfg], no method bodies touched.

  The core fix is correct

  - PTH silent-success fix (real bug): the old Err(_) → Ok(success:false, "not available") was a textbook silent success — a can't-spawn-binary indistinguishable from an auth rejection, further hidden by if let 
  Ok(...). Now returns typed PlatformNotSupported. Good catch, right fix.
  - I double-checked the ?-abort concern in the multi-target loop and it's safe: try_pass_the_hash returns Err only when the binary can't spawn (a host-level condition uniform across targets); a per-target auth
  rejection comes back as Ok(success:false) and does NOT abort the loop. So aborting on the missing-binary error is correct, not a multi-target regression. Worth confirming the author reasoned this through — but
  the code is right.

  Minor / optional (non-blocking)

  - Test coverage is 2 of 13 stub methods. The 2 chosen (enable_monitor_mode, crack_wep) are representative bookends and all 13 share the same one-line Err(unsupported()), so this is fine — but a one-line note in
  the PR that the pattern is uniform (so 2 suffices) would preempt the question.
  - lateral_movement.rs has no non-Linux gating — unlike wifi_attack, PTH isn't #[cfg]-gated; it relies on the runtime pth-winexe spawn failing → PlatformNotSupported. That's a reasonable choice (the binary's
  absence is the honest signal on any OS), just an asymmetry with the wifi approach worth noting is intentional.

  Note on the PR's own claims

  The body's "rust-reviewer pre-push: CLEAN" and cfg-exclusivity claims all hold up under my independent check. The Dependency-Audit CI failure (quick-xml RUSTSEC) is genuinely pre-existing/unrelated (clears via
  #190) — not introduced here.

  Bottom line: I'd approve. No blocking issues; the two minor notes are optional polish.

PR #141, opened 2026-06-02
feat: parallel webwright, StrikeKit artifacts, LLM proxy auth

josh-devo:
## Summary

Clean rebased branch (16 commits) with novel work on top of main.

- Parallel webwright execution (per-task sidecar, per-task LLM conversation)
- StrikeKit artifact upload via InvokeCapabilityRequest
- LLM proxy auth from platform session tokens
- nmap --unprivileged when not root
- File browser back button fix
- Clippy/compilation fixes

## Changes

- **Parallel webwright**: per-task sidecar spawn, no global singleton
- **LLM proxy**: per-task conversations, bearer token auth passthrough
- **StrikeKit uploads**: screenshots/scripts uploaded after webwright completion
- **session_token**: process-global RwLock (no env var leakage)
- **param_u64**: parses string values from agent tool calls
- **Timeout**: default 600s, returns 10s before deadline to avoid circuit breaker
- **nmap**: --unprivileged when euid != 0

## Test plan
- [ ] `nix develop --command cargo check -p pentest-core`
- [ ] Run webwright from StrikeKit — verify screenshots produced
- [ ] Parallel webwright calls spawn separate sidecars
- [ ] nmap works without root

PR #215, opened 2026-07-08
fix(platform): gate proot mode() check to unix so Windows builds compile (#214)

### LINKED ISSUE #214: Windows build fails: PermissionsExt::mode() called without cfg(unix) gate in pty_shell.rs
## Summary

The v0.1.7 release run [28901816211](https://github.com/Strike48-public/pick/actions/runs/28901816211) failed on both Windows jobs (Desktop and Headless) with:

```
error[E0599]: no method named `mode` found for struct `Permissions` in the current scope
  --> crates\platform\src\desktop\pty_shell.rs:192:46
   |
192|                     .map(|m| m.permissions().mode() & 0o111 != 0)
   |                                              ^^^^ method not found in `Permissions`
```

## Root cause

`Permissions::mode()` comes from `std::os::unix::fs::PermissionsExt`, which is Unix-only. The import at the top of `crates/platform/src/desktop/pty_shell.rs` is already correctly gated:

```rust
#[cfg(unix)]
use std::os::unix::fs::PermissionsExt;
```

However, the executability check added at line 191-193 is **not** gated behind `cfg(unix)`, so Windows targets fail to compile.

The offending block was introduced in commit faefa8b (\"feat: parallel webwright, StrikeKit artifacts, LLM proxy auth\").

## Impact

- Windows Desktop and Windows Headless builds fail
- The release pipeline never reaches \"Create Release\", so no Windows artifacts are published for the tag
- All other targets (macOS x86_64/aarch64, Linux x86_64, Web) build cleanly

## Fix direction

Gate the mode-bit check behind `#[cfg(unix)]` and use a cross-platform fallback (e.g. `proot_path.exists()`) on Windows. The `SandboxBackend::Proot` branch isn't a realistic runtime path on Windows, but the code still has to compile there.

## Failed jobs

- Desktop (Windows): https://github.com/Strike48-public/pick/actions/runs/28901816211/job/85777115000
- Headless (Windows): https://github.com/Strike48-public/pick/actions/runs/28901816211/job/85777128218

taylor-wood-devo:
## Summary

- Closes #214.
- Windows Desktop and Headless builds have been failing with `E0599: no method named 'mode' found for struct 'Permissions'` at `crates/platform/src/desktop/pty_shell.rs:192`.
- `PermissionsExt::mode()` is Unix-only. The block introduced in faefa8b was not `cfg(unix)`-gated, so the Windows targets no longer compile — v0.1.7 shipped without Windows artifacts.
- This PR gates the mode-bit check behind `#[cfg(unix)]` and falls back to `Path::exists()` on Windows. The `SandboxBackend::Proot` branch isn't a realistic runtime path on Windows, but the code still has to compile there.

## Test plan
- [x] `cargo fmt --all`
- [x] `cargo check --all-targets` (aarch64-apple-darwin) — clean
- [x] `cargo check --target x86_64-pc-windows-gnu -p pentest-platform` — clean (this reproduces the CI failure locally before the fix)
- [x] `cargo clippy --all-targets -- -D warnings` — clean
- [x] `cargo clippy --all-targets --target x86_64-pc-windows-gnu -p pentest-platform -- -D warnings` — clean
- [x] `cargo test --lib --bins` — 601 passed, 9 ignored (expected per CLAUDE.md)
- [ ] CI passes on all platforms, especially Desktop (Windows) and Headless (Windows)

### DIFF
--- a/crates/platform/src/desktop/pty_shell.rs
+++ b/crates/platform/src/desktop/pty_shell.rs
@@ -192,6 +192,10 @@ impl PtyShell {
                 let is_executable = std::fs::metadata(&proot_path)
                     .map(|m| m.permissions().mode() & 0o111 != 0)
                     .unwrap_or(false);
+                // Windows: SandboxBackend::Proot is not a runtime target here (no proot
+                // binary is shipped), but this branch must still compile. Fall back to a
+                // plain existence probe rather than re-implementing an executable check
+                // via Windows-specific APIs.
                 #[cfg(not(unix))]
                 let is_executable = proot_path.exists();
                 tracing::info!(
--- a/crates/platform/src/desktop/pty_shell.rs
+++ b/crates/platform/src/desktop/pty_shell.rs
@@ -188,9 +188,12 @@ impl PtyShell {
                 let proot_path = ProotExecutor::get_proot_path(config)
                     .await
                     .map_err(|e| Error::ToolExecution(format!("Proot binary not found: {e}")))?;
+                #[cfg(unix)]
                 let is_executable = std::fs::metadata(&proot_path)
                     .map(|m| m.permissions().mode() & 0o111 != 0)
                     .unwrap_or(false);
+                #[cfg(not(unix))]
+                let is_executable = proot_path.exists();
                 tracing::info!(
                     "[PtyShell] proot binary: path={} exists={} executable={}",
                     proot_path.display(),


cchao-devo (approved):
PR #215 Review — ✅ Approve

  "fix(platform): gate proot mode() check to unix so Windows builds compile (#214)" — Taylor Wood, +7/−0, 1 file.

  What it does: gates the Unix-only Permissions::mode() exec-bit check behind #[cfg(unix)], with a Path::exists() fallback on #[cfg(not(unix))], so the SandboxBackend::Proot branch compiles on Windows.

  Verified correct:
  - ✅ Root cause accurate — the offending m.permissions().mode() block was introduced by faefa8b (#141) — the same SDK-migration commit behind #217 and #219. That commit added an ungated Unix-only call,
  breaking Windows compile (v0.1.7 shipped without Windows artifacts). This is the third Windows/portability regression traced to #141.
  - ✅ PermissionsExt import is already #[cfg(unix)]-gated (line 9-10) — so no unused-import failure on Windows. The fix pairs correctly with the existing import gate.
  - ✅ No other Unix-only calls in the file — grep confirms mode() at line 192 was the sole Windows blocker; this fully fixes the compile.
  - ✅ Semantically safe — is_executable feeds only a tracing::info! log field, gating no control flow. So the Windows fallback's weaker check (exists() vs. exec-bit) has zero functional impact, and only
  on Windows where the Proot branch isn't a real runtime path. The comment says exactly this. Correct call not to re-implement Windows ACL exec-detection for a log line.
  - ✅ Test plan is thorough — includes the cross-compile reproduction (cargo check --target x86_64-pc-windows-gnu) that reproduced the CI failure before the fix, plus clippy on both targets. Exactly the
  right verification for a portability fix. 
  
  No findings. Minimal, correct, well-tested, well-commented. This is a clean approve — it unblocks Windows Desktop/Headless builds.

PR #222, opened 2026-07-10
ci(release): use arduino/setup-protoc on Windows so protoc install can't silently 504 (#221)

### LINKED ISSUE #221: Windows release job silently skips protoc install when Chocolatey feed 504s
## Summary

The v0.9.9 release run [29060198475](https://github.com/Strike48-public/pick/actions/runs/29060198475) failed on both Windows jobs (Desktop and Headless) at the `strike48-proto` build script with:

```
error: failed to run custom build command for `strike48-proto v0.4.4`
...
Error: Custom { kind: NotFound, error: "Could not find `protoc`. If `protoc` is installed, try setting the `PROTOC` environment variable to the path of the `protoc` binary. Try installing `protobuf-compiler` or `protobuf` using your package manager. It is also available at https://github.com/protocolbuffers/protobuf/releases  For more information: https://docs.rs/prost-build/#sourcing-protoc" }
```

This is distinct from #214 (E0599 `mode()` on Windows). The build never reached `pty_shell.rs` this time — `prost-build` couldn't find `protoc` in the first place.

## Root cause

`.github/workflows/release.yml` installs `protoc` on Windows via:

```yaml
- name: Install protoc
  run: choco install protoc -y
```

On the failing run, Chocolatey's community feed returned 504 Gateway Timeout:

```
Failed to fetch results from V2 feed at
  'https://community.chocolatey.org/api/v2/Packages(Id='protoc',Version='35.1.0')'
  Response status code does not indicate success: 504 (Gateway Timeout).
...
Chocolatey installed 0/0 packages.
```

Two combined flaws:

1. **Chocolatey feed is a SPOF.** `community.chocolatey.org` is a third-party service with no SLA for our release pipeline. When it's degraded, we can't cut a Windows release.
2. **`choco install -y` exits 0 even when it installs 0 packages.** The workflow step is marked ✓ Success despite installing nothing, so the failure isn't surfaced until much later at the `prost-build` step, and the error message points at a missing `PROTOC` env var rather than the actual upstream cause.

## Impact

- Windows Desktop and Windows Headless builds fail on any tag push whenever `community.chocolatey.org` is degraded.
- v0.9.9 shipped without Windows artifacts (release job depends on both).
- Failure mode is silent-until-cargo: Install protoc step reports success, then the build fails ~4 minutes later with a misleading \"Could not find protoc\" error.

## Fix direction

Swap `choco install protoc -y` for `arduino/setup-protoc@v3` on both `build-desktop-windows` and `build-headless-windows`:

```yaml
- name: Install protoc
  uses: arduino/setup-protoc@v3
  with:
    version: \"27.x\"
    repo-token: \${{ secrets.GITHUB_TOKEN }}
```

Why this fixes it:

- Pulls the release tarball directly from `github.com/protocolbuffers/protobuf/releases` — no dependency on `community.chocolatey.org`.
- Writes the bin to `\$GITHUB_PATH`, so `protoc` is on PATH in the subsequent Build step.
- Fails hard on download errors, so the workflow surfaces the real problem immediately rather than at cargo-build time.
- Cross-platform, so we can optionally consolidate Linux/macOS to the same action later (out of scope for this fix).

## Failed jobs

- Desktop (Windows): https://github.com/Strike48-public/pick/actions/runs/29060198475/job/86260190730
- Headless (Windows): https://github.com/Strike48-public/pick/actions/runs/29060198475/job/86260190717

## Related

- #214 / PR #215 — separate Windows compile issue in `pty_shell.rs`. Both fixes must land before the next tag will produce Windows artifacts.

taylor-wood-devo:
## Summary

- Closes #221.
- v0.9.9 release run [29060198475](https://github.com/Strike48-public/pick/actions/runs/29060198475) failed on both Windows jobs because `choco install protoc -y` hit `community.chocolatey.org` 504 Gateway Timeout, installed nothing, and still exited 0. The failure only surfaced ~4 minutes later at `prost-build` time as a misleading "Could not find `protoc`" error.
- This PR swaps the chocolatey step on both `build-desktop-windows` and `build-headless-windows` for `arduino/setup-protoc@v3`, which:
  - pulls the release tarball straight from `github.com/protocolbuffers/protobuf/releases` (no chocolatey feed dependency),
  - writes the bin to `$GITHUB_PATH` so `protoc` is available in the next Build step,
  - fails hard on download errors so the workflow can't silently proceed to a broken build.

## Test plan

- [x] Diff limited to `.github/workflows/release.yml` — no code changes.
- [ ] CI: verify both Windows jobs now find `protoc` and reach the Sign / Package steps.
- [ ] Companion to #215 — both must land before the next tag will produce Windows artifacts.

## Related

- #214 / #215 — separate Windows compile issue in `crates/platform/src/desktop/pty_shell.rs`.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -64,7 +64,10 @@ jobs:
       - uses: Swatinem/rust-cache@v2
 
       - name: Install protoc
-        run: choco install protoc -y
+        uses: arduino/setup-protoc@v3
+        with:
+          version: "27.x"
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
 
       - name: Build
         run: cargo build --locked --package pentest-desktop --release
@@ -347,7 +350,10 @@ jobs:
       - uses: Swatinem/rust-cache@v2
 
       - name: Install protoc
-        run: choco install protoc -y
+        uses: arduino/setup-protoc@v3
+        with:
+          version: "27.x"
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
 
       - name: Build
         run: cargo build --locked --package pentest-headless --release


cchao-devo (approved):
PR #222 Review — ✅ Approve

  "ci(release): use arduino/setup-protoc on Windows so protoc install can't silently 504 (#221)" — Taylor Wood, +8/−2, .github/workflows/release.yml only.

  What it does: replaces choco install protoc -y with arduino/setup-protoc@v3 on the two Windows jobs (build-desktop-windows, build-headless-windows).

  Root cause is well-diagnosed and the fix is correct:
  - ✅ Real failure mode: choco install hit a chocolatey-feed 504 and exited 0 (installed nothing but reported success) — a classic silent-failure that only surfaced 4 min later as a misleading "Could
  not find protoc" at prost-build. Swapping to arduino/setup-protoc (pulls from GitHub releases, fails hard on download error) both removes the flaky chocolatey dependency and eliminates the
  silent-success footgun. This is exactly the right instinct — fail loud at the install step, not cryptically at build. 
  - ✅ Scoped correctly: only the two runs-on: windows-latest jobs are changed. Linux uses apt-get protobuf-compiler, macOS uses brew install protobuf — both unaffected and appropriate for their
  platforms.
  - ✅ repo-token: ${{ secrets.GITHUB_TOKEN }} is the recommended usage for arduino/setup-protoc (avoids GitHub API rate-limiting when fetching the release) — correct.
  - ✅ YAML shape is right: the run: step correctly becomes a uses: + with: step; indentation and placement (before the Build step) are sound.

  Two minor, non-blocking observations (worth a comment, not a block):

  1. Version pin 27.x isn't cross-checked against the other platforms. Linux/macOS install unpinned distro/brew protoc (whatever's current), while Windows now pins 27.x. protoc is highly wire-compatible
  so this won't break prost-build, but there's a mild inconsistency — the fleet isn't on a single protoc version. Not worth blocking (prost only needs a working protoc for codegen), but if the team wants
  reproducible builds, pinning all three the same would be cleaner. Optional follow-up.
  2. Can't fully verify green — the test plan's CI checkboxes are unticked (verify both Windows jobs reach Sign/Package). That's inherent to a release-workflow change (only a tagged release run exercises
  it), and the PR correctly notes it's a companion to #215 — both must land before the next tag produces Windows artifacts. So final proof comes from the next release run, which is acceptable.

  No blocking findings. Minimal, correct, fixes a genuine silent-failure in CI, properly scoped. Approve.

PR #228, opened 2026-07-13
fix(#224): route chat input through JS bridge on macOS/Linux

### LINKED ISSUE #224: Standalone Pick (macOS) aborts on first chat keystroke: convert_form_data unwrap-on-None
## Symptom

Running the standalone `pentest-connector` binary on macOS against a Strike48 host: opening the Chat panel, clicking into the textarea, and pressing any character key (repro'd with `h`) immediately aborts the process. The panic is inside `dioxus-liveview` — `Option::unwrap()` on `None` at `events.rs:62` (`convert_form_data`).

Reproduced on:

- Branch: `fix/223-standalone-pick-auth` (off current `main`, `370fe6c`)
- Platform: macOS Darwin 25.3.0, standalone binary via `./run-pentest.sh headless dev`
- Target: `wss://non-prod.strike48.test:443`
- `dioxus-liveview = 0.7.3` per `Cargo.lock`

Log tail before abort:

```
INFO  pentest_core::matrix::client: [gql] response status=200 OK
INFO  pentest_ui::components::chat_panel: ChatPanel: updated agent tools for pentest-connector
WARN  pentest_ui::components::chat_panel: JS eval failed (utils.js init): EvalError::Finished - eval has already ran
WARN  pentest_ui::components::chat_panel: JS eval failed (chart processor init): EvalError::Finished - eval has already ran
WARN  pentest_ui::components::chat_panel::messages: JS eval failed (scroll listener install): EvalError::Finished - eval has already ran

thread 'main' panicked at .../dioxus-liveview-0.7.3/src/events.rs:62:14:
called `Option::unwrap()` on a `None` value
Abort trap: 6
```

## Root-cause pointer

`events.rs:62` in `dioxus-liveview-0.7.3` is inside `convert_form_data`:

```rust
fn convert_form_data(&self, event: &PlatformEventData) -> FormData {
    event
        .downcast::<SerializedFormData>()
        .cloned()
        .unwrap()          // ← line 62
        .into()
}
```

This fires when Pick's chat textarea receives an `oninput` event. See `crates/ui/src/components/chat_panel/input.rs:52-62` — `oninput: on_input` is bound native, so every keystroke routes an `Event<FormData>` through this converter.

## History (this is a re-emergence of the class of bug #132 originally worked around)

- **#130** — panic at `events.rs:91` (`convert_mounted_data`) when loading conversation history. Fixed by removing per-bubble `onmounted` listeners.
- **#132 / f0e5455** — bundled two changes: (a) the per-bubble `onmounted` removal above, and (b) rerouting `oninput` / `onkeydown` / `onclick` through a JS bridge (`dioxus.send` reverse channel). The bridge's stated purpose was to guard against `convert_form_data` / `convert_keyboard_data` panicking once chat DOM state accumulated.
- **#189 / #191 / da4f6e1** — reverted the JS bridge because it silently broke chat send on Windows/WebView2 hosted inside StrikeHub's iframe. Kept the `onmounted` removal (#130 stays fixed). The revert commit claimed `convert_form_data` / `convert_keyboard_data` weren't the real crash sites — but this current report is direct evidence that they are, at least on macOS standalone.

Relevant excerpt from **#189**'s issue body:

> Pick had to move off native events in commit `f0e5455` (\"route chat panel off dioxus-liveview broken event converters, #130\") because dioxus-liveview 0.7's `convert_form_data` / `convert_keyboard_data` panics once chat DOM state accumulates. KubeStudio uses dioxus-liveview 0.6 and doesn't hit that panic. Net: the earlier fix traded one bug for another, and the second bug is Windows-only.

That prediction has landed — reverting to native events fixed Windows but re-broke macOS standalone.

## Version-bump note

Per da4f6e1's commit message, `dioxus-liveview-0.7.9` (what KubeStudio ships) has byte-identical sources to 0.7.3 for `events.rs`, so a simple version bump to 0.7.9 does nothing here.

## Impact

- Standalone Pick on macOS: chat is unusable — first keystroke aborts the whole process.
- StrikeHub-hosted Pick on Windows: unaffected (the #191 fix stands, `document::eval` reverse channel is no longer relied on for send).
- Blocks maintainer dev flow (`./run-pentest.sh` against a remote studio).

Suggests P1.

## Suggested resolution direction (not prescriptive)

1. **Scoped JS bridge, platform-conditional.** Reinstall the `installChatSendBridge` from utils.js only when running outside WebView2 (i.e. non-Windows, or user-agent gate) — keeps native events for the iframe/Windows case that #191 solved.
2. **Patch dioxus-liveview locally.** Add a `[patch.crates-io]` fork that turns `convert_form_data` / `convert_keyboard_data`'s `.unwrap()` into `.unwrap_or_default()` — the event silently drops instead of aborting. Upstream may or may not accept, but the local patch is cheap.
3. **Drop `oninput` entirely.** Auto-resize via a JS input listener (already how `utils.js` does other things), then read `.chat-textarea` value from JS at send time on `onclick` / `onkeydown`. No per-keystroke Rust round trip. `onkeydown` still routes through `convert_keyboard_data` — if that panics too, gate via option 2.

## Reproduction

```bash
git checkout main               # or fix/223-standalone-pick-auth
./run-pentest.sh headless dev   # against a reachable Strike48 host
# In the UI: dismiss any auth prompt, open Chat, click textarea, press any letter.
```

Owner: I'll pick this up. Cross-refs: #130, #132, #189, #191.

taylor-wood-devo:
Closes #224.

## Summary

Standalone Pick on macOS aborts on the first chat keystroke: dioxus-liveview 0.7.x's `convert_form_data` unwraps a `None` downcast whenever `oninput` fires (events.rs:62). #132 originally worked around this with a document-level JS bridge, but #189/#191 reverted it because `document::eval`'s reverse channel (`dioxus.send` -> `eval.recv`) does not deliver in the WebView2 + LiveView + iframe combination StrikeHub uses on Windows. Restoring the bridge unconditionally would re-break Windows chat send.

This PR splits `ChatInput` into two `cfg`-gated implementations sharing the same `ChatInputProps` signature, so each host OS gets the path that has been verified to work there.

## Changes

- **`crates/ui/src/components/chat_panel/input.rs`** now has two implementations of `ChatInput`:
  - `cfg(any(target_os = "macos", target_os = "linux"))` uses the JS bridge. The textarea binds no `oninput`, so `convert_form_data` is never invoked. Enter and Send-button clicks are caught by delegated document listeners and flushed back to Rust via the eval reverse channel, which is known to work on WebKit (macOS) and WebKitGTK (Linux).
  - `cfg(not(...))` (Windows, Android, iOS, and any other target) keeps the native-events implementation byte-for-byte, preserving #191's fix for StrikeHub-hosted Pick on Windows/WebView2 and leaving Android on the path it currently ships with.
- **`crates/ui/src/assets/utils.js`** reinstates `installChatSendBridge` as the platform-neutral half of the bridge (idempotent listener install, callback rebind on re-mount). Loaded on every platform, wired to Rust only on the macOS/Linux path.

## Non-regression coverage

| Host | Path | Status |
|------|------|--------|
| Standalone macOS | JS bridge | Fixes #224 (verified locally) |
| Standalone Linux | JS bridge | Same converter class, same fix |
| StrikeHub-hosted macOS (WebKit) | JS bridge | Was fine on #132 pre-#191, no known regression |
| StrikeHub-hosted Linux (WebKitGTK) | JS bridge | Same |
| Standalone Windows | Native events | Unchanged from `main` |
| StrikeHub-hosted Windows (WebView2 iframe) | Native events | Unchanged from `main`, preserves #191 |
| Android / iOS / other | Native events | Unchanged from `main` |

## Test plan

- [x] `cargo fmt --all --check`
- [x] `cargo check --all-targets`
- [x] `cargo clippy --workspace --locked --features pentest-platform/desktop-pcap -- -D warnings`
- [x] `cargo test --workspace --locked --features pentest-platform/desktop-pcap --lib` (623 passed, 9 ignored, 0 failed)
- [x] Manual macOS standalone: `./run-pentest.sh headless dev`, open Chat, click textarea, type text, Enter sends, Shift+Enter inserts newline, Send button submits, textarea auto-resizes between 40px and 200px. No abort.
- [ ] Manual Windows StrikeHub-hosted (WebView2 iframe): code path is byte-identical to `main`, so covered by existing #191 verification.

## Stacking

Based on `fix/223-standalone-pick-auth` (#225) because a working standalone auth flow is required to reach the chat panel and reproduce the panic. Once #225 merges, this PR will retarget to `main`.

Refs: #130, #132, #189, #191.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -85,8 +85,14 @@ jobs:
         with:
           fetch-depth: 0
       - name: Install gitleaks
+        env:
+          GH_TOKEN: ${{ github.token }}
         run: |
-          GITLEAKS_VERSION=$(curl -s https://api.github.com/repos/gitleaks/gitleaks/releases/latest | grep '"tag_name"' | sed 's/.*"v\(.*\)".*/\1/')
+          # `gh api` gives us a proper JSON parse via --jq so we don't have to
+          # regex our way through GitHub's compact JSON, which used to catch
+          # the whole `"tag_name":"v8.30.1","draft":...` blob and produce a
+          # malformed download URL.
+          GITLEAKS_VERSION=$(gh api repos/gitleaks/gitleaks/releases/latest --jq '.tag_name' | sed 's/^v//')
           curl -sSfL "https://github.com/gitleaks/gitleaks/releases/download/v${GITLEAKS_VERSION}/gitleaks_${GITLEAKS_VERSION}_linux_x64.tar.gz" | tar xz
           sudo mv gitleaks /usr/local/bin/
       - run: gitleaks detect --source . --verbose
--- a/crates/core/src/config.rs
+++ b/crates/core/src/config.rs
@@ -351,6 +351,45 @@ impl ConnectorConfig {
         }
     }
 
+    /// Read the tenant UUID the SDK stored during OTT approval, if any.
+    ///
+    /// Studio addresses App-behavior connectors by tenant UUID
+    /// (`/#/apps/matrix%3A<uuid>%3A<connector>%3A...`). The connector's
+    /// initial pending-approval registration can carry either a tenant
+    /// slug or a UUID — the server accepts both — but Studio only routes
+    /// app-viewer traffic to the identity that matches its URL. When the
+    /// operator types the slug, the server-issued JWT is minted against
+    /// the canonical UUID and the SDK writes it to
+    /// `~/.strike48/credentials/<connector>_<instance_id>.json`. Reading
+    /// that file at connect time lets us reuse the UUID for subsequent
+    /// registrations so the slug→UUID gap is invisible to the operator
+    /// (pick#223).
+    ///
+    /// StrikeHub does the equivalent server-side: it pre-resolves the
+    /// user's tenant via a `userDetails { domain.id }` GraphQL query
+    /// (`strikehub/crates/sh-core/src/auth.rs::fetch_tenant_id`) and
+    /// injects the UUID into every spawned connector's env. Standalone
+    /// Pick has no authenticated context up front, so we rely on the
+    /// post-OTT credentials file instead.
+    ///
+    /// Returns `None` when the file is missing, unreadable, or the
+    /// `tenant_id` field is absent — callers should fall back to the
+    /// user-supplied tenant string.
+    pub fn read_credentials_tenant_id(connector_name: &str, instance_id: &str) -> Option<String> {
+        let home = std::env::var("HOME").ok()?;
+        let path = std::path::PathBuf::from(home)
+            .join(".strike48")
+            .join("credentials")
+            .join(format!("{connector_name}_{instance_id}.json"));
+        let content = std::fs::read_to_string(&path).ok()?;
+        let value: serde_json::Value = serde_json::from_str(&content).ok()?;
+        value
+            .get("tenant_id")
+            .and_then(|v| v.as_str())
+            .map(|s| s.trim().to_string())
+            .filter(|s| !s.is_empty())
+    }
+
     /// Split a Strike48-style URL (accepts anything [`normalize_host`] produces
     /// plus bare hosts) into `(hostname, port)`.
     ///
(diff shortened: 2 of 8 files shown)


taylor-wood-devo:
Addressing the review nits:

**Fixed — Additional finding #1 (textarea height reset on macOS/Linux path)**

Real bug, valid catch. `send_message` in `mod.rs:657` calls `clearTextarea` which sets `el.value = ''` but doesn't touch `style.height`, and programmatic value writes don't fire an `input` event so the delegated auto-resize listener at `utils.js:154-161` never ran. A multi-line send left the textarea visually expanded until the next keystroke.

Fix in 9eee591: `fireSend()` in `utils.js` now zeroes the value and resets `style.height='40px'` synchronously right before invoking `__chatSendDispatch`, mirroring the native path's inline clear in `input.rs`.

**Not addressing — Additional finding #2 (Windows StrikeHub unable-to-verify)**

Reviewer's stated concern is that the document-level `input`/`keydown`/`click` listeners will still register on Windows because `installChatSendBridge` is loaded on every platform. That's not the case: `installChatSendBridge` is a function *definition*; the listeners only register when it is *called*, and the call lives inside `use_hook` gated by `#[cfg(any(target_os = "macos", target_os = "linux"))]` in `input.rs:71-87`. On Windows the helper is loaded but never invoked, so no listeners register and the native events path is untouched. Byte-for-byte equivalence with `main` on Windows holds.

**Not addressing — Axis 7 (screen recording) and Axis 8 (tests)**

Screen recording is a convention item; happy to attach a Loom after review if wanted. Reviewer explicitly acknowledged the tests axis as "defensible for a JS-event fix that can only be observed under a real WebView" — the project has no headless-browser test infrastructure that could exercise this path.

edbonddevo (approved):

PR #227, opened 2026-07-13
fix(run): restore ~/.strike48 key ownership after sudo launches (#226)

jtomek-strike48:
Closes #226 (child of parity epic #183).

## Problem

WiFi hardware access needs `sudo`, so several launch paths run the whole connector as root with `HOME` preserved (`sudo -E ... HOME=$HOME`). The SDK then writes its `0600` private key + credentials into the operator's `~/.strike48` owned by `root`. A later **non-sudo** launch can no longer read that key, so OTT registration fails 3x with `Failed to read private key: Permission denied (os error 13)` and the agent sits in `WaitingForApproval`. The key/credential pair is intact and valid - it is just unreadable by the current uid, so the "re-issue approval in Studio" hint is misleading.

## Fix

- **New `scripts/restore-key-ownership.sh`** - idempotent, no-op when not root; restores `~/.strike48` (+ `~/.pick`) ownership to `$SUDO_USER`. Honors `$STRIKE48_KEYS_DIR`.
- **Wired as an `EXIT` trap** into every sudo launch path: `run-pentest.sh` and just recipes `run-headless-sudo`, `run-headless-dev`, `run-desktop-sudo`, `run-desktop-release-sudo`. Covers Ctrl+C. Previously `run-pentest.sh` only chowned `~/.pick`, not `~/.strike48` where keys live.
- **`RUNNING.md`** documents the symptom, the immediate `sudo scripts/restore-key-ownership.sh` fix, and the `setcap cap_net_raw,cap_net_admin` no-sudo alternative for raw-socket scans.

## Test plan

- [x] `bash -n` clean on helper + `run-pentest.sh`; `just --list` parses all edited recipes
- [x] Non-root invocation is a no-op (exit 0, no output)
- [x] Reproduce break (chown key to root) -> run helper as root with `SUDO_USER` -> key restored to user and readable
- [x] Idempotent: second run is a quiet no-op, no stray root-owned files remain
- [ ] CI green on push

## Notes

Complements capability-gating epic #183: same "silent operator footgun" family. An optional upstream follow-up (noted in #226) is for `strike48-connector` to emit a clearer message than bare os-error-13 when the key is owned by another user.

PR #230, opened 2026-07-14
fix(nmap): accept a `flags` string arg and steer the schema away from it

jtomek-strike48:
## Problem

The agent called nmap with a raw-CLI-style payload:

```json
{"flags": "-sn -T4 -PS22,80,443,445,3389 -PA80,443 -PE -PP", "target": "10.0.8.0/22"}
```

There is no `flags` parameter in the nmap schema, and because a valid `target` was present, `normalize_legacy_nmap_params` hit its target-present early-return and **dropped the entire `flags` string** - so `-sn` (ping) and `-T4` (timing) never took effect. The scan silently ran with defaults, and the malformed shape contributed to the stuck engagement turn we saw in the field.

## Fix

- **Fold a `flags` string into structured params BEFORE the target-present short-circuit**, so recognized flags apply even when `target` is already set. The raw `flags` key is always consumed - never passed through or shelled out.
- **Extract the flag-token allowlist into a shared `apply_flag_tokens` helper** reused by the existing `args`/`command` path and the new `flags` path (DRY). A `flags` string with no separate `target` can still recover the target token.
- **Tighten the tool description** to tell the model to use the structured params (`scan_type`/`timing`/`service_detection`/`no_ping`/...) and that there is no `flags`/`args`/`command` param - reducing malformed calls at the source.

## Safety

Unchanged from the existing legacy path: recovered targets go through `validate_target` and are passed as a **structured positional arg** (never shell-interpolated). A token with shell metacharacters (`;`, space) never validates as a target, and bare integers are never treated as targets. Unrecognized flags (`-PS`/`-PA`/`-PE`/`-PP`/`--min-rate`) are dropped.

## Test plan

- [x] RED verified: 3 new tests fail on the old code (flags dropped when target present)
- [x] 5 new tests: flags+target maps recognized flags; flags-only recovers target+ports; flags don't clobber explicit params; injection payload never becomes target; `;`-bearing token rejected
- [x] Full nmap suite **51/51**, `security_tests` **55/55** pass
- [x] `cargo fmt` + `cargo clippy -p pentest-tools --lib` clean
- [x] CI green (Build amd64/arm64, Check Linux+macOS, Clippy, Test, Dependency Audit)

---

## Full investigation context (for the dev team)

This PR is one piece of an end-to-end investigation into the nmap "Running… / Generating…" hang during an engagement phase. The hang had **two independent root causes**, plus adjacent bugs found along the way. Recording the whole picture here so it's in one place.

### The nmap hang: two root causes

1. **Permissions - agent granted zero connector tools.** `StrikekitCore.PhaseExecutor.build_tools_config/2` looked up the connector-tool registry by `ctx.tenant.name` ("non-prod"), but connectors register under the tenant **UUID** prefix (`"<uuid>.<connector_type>.<instance>"`). The name key matched **0** connectors -> `connectors: {}` -> the agent got only `sk_*`/`document_*` system tools, never nmap. Live proof: `list_all_tools("non-prod")` -> 0 tools vs `list_all_tools("<uuid>")` -> 115 (incl. nmap).
   **Status: already fixed on matrix `develop`** (`discover_pentest_connector/2` queries both name and UUID). A redundant standalone PR (matrix#3131) was closed as superseded.

2. **nmap argument shape (this PR).** Even once granted, the model emits `{"flags": "...", "target": "..."}`; Pick dropped the flags. Fixed here.

### Still open / needs verification

- **Client-tool dispatch / consent bridge (matrix#2941).** A granted client tool_call can still park `pending` if the agent-framework->connector dispatch/consent path isn't wired. Partially addressed on `develop` (#2482 consent, #2524 scope enforcement). Needs a fresh end-to-end test on a current-`develop` studio build. Engagement agents use `consent_mode:"auto"`, which may bypass the manual-consent gap.

### Adjacent bugs found & fixed (same investigation)

| Item | Cause | Status |
|------|-------|--------|
| Pick OTT `Permission denied (os error 13)` | prior `sudo` run left `~/.strike48` key root-owned; non-sudo launch can't read it | Fixed - Pick #227 (self-healing chown) + setcap |
| "Seed N/M failed" | `extract_host` didn't strip URL path -> SSRF validator DNS-resolved `host/path` -> mislabeled "private IP" | Fixed - Pick #229 |
| alpaca "no response" | agent model alias `smart` -> `bedrock/qwen.qwen3-next-80b-a3b-v1:0`, which **Bedrock rejects** ("invalid model identifier") - verified unavailable in this account/region even with the correct `-v1:0` suffix (`qwen3-32b-v1:0` and `qwen3-coder-30b-a3b-v1:0` DO work) | Live-patched (configmap -> claude-sonnet-4-5); **durable init-dev fix ON HOLD** - init-dev defaults `ALPACA_FLAVOR=qwen` and its qwen `smart`/`reasoning` aliases point at the unavailable model (also missing the `-v1:0` suffix) |
| "Invalid or expired OTT" | one-time registration token in `.env` consumed; saved cred was for a different tenant | Resolved (operator) - post-approval flow; stale cred moved aside |

### Environment note

The deployed studio image is `matrix-studio:2026-06-13-develop-df78a33` (June 12), `imagePullPolicy: IfNotPresent`, so a restart re-pulls the same stale image. The permissions fix reaches the running pod only via an image rebuild (`task newbuild:build:studio`), which is **blocked**: the zot step decrypts ECR pull credentials (SOPS/age) to a recipient the operator's SSH key isn't. **Dev-team action:** re-encrypt `dev-secrets/secrets/dev-vm/ecr-pull-credentials.yaml` to the operator's age recipient, or provision the authorized key. (Testing was unblocked via an in-memory hot-patch of the permissions fix into the running BEAM node - ephemeral, reverts on restart.)

### DIFF
--- a/crates/tools/src/external/nmap.rs
+++ b/crates/tools/src/external/nmap.rs
@@ -496,14 +496,15 @@ async fn validate_nse_scripts<P: CommandExec>(
 /// target can be recovered, the original params are returned so the normal
 /// "target parameter is required" error still fires.
 fn normalize_legacy_nmap_params(params: Value) -> Value {
-    // A `flags` string (e.g. {"flags": "-sn -T4 ...", "target": "..."}) is a
-    // shape the LLM emits by analogy to a raw CLI, but there is no `flags`
-    // schema param. Fold its recognized flags into structured params FIRST —
-    // this must run even when `target` is already present (otherwise the
-    // target-present short-circuit below would silently drop the flags, which
-    // is exactly the bug where -sn/-T4 never took effect). Same allowlist +
-    // drop-unknown safety as the `args`/`command` path.
-    let params = fold_flags_string(params);
+    // A `flags` value (e.g. {"flags": "-sn -T4 ...", "target": "..."} or the
+    // array shape {"flags": ["-sn", "-T4"], "target": "..."}) is a shape the LLM
+    // emits by analogy to a raw CLI, but there is no `flags` schema param. Fold
+    // its recognized flags into structured params FIRST — this must run even
+    // when `target` is already present (otherwise the target-present
+    // short-circuit below would silently drop the flags, which is exactly the
+    // bug where -sn/-T4 never took effect). Same allowlist + drop-unknown safety
+    // as the `args`/`command` path.
+    let params = fold_flags(params);
 
     // Already in the structured shape — nothing more to do.
     if params
@@ -517,19 +518,7 @@ fn normalize_legacy_nmap_params(params: Value) -> Value {
     // Collect the flag/argument tokens from either `args` (array or string) or
     // a `command` string (with a leading "nmap" stripped).
     let tokens: Vec<String> = match params.get("args") {
-        Some(Value::Array(items)) => items
-            .iter()
-            .filter_map(|v| v.as_str().map(|s| s.to_string()))
-            .collect(),
-        Some(Value::String(s)) => {
-            // The args string may itself be a JSON-encoded array, e.g.
-            // "[\"-sn\", \"10.0.8.0/22\"]" — parse that first, else split on
-            // whitespace.
-            match serde_json::from_str::<Vec<String>>(s) {
-                Ok(parsed) => parsed,
-                Err(_) => s.split_whitespace().map(|t| t.to_string()).collect(),
-            }
-        }
+        Some(v @ (Value::Array(_) | Value::String(_))) => tokenize_flag_value(v),
         _ => match params.get("command").and_then(|v| v.as_str()) {
             Some(cmd) => cmd
                 .split_whitespace()
@@ -547,12 +536,16 @@ fn normalize_legacy_nmap_params(params: Value) -> Value {
     // Start from any structured params the caller DID provide so we don't clobber them.
     let mut out = params.as_object().cloned().unwrap_or_default();
 
-    // Map recognized flags onto `out`, recover ports, and return any target the
-    // tokens contained (shared with the `flags`-string path).
-    let target = apply_flag_tokens(&mut out, &tokens);
+    // Map recognized flags to a fresh `derived` map and recover any target the
+    // tokens contained (shared with the `flags` path). Merge derived values
+    // ONLY into gaps — an explicit structured param the caller set always wins.
+    let (derived, target) = apply_flag_tokens(&tokens);
+    merge_fill(&mut out, derived);
 
     match target {
         Some(t) => {
+            // No valid `target` reached here (the target-present short-circuit
+            // above returned already), so the recovered token is the target.
             out.insert("target".into(), json!(t));
             Value::Object(out)
         }
@@ -562,38 +555,46 @@ fn normalize_legacy_nmap_params(params: Value) -> Value {
     }
 }
 
-/// Fold a `flags` STRING param into structured scan params.
+/// Fold a `flags` param into structured scan params.
 ///
 /// Callers (the LLM generalizing from a raw CLI) sometimes pass
-/// `{"flags": "-sn -T4 ...", "target": "..."}`. There is no `flags` schema
-/// param, so without this the flags are silently ignored (and previously, when
-/// `target` was present, dropped entirely by the early return). This tokenizes
-/// `flags` (whitespace-split), maps the recognized allowlist onto structured
-/// params via the shared [`apply_flag_tokens`], recovers a target from the
-/// flags string if the caller didn't provide one, and removes the raw `flags`
-/// key so it is never passed through or shelled out. Unrecognized flags are
-/// dropped, exactly like the `args`/`command` path.
+/// `{"flags": "-sn -T4 ...", "target": "..."}` or the array form
+/// `{"flags": ["-sn", "-T4"], "target": "..."}` — the same string-or-array shape
+/// the sibling `args` path already tolerates. There is no `flags` schema param,
+/// so without this the flags are silently ignored (and previously, when
+/// `target` was present, dropped entirely by the early return — the exact bug
+/// this PR kills, which an array-shaped `flags` would otherwise reintroduce).
 ///
-/// Non-string or absent `flags` is a no-op.
-fn fold_flags_string(params: Value) -> Value {
-    let Some(flags) = params.get("flags").and_then(|v| v.as_str()) else {
+/// This tokenizes `flags` via the shared [`tokenize_flag_value`], maps the
+/// recognized allowlist onto structured params via [`apply_flag_tokens`],
+/// recovers a target from the flags tokens if the caller didn't provide one,
+/// and ALWAYS removes the raw `flags` key (for any shape — string, array, or an
+/// unexpected object) so it is never passed through or shelled out. Derived
+/// values fill only gaps: an explicit structured param the caller set wins over
+/// a flag (matching the tool description, which makes the typed params
+/// authoritative). Unrecognized flags are dropped, exactly like the
+/// `args`/`command` path.
+///
+/// Absent `flags` is a no-op.
+fn fold_flags(params: Value) -> Value {
+    if params.get("flags").is_none() {
         return params;
-    };
-
-    let tokens: Vec<String> = flags.split_whitespace().map(|t| t.to_string()).collect();
-    if tokens.is_empty() {
-        // Still strip the (empty/whitespace) flags key.
-        let mut out = params.as_object().cloned().unwrap_or_default();
-        out.remove("flags");
-        return Value::Object(out);
     }
 
     let mut out = params.as_object().cloned().unwrap_or_default();
-    out.remove("flags");
+    // Strip the raw `flags` key regardless of its shape — a `flags` value that
+    // is neither a string nor an array (e.g. an object) must NOT ride through
+    // to `execute()`, which would ignore it and silently scan with defaults.
+    let flags = out.remove("flags").unwrap_or(Value::Null);
 
-    let recovered = apply_flag_tokens(&mut out, &tokens);
+    let tokens = tokenize_flag_value(&flags);
+    if tokens.is_empty() {
+        return Value::Object(out);
+    }
 
-    // Only adopt a target from the flags string if the caller didn't set one.
+    let (derived, recovered) = apply_flag_tokens(&tokens);
+
+    // Only adopt a target from the flags tokens if the caller didn't set one.
     let has_target = out
         .get("target")
         .and_then(|v| v.as_str())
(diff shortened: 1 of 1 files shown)


ngdevo:
## PR #230 — Accept a `flags` string arg and steer the schema away from it

PR #230 Review — fix(nmap): accept a flags string arg and steer the schema away from it

Reviewed PR #230 (fix(nmap): accept a flags string arg and steer the schema away from it). Single-file,
+180/-29. Focused, well-tested. Findings below.

---
CRITICAL

None.

---
HIGH

H1 — flags string silently overrides explicit structured params (clobber contradicts test name)
crates/tools/src/external/nmap.rs — apply_flag_tokens writes every recognized flag onto out unconditionally
via .insert(...). If a caller passed service_detection: false explicitly and then also flags: "-sV", -sV
flips service_detection to true. The PR's own test flags_does_not_clobber_explicit_structured_params (line
1259-ish) asserts exactly this clobber — the name promises "does not clobber" but the assertions prove
service_detection: false → true. This is confusing at minimum and likely wrong behavior:

- The stated principle "last-writer = flags, matching how a real nmap invocation would win" doesn't hold — this
isn't a real invocation; the caller passed the structured param on purpose. The flags string is a legacy
shape being tolerated, not authoritative.
- From an operator-safety perspective, a flags: "-A" (aggressive scan) will now silently enable OS detection,
service detection, script scan, and traceroute even if the operator's structured params say otherwise. -A
is scope-expanding.
- Same applies to -sn, -sS, -sT, -sU, -O, -A, -T<n>, -Pn, -p ….

Two options:
1. Only insert a flag-derived value when the structured key is absent (entry(...).or_insert(...)), so explicit
params win. This matches the fold_flags_string "fill gaps" wording.
2. Keep the clobber but rename the test (flags_override_structured_params_when_present) and add a docstring
note on apply_flag_tokens that it is authoritative. Least-surprising for tool descriptors would be option 1.

Recommend option 1. Rename or delete the misleading test.

H2 — Non-string flags bypasses removal → key survives into out
crates/tools/src/external/nmap.rs:~493 — fold_flags_string:
let Some(flags) = params.get("flags").and_then(|v| v.as_str()) else {
    return params;
};
If flags is an array (["--sn", "-T4"]), an object, a number, or null, the function returns params
unchanged, meaning the flags key remains in the JSON handed to the rest of the pipeline. The PR body
promises "The raw flags key is always consumed - never passed through or shelled out." That promise fails for
any non-string type.

Downstream this probably lands as an "unknown parameter" schema error rather than a silent shell-out (so not
exploitable), but it violates the documented contract and is a plausible thing for an LLM to emit given it saw
flags work as a string once. Also inconsistent with the treatment of empty strings, which do get the key
stripped.

Fix:
let mut out = params.as_object().cloned().unwrap_or_default();
let flags_val = out.remove("flags");
let Some(flags) = flags_val.as_ref().and_then(|v| v.as_str()) else {
    return Value::Object(out);
};

Alternatively, accept flags as Value::Array<String> too — the LLM is at least as likely to emit that shape.

---
MEDIUM

M1 — Dropped-flag values can become the recovered target (carryover from args/command path)
apply_flag_tokens drops unknown -* tokens, but their values (the next token) fall through to the
target-recovery heuristic. Example: flags: "--script vulners" splits to ["--script", "vulners"]; --script
is dropped, vulners validates as a hostname, and — if no explicit target is set — becomes the scan target.
Nmap will then try to resolve and scan the host vulners.

This is not a new bug (existed in the args/command path pre-PR), but the PR extends the same surface to a shape
the LLM emits more frequently. Real-world exposure: --min-rate, --max-retries, --top-ports, --script,
--script-args, --proxies, --source-port, --data-length, -oN, -oX, -oG, --append-output, etc. —
any of these with a hostname-shaped value drops the flag but promotes the value.

Two fixes:
1. Skip the value of unknown flags: keep a small allowlist of flags that take a value (--min-rate,
--max-retries, --script, --script-args, -oN/-oX/-oG, --top-ports, --source-port, --data-length,
--proxies, --http-proxy, …) and consume tokens[i+1] when they appear. The generic version: any
--long that is not in the recognized set consumes both itself and its next token.
2. Only accept an IP or CIDR as the recovered target (not a hostname). Hostnames-as-targets from a legacy
shape are inherently ambiguous with flag values.

Recommend #2 for tight coupling: it eliminates the entire class silently, whereas #1 requires a
maintenance-heavy allowlist.

M2 — Description bloat may be truncated by some model backends
fn description — this string is now 4x longer and contains multi-clause guidance ("do NOT pass raw nmap CLI
flags", "there is no 'flags'/'args'/'command' param", "Unrecognized raw flags are dropped"). Some model
connectors truncate long tool descriptions or downweight late tokens. Consider moving the "there is no
flags/args/command param" line into a x-tool-hint field the schema supports, or into a system-prompt-level
instruction, rather than the tool description. Empirically: LLMs pay less attention to the middle of long
descriptions than to short ones.

Also, publishing the exact allowlist inside the description ("scan_type/timing/service_detection/no_ping/…")
makes the model less likely to try other structured params that the schema does support. Minor.

M3 — -Pn mapping is one-way (no_ping: true); no way to clear it via flags
If a caller sets no_ping: true in structured params and later passes flags: "-PS22" (a discovery probe
flag), the caller likely intended to override no_ping to false. But -PS22 is a dropped unknown flag →
no_ping stays true. The user's flags string is silently a no-op. Not exploitable, but combined with H1 the
semantics are inconsistent: some flags override structured params, others silently don't. Document the
asymmetry, or make all flag-application consistent with H1's fix (fill-gaps only).

---
LOW

L1 — No test for empty flags string alongside a valid target
Code correctly handles it (falls through with only the flags key stripped), but there's no test pinning the
behavior. Trivial to add.

L2 — No test for flags as non-string (see H2)
Add flags_non_string_is_stripped (or flags_non_string_is_rejected, depending on which behavior you pick).

L3 — has_shell_metacharacters doesn't include /, :, or ?, and that's fine but worth pinning
validate_target accepts CIDR (/) and IPv6 (:). Passing flags: "-sn http://evil.com/path" splits into
["-sn", "http://evil.com/path"]. The URL goes through validate_cidr (which fails on parse) and then
validate_hostname (which fails on /). Verified safe today. Consider a test asserting URL-shaped tokens
never validate — future changes to validate_hostname could regress this silently.

L4 — -p<spec> and -p <spec> write ports without validating
The port spec is validated later at execution time via validate_port_spec. Correct, but shell-metacharacter
rejection could happen at fold time too so bad values fail with a more actionable error. Optional.

L5 — apply_flag_tokens name doesn't reflect the mutation contract
The function both mutates out and returns a target Option<String>. Consider fold_flag_tokens_into or
apply_and_recover_target to make the two effects explicit.

L6 — flags_semicolon_target_token_is_rejected doesn't cover the whole cross-product
There is one negative test for ;. Consider a table-driven test enumerating every shell metacharacter from
has_shell_metacharacters (semicolon, pipe, backtick, $, parens, braces, brackets, <, >, backslash,
quotes, newline, tab) as a target-token candidate and asserting none pass. Guards against validate_target
regressions.

---
SECURITY-FOCUSED SUMMARY

From a pentester/security perspective, the PR is safe with respect to command injection. The critical seams are
all preserved.

┌───────────────────────────┬───────────────────────────────────────────────────────────────────────────────┐
│          Vector           │                                    Status                                     │
├───────────────────────────┼───────────────────────────────────────────────────────────────────────────────┤
│ Shell interpolation of    │ Not possible — recovered tokens go through validate_target (rejects ;|&$      │
├───────────────────────────┼───────────────────────────────────────────────────────────────────────────────┤
│ flags content             │ and backticks, etc.) and are passed as structured positional args via         │
├───────────────────────────┼───────────────────────────────────────────────────────────────────────────────┤
│                           │ CommandBuilder::positional, not shell-expanded.                               │
├───────────────────────────┼───────────────────────────────────────────────────────────────────────────────┤
│ Metachar injection via    │ Blocked at validate_port_spec downstream (which calls                         │
├───────────────────────────┼───────────────────────────────────────────────────────────────────────────────┤
│ -p <spec>                 │ has_shell_metacharacters).                                                    │
├───────────────────────────┼───────────────────────────────────────────────────────────────────────────────┤
│ Scope expansion via flag  │ NEW RISK (H1) — flags: "-A" overrides caller's structured params and          │
├───────────────────────────┼───────────────────────────────────────────────────────────────────────────────┤
│ clobber                   │ enables OS detection / script scan / traceroute even if the operator disabled │
├───────────────────────────┼───────────────────────────────────────────────────────────────────────────────┤
│                           │ them. Not a shell escape, but a policy escape (scope creep).                  │
├───────────────────────────┼───────────────────────────────────────────────────────────────────────────────┤
│ Target promotion from a   │ NEW SURFACE (M1) — carryover, now more likely to trigger given the shape      │
├───────────────────────────┼───────────────────────────────────────────────────────────────────────────────┤
│ dropped flag's value      │ LLMs emit.                                                                    │
├───────────────────────────┼───────────────────────────────────────────────────────────────────────────────┤
│ Prompt injection of flags │ Bounded by the allowlist. Unrecognized flags are dropped. The allowlist is    │
├───────────────────────────┼───────────────────────────────────────────────────────────────────────────────┤
│ to reach dangerous nmap   │ small and safe (-sn, -sS, -sT, -sU, -sV, -O, -A, -Pn, -T,                     │
├───────────────────────────┼───────────────────────────────────────────────────────────────────────────────┤
│ options                   │ -p). No -oX/-oN (filesystem write), no --script (LSE/NSE unsafe               │
├───────────────────────────┼───────────────────────────────────────────────────────────────────────────────┤
│                           │ scripts), no --iflist, no --send-eth, no -D (decoys).                         │
└───────────────────────────┴───────────────────────────────────────────────────────────────────────────────┘

Net: the fix closes the observed hang and does not introduce shell-injection or arbitrary-file-write vectors.
It does introduce a small policy-escape risk via unconditional flag clobbering (H1) that should be addressed
before merge.

---
RECOMMENDATIONS BEFORE MERGE

1. Fix H1 — make apply_flag_tokens fill gaps rather than clobber, and either rename or delete the
misleading flags_does_not_clobber_explicit_structured_params test. This is the biggest
correctness/least-surprise ask.
2. Fix H2 — strip the flags key regardless of its JSON type, and either accept arrays too or explicitly
reject them with a clear error.
3. Address M1 — tighten target-recovery to IPs/CIDRs only when the source is a flags string, or maintain
an allowlist of "flags that take a value" so unknown-flag values don't fall into the target slot.
4. Consider moving the "no flags/args/command param" note out of the description into either a schema-level
hint or the system prompt (M2).
5. Optional cleanups: L3, L4, L6 as follow-ups.

The core fix is correct and the test suite meaningfully exercises the RED scenarios. With H1 and H2 addressed,
this is ready to land.
```

ngdevo (approved):

cchao-devo (changes requested):
🔧 Request changes (minor)

  The core fix is correct and the safety contract holds end-to-end. There's one substantive gap to close before merge (a flags shape that silently reintroduces the very bug this PR fixes),
  plus a misleading test name and one deferrable follow-up. Once the shape gap is handled (or explicitly deferred with a note), this is an approve.

  Reviewed across four lenses — tests, correctness, contract/type-design, and silent-failure — not just test coverage.

  ⚠️  Blocking: flags as a JSON array/object silently drops the flags

  fold_flags_string only engages when flags is a string (.and_then(|v| v.as_str())). For {"flags": ["-sn","-T4"], "target": "..."} (array) or an object, it returns params unchanged, the
  flags key rides through, execute() ignores it, and the scan silently runs with defaults — exactly the "flags present + target present → silently dropped" failure this PR exists to kill,
  just via a different shape. The sibling args path already handles both array and string (nmap.rs:519–532), so this shape is well within what the same LLM will emit. No security exposure
  (never shelled), purely functional. Fix: mirror the args path's array-or-string tokenization in fold_flags_string.

  Non-blocking

  - Misleading test name. flags_does_not_clobber_explicit_structured_params actually asserts the opposite — flags do clobber a caller's explicit service_detection:false (→ true via -sV).
  target is correctly protected by the has_target guard, but the boolean/enum toggles are overridden. Defensible as "last CLI flag wins," but it's in tension with the new tool description
  telling the model structured params are authoritative. Suggest (a) renaming the test to reflect reality, and (b) a conscious call on whether the typed param should win for the toggles.
  - Dropped-flag observability (nit). Unrecognized flags (--min-rate, -PS/-PA/-PE/-PP) are dropped with no runtime signal. Documented in the description and safe, but a tracing::debug!
  listing dropped tokens would make a "why did my scan run differently" investigation tractable — relevant given this tool's failure mode already contributed to a stuck engagement.
  - Pre-existing -p- gap (separate follow-up, not this PR). flags:"-p-" → ports="-" → validate_port_spec errors, never reaching the "all" => -p- mapping. Identical on main, so not a
  regression — but this PR widens the flags surface where -p- appears, and the description advertises -p-, so the model will emit it and hit an error. Worth a LOW issue.

  What checks out ✅

  - Tests: fmt + clippy clean; nmap suite 55/55; all 5 new flags_* tests pass, including the injection / ;-rejection cases.
  - Correctness: tok[2..] on -p<spec> is panic-safe (bytes 0–1 are ASCII -p, byte 2 is always a char boundary); -T<n> parse is clean (len()==3 + to_digit); no off-by-one in the token loop.
  - Contract / type-design: the apply_flag_tokens extraction (mutate out, return target) is honored by both callers; the args/command path preserves an explicit target via the
  top-of-function early-return, so no clobber; no key loss/duplication across the two-phase pipeline; string flags is always stripped.
  - Safety (verified end-to-end): execute() builds a Vec<&str> argv and calls execute_command("nmap", &args_refs, …) — no shell string. Recovered target goes through validate_target; ports
  (even those from the flags path) go through validate_port_spec (rejects metacharacters) before -p. The structured-positional claim holds. Nice DRY consolidation.

jtomek-strike48:
Thanks for the four-lens review, @cchao-devo. Addressed in 85a6521:

**Blocking - `flags` as array/object silently drops flags.** Fixed. `fold_flags_string` -> `fold_flags`: it now tokenizes the same string-or-array shape the `args` path accepts, via a shared `tokenize_flag_value` helper (so `{"flags": ["-sn","-T4"], "target": "..."}` folds in exactly like the string form). Crucially it now ALWAYS strips the raw `flags` key for *any* shape - string, array, or an unexpected object - so a non-tokenizable value can never ride through to `execute()` and silently scan with defaults. New tests: `flags_as_array_with_target_maps_recognized_flags`, `flags_array_recovers_target_from_tokens`, `flags_as_object_is_stripped_not_passed_through`.

**Misleading test name + toggle semantics.** Made the conscious call you asked for: **typed structured params are authoritative**, resolving the tension with the tool description rather than just documenting it. Flag-derived values now fill only gaps (`merge_fill`, `entry().or_insert()`) - an explicit param the caller set is never overwritten. The test is renamed `flags_do_not_clobber_explicit_structured_params` and now asserts `service_detection:false` survives `-sV` (while `-T2` still fills unset `timing`). This applies to both the `flags` and `args`/`command` paths.

**Dropped-flag observability.** Added. `apply_flag_tokens` collects every dropped/unrecognized token and emits one `tracing::debug!` with `dropped_tokens = ?dropped`, so a "why did my scan run differently" investigation can see exactly what the normalizer discarded.

**Pre-existing `-p-` gap.** Filed as #257 (LOW) with root cause, repro, fix options, and tests - and noting it's not a regression. Deferred out of this PR per your call; the description already advertises `-p-`, so the follow-up is warranted.

Verification: nmap suite **54/54** (was 51 + 3 new), `security_tests` **55/55**, `cargo fmt` + `cargo clippy -p pentest-tools --lib -- -D warnings` clean. The safety contract you verified end-to-end is unchanged - recovered targets still go through `validate_target` and are passed as structured positional args; ports still go through `validate_port_spec`.

cchao-devo (approved):
Re-review: Pick PR #230 — fix(nmap): accept a flags string arg (commit 85a6521)

  Verdict: ✅ Approve — dismiss my prior "request changes." The blocking finding is genuinely fixed, all non-blocking points addressed or properly deferred, safety contract intact,
  gates green. The two new observations below are functional-only nits, not blockers.

  Lenses run: mechanical gates (fmt/clippy/tests) + manual source trace + adversarial claim-verification agent + silent-failure/security agent. Verified in pick_08/pick; pick_07
  untouched, no cluster.

  My prior review's findings — all resolved
  
  ┌───────────────────────────────────────┬──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┬──────────────────┐
  │             Prior finding             │                                                  Status in 85a6521                                                   │     Verified     │
  ├───────────────────────────────────────┼──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┼──────────────────┤
  │ BLOCKING — array/object flags         │ Fixed. fold_flags now strips the raw flags key unconditionally for any shape, and tokenize_flag_value folds string   │ ✅ traced + 3    │
  │ silently drops flags (reintroduces    │ and array. No JSON shape (number/bool/null/object/nested) can ride through to execute() — nothing re-inserts the     │ new tests pass   │
  │ the bug)                              │ key.                                                                                                                 │                  │
  ├───────────────────────────────────────┼──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┼──────────────────┤
  │ Misleading test name + toggle         │ Made the conscious call: typed structured params are authoritative. merge_fill (entry().or_insert()) means flags     │ ✅               │
  │ semantics                             │ fill gaps only; service_detection:false survives -sV. Test renamed and asserts the correct behavior.                 │                  │
  ├───────────────────────────────────────┼──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┼──────────────────┤
  │ Dropped-flag observability            │ Added tracing::debug!(dropped_tokens=?…). Fires only on non-empty drops; doesn't double-count a used target.         │ ✅               │
  ├───────────────────────────────────────┼──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┼──────────────────┤
  │ -p- gap (deferred)                    │ Filed as #257 (LOW, well-documented, correctly marked pre-existing/not-a-regression). Code correctly does not touch  │ ✅               │
  │                                       │ it.                                                                                                                  │                  │
  └───────────────────────────────────────┴──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┴──────────────────┘

  Gates (run locally on pr-230)

  - cargo fmt --all --check clean · cargo clippy -p pentest-tools --lib -- -D warnings clean
  - nmap suite + security_tests 55/55 — all 8 flags_* tests pass incl. injection/;-rejection cases

  Safety contract — re-confirmed end-to-end

  No shell string: execute() builds a Vec argv → execute_command("nmap", &args_refs) (execve-style, no sh -c). Recovered targets → validate_target; ports (incl. from -p<spec>) →
  validate_port_spec — both reject shell metacharacters. Non-string flags stripped before execute. No command-injection or scope-escalation-of-explicit-params exposure.

  Two new functional-only notes (non-blocking — optional follow-ups)

  1. -A umbrella bypasses an explicitly-disabled toggle. {flags:"-A", service_detection:false} → aggressive:true fills the unset gap, and execute()'s if aggressive {…} else 
  {service_detection…} makes -A enable service+OS detection regardless. Pre-existing execute() semantics (identical on main and via direct structured params); PR only widens reach to
  the flags shape; input is self-contradictory. Not a regression — mention only if you want the -A-vs-explicit-toggle precedence documented.
  2. Non-string array elements filtered before the drop-log. {flags:["-p",80]} — the numeric 80 is dropped inside tokenize_flag_value (filter_map(as_str)) before apply_flag_tokens, so
  it never appears in the new dropped_tokens telemetry; ports silently stays default. Small gap in the observability this very commit added. Malformed shape, "runs with defaults," no
  security impact. Trivial optional hardening: log elements dropped at tokenize time too.

  Both are LOW/optional. Nothing here should block merge.

jtomek-strike48:
Confirming the review feedback is resolved as of `85a6521`:

- **Array/object `flags` shape** (blocking): now folded like the args path - `flags_as_array_with_target_maps_recognized_flags` and `flags_as_object_is_stripped_not_passed_through` cover it; the key no longer rides through unhandled.
- **Typed params authoritative** (H1): `apply_flag_tokens` results are merged with `entry(k).or_insert(v)`, so an explicit structured param wins over a flag-derived one. The test is renamed `flags_do_not_clobber_explicit_structured_params` and now asserts the true (non-clobber) behavior.
- **Dropped-flag observability** (nit): unrecognized tokens are collected and emitted via `tracing::debug!(dropped_tokens = ?dropped, ...)`.
- **`-p-` gap**: confirmed pre-existing (identical on main), tracked as a separate follow-up, not this PR.

Both reviews are now APPROVED and CI is green.

PR #231, opened 2026-07-15
feat(config): add PENTEST_ALLOW_PRIVATE_IPS opt-in for in-cluster pri…

cchao-devo:
…vate ClusterIP hosts

 What & why
  
  ConnectorConfig::validate()'s SSRF guard runs in Production mode in release builds (ValidationMode::default() is #[cfg(debug_assertions)]-gated), which blocks private IPs. In an
  in-cluster/local-dev deployment the connector must dial the platform over a private ClusterIP (e.g. connectors-studio-grpc.default.svc:50061 → 10.x), so the release binary rejects it as
  SSRF and the connector can't start.

  This adds an opt-in env var PENTEST_ALLOW_PRIVATE_IPS=true|1 that relaxes the guard narrowly: it selects a new ValidationMode::PrivateNetwork that permits RFC-1918 ranges (10/8,
  172.16/12, 192.168/16) + loopback + IPv6 ULA, while still blocking the cloud-metadata / link-local range (169.254.0.0/16, fe80::/10) and other reserved ranges.

  Design notes

  - Secure by default: with the var unset (or non-truthy), behavior is unchanged — ValidationMode::default() (= Production in release). Verified by a release-gated test.
  - Narrow, not "disable SSRF": deliberately chose a new PrivateNetwork mode over full Development mode, so the prime SSRF credential-theft target (169.254.169.254) stays blocked even
  when the opt-in is on.
  - Auditable: logs a WARN when the guard is relaxed.
  - Pure mode-selection: the env→mode mapping is a pure function (resolve_validation_mode), unit-testable independent of build profile.
  - Parsing is case-insensitive + whitespace-trimmed (true/1).

  Tests
  
  7 new tests (run in CI's debug build): pure mode-selection (truthy variants, secure-default fallback, never-Development), end-to-end validate() (ClusterIP allowed, metadata blocked),
  and url_validation mode behavior (RFC-1918/loopback allowed; link-local/metadata/multicast/broadcast blocked; public allowed).

  CI note

  The Test job may show a pre-existing failure: all_tools_registered (wifi_scan_detailed not in capabilities) fails identically on main — it is not introduced by this PR (verified by
  checking out main and running the same cargo test --workspace --features pentest-platform/desktop-pcap).

  Consumer / follow-up

  Consumed by the init-dev local-dev stack (sets this env var on the Pick connector pod). For the published latest image to carry this, this PR should merge first so Pick CI rebuilds
  latest; init-dev then relies on it.

### DIFF
--- a/crates/core/src/config.rs
+++ b/crates/core/src/config.rs
@@ -188,14 +188,49 @@ impl ConnectorConfig {
             return Err("Tenant ID is required".to_string());
         }
 
-        // Validate host URL for SSRF protection
-        use crate::url_validation::{validate_url, ValidationMode};
+        // Validate host URL for SSRF protection.
+        use crate::url_validation::validate_url;
 
-        let validation_mode = ValidationMode::default();
+        let validation_mode = Self::resolve_validation_mode(
+            std::env::var("PENTEST_ALLOW_PRIVATE_IPS").ok().as_deref(),
+        );
         validate_url(&self.host, validation_mode, None)
             .map_err(|e| format!("Invalid host URL: {}", e))?;
 
         Ok(())
+    }
+
+    /// Select the SSRF [`ValidationMode`] for the connector host from the
+    /// `PENTEST_ALLOW_PRIVATE_IPS` opt-in.
+    ///
+    /// In-cluster/local dev connects to the platform over a private ClusterIP
+    /// (e.g. `connectors-studio-grpc.default.svc:50061` -> `10.x`), which the
+    /// default Production guard blocks as SSRF. Setting
+    /// `PENTEST_ALLOW_PRIVATE_IPS=true|1` selects [`ValidationMode::PrivateNetwork`],
+    /// which permits RFC-1918 / loopback targets but STILL blocks the
+    /// cloud-metadata / link-local range (169.254.0.0/16) — a deliberately
+    /// narrower relaxation than full Development mode.
+    ///
+    /// Secure by default: any other value (including unset) falls back to
+    /// [`ValidationMode::default()`], which is Production in release builds.
+    /// Pure (no I/O) so it can be unit-tested independent of the build profile.
+    fn resolve_validation_mode(env_val: Option<&str>) -> crate::url_validation::ValidationMode {
+        use crate::url_validation::ValidationMode;
+        let allow_private = env_val
+            .map(|v| v.trim().to_ascii_lowercase())
+            .is_some_and(|v| v == "true" || v == "1");
+        if allow_private {
+            // Loud, auditable: an operator (or a leaked env var) has relaxed the
+            // SSRF guard. Emitted once per validate() call, not per request.
+            tracing::warn!(
+                "PENTEST_ALLOW_PRIVATE_IPS is set: connector host SSRF validation \
+                 relaxed to PrivateNetwork mode (RFC-1918/loopback allowed; \
+                 link-local/metadata still blocked). Do NOT use in production."
+            );
+            ValidationMode::PrivateNetwork
+        } else {
+            ValidationMode::default()
+        }
     }
 
     /// Check if this config has authentication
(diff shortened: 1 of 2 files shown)


jtomek-strike48:
## Multi-perspective review (security, Rust, tests, silent-failure)

Overall: a solid, security-conscious change. The relaxation is genuinely narrow (new `PrivateNetwork` mode, secure-by-default, metadata still blocked), and I confirmed the scope is contained to the connector's own host: ffuf/nikto/gobuster/dirb/seed all pass `ValidationMode::Production` directly, so the opt-in cannot leak into tool-target SSRF checks. CI is green, clippy/fmt clean, exhaustiveness over the new variant is fine.

One HIGH issue to resolve before merge, plus a missing test for a claimed security property.

### HIGH - the documented use case (ClusterIP by DNS name) is silently blocked

The docstring's motivating example is `connectors-studio-grpc.default.svc:50061 -> 10.x`. Tracing the flow:

- A hostname (not a literal IP) is not parseable as `IpAddr`, so `classify_lan_private` returns `LanClass::Public`.
- `Public` falls through to the Production path, where `classify_host` DNS-resolves the name to `10.x` -> `HostClass::Private` -> `Err("Private IP addresses are not allowed in production mode")`.

So with `PENTEST_ALLOW_PRIVATE_IPS=true` set, a ClusterIP DNS name still fails to validate. Only a literal ClusterIP works - which is exactly (and only) what the tests exercise (`grpc://10.109.18.109:50061`). That literal-IP-only test coverage is why CI is green.

This is a deliberate DNS-rebinding defense (correct instinct, and the comments explain it), but it collides with the documented deployment model. Please either:

- (a) fix the docstring / PR description to state operators must configure a literal ClusterIP, not a `.svc` DNS name, and confirm the `init-dev` consumer sets a literal IP; or
- (b) if DNS names must work, treat it as a design change with explicit DNS-rebinding tradeoffs.

As written, an operator following the docstring will hit a confusing failure despite opting in.

### MEDIUM - missing test for a claimed security property

The code explicitly claims it does not let a hostname resolve into the LAN-private allow-set (DNS-rebinding defense). There is no test for this - all 7 new tests use literal IPs. A test asserting `validate_url("grpc://<name-resolving-to-10.x>", PrivateNetwork, None).is_err()` would protect the property and would have immediately surfaced the HIGH finding above.

### MEDIUM - "still blocks cloud-metadata" claim is incomplete for IPv6-encoded forms

`::ffff:169.254.169.254` (IPv4-mapped) and `64:ff9b::a9fe:a9fe` (NAT64) classify as public and would be allowed - `is_private_ipv6` never canonicalizes embedded IPv4. Important: this is pre-existing on `main` (I diffed `is_private_ipv6`; it is byte-identical) and affects the default Production guard identically, so this PR does not regress anything and it should not block merge. But this PR newly and prominently asserts `PrivateNetwork` "STILL blocks the cloud-metadata / link-local range", which overstates the protection for IPv6-encoded metadata. Suggest a follow-up issue to canonicalize IPv4-mapped/NAT64 in `is_private_ipv6` (benefits Production too) and softening the absolute wording.

### MEDIUM - no diagnostic when the env var is set to a non-truthy value

A typo like `PENTEST_ALLOW_PRIVATE_IPS=ture` silently falls back to the secure default with no log; the operator then debugs a "private IP blocked" failure with no hint the opt-in was ignored. The WARN only fires on the success path. Suggest adding a WARN on the set-but-non-truthy branch. Secure-default behavior itself is correct; this is operability only.

### LOW / advisory

- `is_lan_private_ipv4/ipv6` duplicate RFC-1918 + loopback ranges already in `is_private_ipv4/ipv6`, in two different styles. Could use std (`Ipv4Addr::is_private() || is_loopback()`, `Ipv6Addr::is_unique_local() || is_loopback()`) or define `is_private_ipv4` as a superset. Matches existing file convention, so cleanup-not-blocker.
- Error messages reveal the active mode/range, and the env var is re-read per `validate()` rather than cached. `validate()` runs once at startup (apps/headless/src/main.rs:60), so neither is a practical concern.

### Nits that are actually strengths

Secure-by-default is real and release-gated. Choosing a narrow `PrivateNetwork` mode over full `Development` keeps 169.254.0.0/16 blocked. Env-var parsing is robust (case-insensitive, trimmed, empty/unset both safe).

Bottom line: the security model is sound and this does not weaken the connector versus `main`. The one thing worth holding for is the HIGH finding - confirm literal ClusterIP vs DNS name - and adding the hostname-resolution test while in there.

cchao-devo:
@jtomek-strike48 

  Thanks for the thorough review — all findings addressed. Pushed two commits (d88e985, 7e91b6b).

  HIGH — DNS-name ClusterIP blocked: Confirmed and fixed. You were exactly right, and it bites the actual consumer: init-dev sets connector.strikeUrl: 
  "grpc://connectors-studio-grpc.default.svc.cluster.local:50061" — a .svc DNS name, not a literal IP. I reproduced the crashloop with the real PrivateNetwork binary on the live cluster
  (Private IP addresses are not allowed in production mode), then fixed it: classify_lan_private now resolves a hostname and admits it iff every resolved IP is LAN-private/loopback, still
  deferring/blocking if any resolved IP is public or link-local/metadata — so the .svc deployment works while the DNS-rebinding defense is preserved. Re-verified end-to-end on the cluster: the
  connector now validates the DNS name, gets approved, and executes tools.

  MEDIUM — missing hostname test: Added test_private_network_mode_allows_hostname_resolving_to_private (localhost→loopback, deterministic/offline-safe) and ..._hostname_to_public_not_admitted.

  MEDIUM — non-truthy env WARN: Added — resolve_validation_mode now warns when PENTEST_ALLOW_PRIVATE_IPS is set to an unrecognized value (e.g. ture) and falls back to the secure default, so a
  typo isn't silently ignored.

  MEDIUM — IPv6-encoded metadata wording: Agreed it's pre-existing on main and shouldn't block. Softened the docstring to stop overclaiming, and filed #236 to canonicalize IPv4-mapped/NAT64 in
  is_private_ipv6 (benefits Production too).

  LOW — dedup is_lan_private_* via std: Left as-is for now (matches the file's existing hand-rolled convention); can fold into #236's cleanup.

jtomek-strike48 (approved):
## Approve

Reviewed across the full stack (rust / security / test-coverage agents) plus empirical verification in an isolated worktree at the PR head. Safe to merge.

### Verified empirically (ran it, not just read the diff)
- **Secure by default**: env unset -> `ValidationMode::default()` = Production in release; behavior unchanged from before this PR.
- **Narrow, not "disable SSRF"**: `PrivateNetwork != Development`. Canonical `169.254.169.254` (and `169.254.170.2`, `fe80::1`) stay **blocked** in PrivateNetwork mode. Confirmed by running `validate_url`.
- **Guard test genuinely guards** (mutation check): neutering the metadata block (treating internal ranges as LAN-private) turns `test_private_network_mode_still_blocks_link_local_and_metadata` RED; restoring makes it green. Not a tautology.
- **DNS-rebinding defense**: a hostname is admitted only if every resolved IP is LAN-private; an empty/failed resolve falls through to Production (fail-closed).
- **RFC-1918 + ULA boundaries** correct (172.16/12, `fc00::/7` mask, 10/8, 192.168/16, loopback).
- **CI-scope**: the 7 new tests live in `pentest-core`, run by both the Linux and macOS required lanes, with no `debug_assertions`/feature gate hiding them.

### On the IPv6-encoded metadata caveat (the one thing worth naming)
`::ffff:169.254.169.254` and NAT64 `64:ff9b::a9fe:a9fe` do reach the metadata endpoint. Verified empirically that this is **allowed in BOTH Production and PrivateNetwork** (canonical `169.254.169.254` blocked in both) -> it is a **pre-existing Production limitation this PR inherits, not a new hole**. The PR body's disclosure ("shared with Production mode, not specific to this variant") is accurate. Not a blocker.

### Non-blocking suggestions (follow-up, not gating this approval)
1. Add a known-limitation test pinning the IPv6-encoded-metadata behavior, so the future `is_private_ipv6` canonicalization fix has an assertion to flip.
2. File the `is_private_ipv6` embedded-IPv4 canonicalization follow-up as a tracked issue (it fixes both Production and PrivateNetwork), rather than leaving it only in a code comment.

Issue #194, opened 2026-07-04
Browser OAuth callback times out: token relay lands on raw JSON instead of posting back to local /token


jtomek-strike48:
## Summary

The desktop connector's browser-based OAuth login (`FetchTokenFallback` -> `[BROWSER_AUTH]`) completes authentication on the Keycloak/Matrix side but **never delivers the token back to the connector**, so the connect flow dies with `Timeout waiting for token` after 120s. Observed on a local `studio.strike48.test` stack (Ubuntu/GNOME/Wayland, WebKitGTK desktop build).

## Impact

The desktop connector cannot authenticate via browser OAuth against the local instance. This blocks any desktop end-to-end run (including manual verification of connector features). It is not a connectivity problem — the WebSocket transport connects and registers fine; only the token hand-off fails.

## Repro

1. Launch the desktop connector (`pentest-connector`) against `studio.strike48.test:443` with no saved JWT and no valid OTT (so it falls to browser OAuth).
2. `[BROWSER_AUTH]` opens the browser to `{api_url}/auth/login?redirect=http://localhost:5173/callback` and starts the local callback server on port 5173.
3. Complete login in the browser.
4. **Observed:** the browser ends on a page showing the **raw token JSON** (`{"access_token":"...","refresh_token":"...","expires_at":...}`), and the connector logs `Timeout waiting for token` / `Login timed out — no token received within 120 seconds`.
5. **Expected:** the token is relayed to the connector's local `/token` endpoint and the connector proceeds to Connected.

### Log evidence

```
[BROWSER_AUTH] Opening browser to: https://studio.strike48.test:443/auth/login?redirect=http%3A%2F%2Flocalhost%3A5173%2Fcallback
[BROWSER_AUTH] Callback server listening on port 5173
[BROWSER_AUTH] /callback hit, serving token-fetch page (HTML len=3624)
[BROWSER_AUTH] Timeout waiting for token
[FetchTokenFallback] browser OAuth failed: matrix: Login timed out — no token received within 120 seconds
```

Note `/callback` is hit and the token-fetch page is served, but `/token` is **never** hit (no `[BROWSER_AUTH] /token called with query params` line ever appears).

## Root cause (analysis)

The token-relay contract in `crates/core/src/matrix/auth.rs`:

- The local callback server accepts the token only as a **URL query param**: `/callback?access_token=...` (`auth.rs:513`) or `/token?access_token=...` (`auth.rs:552`). Both read `query.get("access_token")`.
[...]

PR #234, opened 2026-07-15
fix(#194): fast-fail browser-OAuth when server won't relay the token

jtomek-strike48:
## Summary

Fixes the connector-side half of #194: desktop/standalone browser-OAuth login blocked for the full 120s and then reported a generic timeout whenever the server returned the access token as a JSON body instead of redirecting it back to the connector's loopback `/token` endpoint.

**Root cause (verified against production, not a dev-cert artifact).** Probed `/auth/refresh?redirect=` directly against both a self-signed local stack and a valid-TLS production server:

| Probe | Local dev (self-signed) | Production (valid TLS) |
|-------|-------------------------|------------------------|
| `GET /auth/refresh?redirect=<loopback>/token` | `application/json`, no 302 | `application/json`, no 302 (identical) |
| `GET /auth/login?redirect=<loopback>/callback` | 302 to Keycloak | 302 to Keycloak |
| `OPTIONS /auth/refresh` (CORS preflight) | 404 | 404 |

Both servers ignore `?redirect=` on `/auth/refresh` and return the token as JSON, so the callback page's Strategy 3 (top-level redirect expecting a 302 back to loopback) can never complete. Once the browser navigates to that cross-origin JSON page, the callback JS is gone, so the connector cannot recover the token on its own.

## What this PR changes (connector side)

- The callback page pings a same-origin `/diag` beacon immediately before the one-way Strategy 3 redirect, reporting why strategies 1-2 failed.
- Extracts the wait logic into `await_browser_token`: a `biased` select that races token delivery against the diagnostic beacon and an overall ceiling. On the JSON-landing case it waits a 15s grace window for a token, then fails fast with a precise message naming the server-side relay gap and pointing operators at `MATRIX_AUTH_TOKEN` / `KEYCLOAK_USERNAME`+`PASSWORD` as a bypass.
- Aborts the callback server on every exit path so the scarce CORS-whitelisted port (4000/5173) is released even when auth fails.

## Depends on (server side)

The flow only completes end-to-end once the server honors the loopback token relay. Tracked in `Strike48/matrix#3170` (targets `develop`): widen the raw-token relay guard in `AuthPlug.callback/1` to cover `http://localhost:<port>/` redirects, exactly as it already does for the Android `com.strike48.pentest://` scheme. This PR makes the failure fast and legible in the meantime.

## Related

- Not a duplicate of `fix/desktop-no-auto-browser-oauth` (do-not-merge, blocked by #85): that branch *disables* browser-OAuth by default in the UI layer; this fixes *why* the relay fails, in `crates/core`. Different files, complementary.
- #81 (random-port/CORS) and #85 (ChatPanel token regression) are both closed; #194 is the live tracker.

## Test plan

- [x] `cargo test -p pentest-core --features browser-auth browser_token_tests` - 4 new unit tests pass (token-delivered, token-wins-over-diagnostic, fast-fail-on-JSON-landing, hard-timeout)
- [x] `cargo clippy -p pentest-core --features browser-auth --all-targets -- -D warnings` clean
- [x] `cargo fmt` applied
- [ ] Live desktop run against a server with matrix#3170 applied: browser login completes and reaches Connected
- [ ] Live desktop run against a server WITHOUT matrix#3170: connector fails within ~15s with the contract-gap diagnostic (not a 120s hang)

### DIFF
--- a/crates/core/Cargo.toml
+++ b/crates/core/Cargo.toml
@@ -54,3 +54,6 @@ axum = { workspace = true, optional = true }
 pentest-tools = { path = "../tools" }
 pentest-platform = { path = "../platform", features = ["desktop"] }
 tempfile = "3"
+# `test-util` (for `tokio::time::pause`/`start_paused`) is NOT part of tokio's
+# `full` feature, so enable it here for deterministic time in async tests.
+tokio = { workspace = true, features = ["test-util"] }
--- a/crates/core/src/matrix/auth.rs
+++ b/crates/core/src/matrix/auth.rs
@@ -353,6 +353,15 @@ pub async fn fetch_matrix_token_browser(matrix_url: &str) -> crate::error::Resul
     let (tx, rx) = tokio::sync::oneshot::channel::<String>();
     let tx = std::sync::Arc::new(tokio::sync::Mutex::new(Some(tx)));
 
+    // Diagnostic channel: the callback page pings `/diag` immediately before it
+    // falls back to the Strategy 3 top-level redirect. That redirect only
+    // completes if the server 302s `/auth/refresh?redirect=<loopback>/token`
+    // back to us; several server builds instead return the token as JSON and
+    // never redirect (see issue #194), which strands the flow. This ping lets
+    // us fail fast with an actionable message instead of blocking the full 120s.
+    let (diag_tx, diag_rx) = tokio::sync::oneshot::channel::<String>();
+    let diag_tx = std::sync::Arc::new(tokio::sync::Mutex::new(Some(diag_tx)));
+
     // -----------------------------------------------------------------------
     // Bind the callback server FIRST so we know the local port before
     // generating the callback HTML (which embeds it for the redirect fallback).
@@ -479,10 +488,24 @@ pub async fn fetch_matrix_token_browser(matrix_url: &str) -> crate::error::Resul
 
     // Strategy 3: Redirect to Matrix origin for same-site token exchange.
     // Top-level GET navigations send SameSite=Lax cookies. The server's
-    // /auth/refresh?redirect=<url> does the refresh and 302s back to our
-    // /token endpoint with ?access_token=xxx appended.
+    // /auth/refresh?redirect=<url> is *expected* to do the refresh and 302
+    // back to our /token endpoint with ?access_token=xxx appended.
+    //
+    // Known failure (issue #194): some server builds return the token as a
+    // JSON body instead of honoring ?redirect=, so the browser lands on raw
+    // JSON on the Matrix origin, this page's JS is gone, and /token is never
+    // hit. Because that navigation is one-way, we ping our own same-origin
+    // /diag endpoint FIRST so the connector can fail fast with a precise
+    // message instead of blocking for the full 120s timeout.
     var tokenUrl = 'http://localhost:' + LOCAL_PORT + '/token';
     var refreshUrl = MATRIX_URL + '/auth/refresh?redirect=' + encodeURIComponent(tokenUrl);
+    log('[CALLBACK] Strategies 1+2 failed; arming diagnostic before redirect');
+    try {{
+      await fetch('/diag?stage=strategy3_redirect&reason=' +
+        encodeURIComponent(e && e.message ? e.message : 'strategy2_failed'));
+    }} catch (diagErr) {{
+      log('[CALLBACK] /diag ping failed: ' + diagErr.message);
+    }}
     log('[CALLBACK] Redirecting to Matrix origin: ' + refreshUrl);
     s.textContent = 'Completing login…';
     d.textContent = 'Redirecting for token exchange…';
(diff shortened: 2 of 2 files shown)


jtomek-strike48:
## Review pass (multi-agent) - findings and resolution

Ran a 5-perspective review (Rust correctness, silent-failure, test coverage, comment accuracy, simplification). Pushed one follow-up commit (334d8e8) addressing the actionable findings.

### Fixed

- **Untested slow-relay success path (Important).** The nested grace-window `Ok(res)` arm - Strategy 3 legitimately succeeding but slowly (the real path for SameSite=Lax servers) - had no coverage; the existing `token_wins_over_diagnostic` test only hit the biased outer select. Added `token_arrives_within_grace_after_diag_fires`. This is the branch a too-short grace would turn into a spurious login failure, so it is the most important one to lock down.
- **Untested channel-closed paths (Important).** Added `errors_when_token_sender_dropped_before_diag` and `errors_when_token_sender_dropped_during_grace` to cover the `channel_closed()` mapping on both the outer arm and the nested grace wait.
- **Timing-based test flakiness (Important).** Converted the timing tests to `#[tokio::test(start_paused = true)]` so grace/ceiling assertions run on deterministic virtual time (no wall-clock dependence). Enabled tokio `test-util` as a dev-dependency (it is not in tokio's `full` feature). Suite now runs in ~0ms.

Test count: 4 -> 7, all passing; clippy `-D warnings` clean.

### Reviewed and intentionally not changed

- **15s `REDIRECT_GRACE`** caps the legitimate Strategy 3 redirect round-trip (previously it had the full 120s ceiling as a blind budget). A redirect + server refresh realistically completes in 1-3s, so 15s is generous headroom; the new within-grace success test guards this path.
- **`let diag_tx = diag_tx.clone()` twice in the /diag route** - not redundant. It matches the required axum handler idiom used by the sibling `/callback` and `/token` routes (outer clone for the route-builder `Fn` closure, inner clone moved into the per-request future). Verified against lines 567/569.
- **Best-effort `let _ = sender.send(...)`, JS `/diag` catch, and query-param defaults** - all confirmed intentional graceful-degradation, not silent failures: if the diagnostic never fires, the flow correctly falls back to the 120s ceiling.
- **~135s worst-case overshoot** if the diag beacon arrives in the final 15s of the ceiling - pathological and benign (still returns the right error); not worth the added complexity of clamping the nested timeout.
- Comment accuracy: all four load-bearing comment blocks verified accurate against the code.

cchao-devo (approved):
Review — PR #234 (fast-fail browser-OAuth relay)
  
  Careful, well-tested change. The await_browser_token extraction makes the wait logic unit-testable without a browser/socket, the biased select is correct, and the failure message is
  actionable (names the server-side #194 relay gap + the MATRIX_AUTH_TOKEN bypass). No overlap with the in-flight PENTEST_ALLOW_PRIVATE_IPS work in this same file. I checked out the branch
  and ran the tests to verify the findings below.

  🟡 MEDIUM — new tests are silently skipped in the macOS CI lane
  
  browser_token_tests is #[cfg(all(test, feature = "browser-auth"))], and browser-auth is enabled only by the app crates (apps/{web,desktop,mobile}), not by pentest-core itself. I ran both
  CI lanes' exact commands on the branch:

  - Linux cargo test --workspace --features pentest-platform/desktop-pcap → 7 passed (the app crates pull browser-auth onto pentest-core via feature unification). ✓
  - macOS cargo test --package pentest-platform --package pentest-core --lib --features pentest-platform/desktop-pcap → 0 browser_token tests run (240 other core tests pass); they're
  compiled out because that lane is package-scoped and never enables browser-auth.

  So the tests run, but coverage is incidental via cross-crate unification and one lane skips them entirely. Suggest making it explicit — add --features pentest-core/browser-auth to the test
  invocation(s) — so the coverage doesn't silently vanish if the app-crate feature wiring ever changes. Low effort; avoids a future "green but untested" surprise.

  (Also: all 7 tests pass, incl. the start_paused deterministic-time cases — no CI flakiness risk.)

  🟢 Low / advisory

  - /diag is an unauthenticated same-origin beacon on the loopback callback server. Any local process reaching 127.0.0.1:<port>/diag during the auth window can fire the diag oneshot, pushing
  into the 15s grace path early. Impact minimal (worst case: slightly-earlier fail-fast on a user-initiated login; no token exfil, no auth bypass; port is short-lived + CORS-scoped). Noting
  only as a new externally-pingable endpoint.
  - channel_closed vs #194 message in the grace path (documented by errors_when_token_sender_dropped_during_grace): I traced the real flow — the token sender lives in the /token route
  closure inside server_handle, alive until after await_browser_token returns — so in the actual #194 case the sender isn't dropped, the grace window elapses, and the operator gets the
  correct #194 message. Confirmed not a mis-routing bug.
  - tokio test-util Cargo addition is appropriate and test-scoped.

  ✅ Verified (by reading + running)

  - biased select prioritizes token over diag when both ready (token_wins_over_diagnostic_when_both_present ✓).                                                             
  - No lost-token race: the diag arm re-awaits the same token_rx under a timeout, so a token during the grace window is still delivered (token_arrives_within_grace_after_diag_fires,
  start_paused ✓).
  - server_handle.abort() on every exit path (success tail + error match) → port always released.
  - All failure modes covered: #194 fast-fail, 120s ceiling, channel-closed.

  Verdict: approve — the MEDIUM is a test-wiring hardening, not a correctness issue; all functional paths check out and the tests pass.

Issue #172, opened 2026-06-28
Tool-produced evidence never reaches the report (PENDING_EVIDENCE -> EVIDENCE_GRAPH bridge missing in production)


jtomek-strike48:
## Summary

In a native production build, **tool-produced evidence never reaches the final report.** Tools push findings into one process-global static (`PENDING_EVIDENCE`, in `pentest-tools`); the Generate Report path reads a *different*, never-populated static (`EVIDENCE_GRAPH`, in `pentest-ui`). Nothing in production code moves data from the first into the second, so `gate_for_report` builds the manifest from an empty set. Only the `inject_test_evidence` test tool ever reaches the report path (and even it pushes to the wrong buffer).

The report comes out findingless for real engagements, while tests and manual demos pass because they inject directly. Worse than silent: code comments and `ARCHITECTURE.md` assert the flow works, so it reads as "wired" on inspection.

## The two disconnected buffers

| Store | Location | Writer | Reader |
|-------|----------|--------|--------|
| `PENDING_EVIDENCE` (tool buffer) | `crates/tools/src/evidence_producer.rs:44` | `evidence_producer::push_evidence` (`:87`) - called by all real tools | `drain_pending_evidence` (`:126`) - **callers are test-only** |
| `EVIDENCE_GRAPH` (report graph) | `crates/ui/src/session.rs:42` | `session::push_evidence` (`:153`) - **zero callers** | `evidence_snapshot` (`:143`) - the report path |

Real tools write to `PENDING_EVIDENCE`: `external/nmap.rs:287`, `external/web/whatweb.rs:96`, `service_banner.rs:229`, `webwright/evidence.rs` (5 sites), `inject_test_evidence.rs:159`. All are registered in the production registry (`crates/tools/src/lib.rs`).

The report reads `EVIDENCE_GRAPH`: `crates/ui/src/components/chat_panel/mod.rs:760` calls `session::evidence_snapshot()`, then `pentest_core::orchestrator::gate_for_report(&snapshot, engagement)` (`:765`).

## Verification (static)

- `drain_pending_evidence()` has **zero non-test production callers** repo-wide. Every caller is in `#[cfg(test)]` modules or `crates/tools/tests/webwright_integration.rs`.
[...]

PR #240, opened 2026-07-16
fix(#184): fail-closed grounding guardrail — agent can no longer fabricate tool results

### LINKED ISSUE #184: Fail-closed grounding guardrail: agent must never fabricate tool results
# Fail-Closed Grounding Guardrail: agent must never fabricate tool results (no un-grounded findings)

**Issue Type:** Bug (grounding/integrity) - treat as P0
**Component:** Pick agent (`crates/tools` result schema, `crates/core` orchestrator/evidence gate, `crates/ui` prompts)
**Effort:** Medium-Large (1-1.5 weeks)
**Priority:** P0
**Source:** Operator observation - a Pick engagement on Windows fabricated tool results (reported findings/output for scans that never actually ran). Root-cause class is cross-platform, not Windows-specific.
**Target repo:** `Strike48-public/pick`
**Related (do NOT duplicate):**
- `pick#172` / `pick#174` (evidence never reaches the report -> *empty* reports). This issue is the opposite failure: *fake-full* reports. Different seam, arguably higher severity.
- `matrix#2851` (connector tools missing from agent registry -> loud "Unknown tool" error). That fails loudly; this fails silently and confidently.
- Windows silent-failure root cause tracked under the Linux/Windows parity epic (red-team-037). This issue is the cross-platform guardrail; red-team-037 fixes the Windows triggers.

> **Reality check (verified against code 2026-07-01 via trace):**
> - Tool `Err`s are **not dropped**, they are **demoted**: `execute_timed` / `execute_timed_with_provenance` turn any tool-body `Err(e)` into `Ok(ToolResult::error(...))` (`crates/core/src/tools.rs:441,460`). The failure survives only as a `success:false` bool + `error` string; the outer `Result` is always `Ok`. Nothing in-repo forces a downstream consumer (or the model) to fail-closed on `success:false`.
> - **Empty-but-successful vectors exist:** `execute_command` reports `success:true` for any process that ran, regardless of a nonzero exit code or empty stdout (`crates/tools/src/execute_command.rs:120-126`). Scanner wrappers only error on `exit_code != 0 && stdout.is_empty()` (`crates/tools/src/external/rustscan.rs:135`, `ffuf.rs:186`); a degraded/partial scan parses to `Ok(empty)` and `parse_rustscan_output` never returns `Err` for empty output (`rustscan.rs:150-193`). **There is no schema distinction between "ran and legitimately found nothing" and "failed to actually probe."** That ambiguity is the fabrication substrate.
> - The only hard code gate is `gate_for_report` (`crates/core/src/orchestrator.rs:185`): it refuses `Pending` nodes but **does not require provenance** - `ManifestFinding.provenance` and `EvidenceNode.provenance` are both `Option` and `None` passes (`orchestrator.rs:100,118`; `evidence.rs:153,172`).
> - **All anti-fabrication discipline today is prompt text, not code:** `crates/ui/.../chat_panel/constants.rs:707` ("Do not invent findings... if the manifest is empty, say so"), `:892`/`:895` (Validator "ground every verdict... never invent evidence"). No `fail_closed` identifier exists anywhere in code.
> - **`inject_test_evidence` ships in production, ungated.** `crates/tools/src/lib.rs:241` registers `InjectTestEvidenceTool` with **no `#[cfg(debug_assertions)]`/test gate**. It is a first-class, always-available path to inject arbitrary fabricated findings (caller-supplied severity + provenance) straight into the evidence graph, and they can pass `gate_for_report`.

---

## Problem Statement

For a pentest platform whose entire premise is "no hallucinations - ground every claim in real tool output," the agent can currently present results that never happened. Three independent code weaknesses combine:

1. A failed or degraded tool run is indistinguishable from a genuine empty result (only a `success` bool separates them, and several tools set `success:true` on functional failure).
2. Nothing in code stops the model from narrating findings that have no backing `ToolResult`; the only guardrail is prompt instructions the model can ignore.
3. A production-registered tool (`inject_test_evidence`) can write fabricated findings directly into the pipeline.

On Windows this is trivially reachable because several probes return silent empties instead of errors (see red-team-037), so the model receives no failure signal and "helpfully" fills the gap. But the invariant - **never assert a result that isn't grounded in actual tool output** - is platform-independent.

---

### Lever 1: Remove the production fabrication path (fastest, standalone)
Gate `inject_test_evidence` out of production builds. Either `#[cfg(any(test, debug_assertions))]` on the `registry.register(InjectTestEvidenceTool)` call (`crates/tools/src/lib.rs:241`) and its module, or move it behind an explicit dev/test feature flag. No production build should ship a tool that injects arbitrary findings.

### Lever 2: Distinguish "failed" from "empty" in the tool-result schema
Introduce an explicit outcome that the model and the report gate can trust, rather than inferring from `success` + exit code:
- Add a `completed: bool` (or an `outcome: Ran | RanEmpty | Failed | Skipped` enum) to `ToolResult`, distinct from `success`.
- `execute_command`: a nonzero exit with empty stdout should surface as `Failed`/`RanEmpty`, not silent `success:true` (`execute_command.rs:120-126`).
- Scanner wrappers: a nonzero exit (even with partial stdout) or an unparseable output should be `Failed`, not parsed as `Ok(empty)` (`rustscan.rs:135`, `ffuf.rs:186`). A genuine zero-finding scan stays `RanEmpty` (truthful).

### Lever 3: Require grounding at the report gate (fail-closed)
Make `gate_for_report` (`orchestrator.rs:185`) refuse to emit a finding-type node that lacks provenance tying it to a real `ToolResult`. Provenance moves from advisory (`Option`, opt-in via `with_provenance`) to **required for finding nodes** entering the manifest. A finding the model asserted with no backing tool output cannot reach the report. (Coordinate with `pick#174` seam-3: the Validator path must be able to attach/clear provenance, or fixing this deadlocks on Pending nodes - see that epic.)

---

## Testing Strategy (Rust)

```rust
// Lever 1
#[test] fn inject_test_evidence_not_registered_in_release() { /* assert absent from default_registry() in non-test build */ }

// Lever 2
#[test] fn execute_command_nonzero_exit_empty_stdout_is_not_success() { /* Failed/RanEmpty, not success:true */ }
#[test] fn rustscan_nonzero_exit_is_failed_not_empty_ok() { /* Failed, not Ok(open_ports:[]) */ }
#[test] fn genuine_zero_finding_scan_reports_ran_empty_truthfully() { /* RanEmpty, success true, count 0 */ }

// Lever 3
#[test] fn report_gate_rejects_finding_node_without_provenance() { /* GateError, manifest not built */ }
#[test] fn report_gate_accepts_finding_with_real_provenance() { /* Ok */ }
```

Plus one production (non-test) integration test that runs a tool that fails, and asserts the failure is visible to the report gate as non-grounded (ties to the `pick#172`/`#174` "no test-only path" concern).

---

## Acceptance Criteria

- [ ] `inject_test_evidence` is not registered in production/release builds (test/dev only)
- [ ] Tool results distinguish "ran and found nothing" from "failed to probe" (schema-level, not inferred from exit code)
- [ ] `execute_command` and scanner wrappers no longer report functional failures as `success:true`/`Ok(empty)`
- [ ] `gate_for_report` fails closed on finding-type nodes lacking provenance (no un-grounded finding reaches a report)
- [ ] The model receives a hard, structured failure signal when a tool did not actually complete
- [ ] Unit tests for all three levers + one production-path integration test; existing tests still green
- [ ] Cross-links: red-team-037 (Windows triggers), `pick#174` (Validator/provenance seams), `matrix#2851` (loud-failure sibling)

---

## Guardrails / Notes

- **Never weaken a real result to satisfy the gate.** Requiring provenance must not tempt auto-attaching synthetic provenance; the point is to *block* un-grounded findings, not to rubber-stamp them.
- **Coordinate ordering with `pick#174`.** Lever 3 interacts with the Pending-node deadlock (`orchestrator.rs:190-196`): if real nodes arrive Pending and the Validator can't clear them, the report deadlocks. Sequence with the `pick#174` seams.
- **Prompt text stays, but is no longer the only defense.** Keep the `constants.rs` anti-fabrication rules; this issue adds the code-level backstop they currently lack.
- **Keep the `ToolResult` schema change backward-compatible on the wire.** Pick is a connector; the backend runner serializes `ToolResult` to JSON `{success, data, error, provenance}` over the Matrix/Strike48 wire. Add the new `completed`/`outcome` field **additively** (optional, defaulted) so the existing Matrix connector path keeps deserializing older/newer payloads without a lockstep deploy. Don't repurpose or remove `success`.

---

## References

- `crates/core/src/tools.rs:430,441,447,460` - `execute_timed*` error demotion
- `crates/tools/src/execute_command.rs:120-126` - nonzero exit reported as success
- `crates/tools/src/external/rustscan.rs:135,150-193`, `ffuf.rs:186` - failed-vs-empty ambiguity
- `crates/core/src/orchestrator.rs:185,100,118` - `gate_for_report`, optional provenance
- `crates/core/src/evidence.rs:153,172,181` - `EvidenceNode` provenance opt-in
- `crates/tools/src/lib.rs:241` - `inject_test_evidence` registered unconditionally
- `crates/ui/src/components/chat_panel/constants.rs:707,892,895` - prompt-only anti-fabrication
- `pick#172`, `pick#174` (evidence pipeline), `matrix#2851` (loud sibling), red-team-037 (Windows triggers)

---

**Proposed labels:** `product/pick`, `type/bug`, `priority/P0`, `feature/evidence-chains`, `feature/security`, `area/orchestrator`, `size/L`

**Risk:** Medium. Lever 1 is trivial and high-value. Levers 2-3 touch result/gate semantics; mitigated by preserving truthful zero-finding results and sequencing lever 3 with `pick#174`.

jtomek-strike48:
## Summary

Fixes #184. For a pentest platform whose premise is "no hallucinations — ground every claim in real tool output," the agent could present findings that never happened. Three independent code weaknesses combined to allow it; this PR closes each with a code-level backstop. Prompt anti-fabrication text stays, but is no longer the only defense.

Delivered as three reviewable commits (one per lever).

### Lever 1 — remove the production fabrication path (`1e63bd5`)
`inject_test_evidence` pushes arbitrary, caller-supplied findings (with fabricated provenance) straight into the evidence graph, and it shipped registered in every build. Gated behind a new off-by-default `inject-test-evidence` cargo feature, so release/default builds never register or advertise it. The module stays compiled (its own unit test still builds); only the registry registration and re-export are gated.

A cargo feature (not `cfg(debug_assertions)`) is deliberate: it is deterministically testable in both directions and decoupled from the build profile. The security-critical assertion — tool absent from the default registry — runs in normal CI.

### Lever 2 — distinguish "failed" from "empty" (`608e5db`)
A degraded/failed tool run was indistinguishable from a genuine zero-finding run: only a `success` bool separated them, and several tools set `success:true` on functional failure. That ambiguity is the fabrication substrate — the model can't tell "ran, found nothing" from "never ran," so it fills the gap.

- Additive, wire-compatible `ToolOutcome { Ran, Failed, Skipped }` on `ToolResult` (`#[serde(default)]` → older connector payloads without the field deserialize as `Ran`; `success` unchanged). `with_outcome` enforces the invariant that Failed/Skipped is never a success.
- `execute_command`: marks a run `Failed` on timeout, on nonzero exit **with empty stdout**, or on unreadable exit status (fail closed) — while deliberately keeping a nonzero exit that produced stdout as `Ran` (grep/diff/test exit nonzero legitimately). Unsupported platform → `Skipped`. stdout/stderr/exit_code stay in `data`.
- Scanner wrappers (rustscan, ffuf, masscan): any nonzero exit fails closed instead of parsing partial/empty output as a full result. A genuine zero-finding scan still exits 0 and yields a truthful empty set.
- Red-team prompt: a non-negotiable grounding rule — `failed`/`skipped` results are "no data"; never create a node or narrate a finding from one. (The schema field is inert without this clause.)

### Lever 3 — require grounding at the report gate (`e31b55d`)
`gate_for_report` refused Pending nodes but did not require provenance; `EvidenceNode.provenance` is `Option` and `None` passed. It now fails closed on any publishable finding (Confirmed/Revised) whose provenance is `None`, via a new `GateError::UngroundedFindings { ids }`. Anchored on `is_publishable_finding()` (not a free-form `node_type` string); InfoOnly/FalsePositive/context nodes are exempt. Ordered after the Pending gate. **Never** auto-attaches synthetic provenance — it blocks, per the issue guardrail.

## Why this is safe (verified preconditions)

- The #172/#174 Validator round-trip is live on `main` (PR #190), so a real node can be cleared from `Pending` — Lever 3 does not weld a deadlock shut.
- Every finding-producing tool already attaches provenance (verified across all 8 evidence-producer sites), so the gate passes legitimate findings and blocks only ungrounded ones. Fixtures updated to match production reality.

## Scope limit (please note)

These levers make the **report** trustworthy (report ← manifest ← gated graph ← provenance-bearing tool pushes). They do **not** fully close the **chat-narration** path: the red-team agent's free-text reply can still *claim* a finding that never entered the graph — there is no code seam between LLM conversational text and the operator. Post-fix the *report* is grounded; the *transcript* remains prompt-governed. A follow-up (surface `outcome=Failed` in the transcript, or a narration grounding cross-check) should track that. This PR does not claim fabrication is now impossible everywhere — it closes the report path and the three named code weaknesses.

## Test plan

- [x] `cargo clippy -p pentest-core -p pentest-tools -p pentest-ui --features connector,desktop --all-targets -- -D warnings` — clean (both `inject-test-evidence` on and off)
- [x] `cargo fmt --all -- --check` — clean
- [x] Lever 1: `inject_test_evidence` absent from default registry, present under the feature (both states tested)
- [x] Lever 2: `ToolOutcome` constructor mapping, `with_outcome` invariant, snake_case + legacy-JSON wire compat, `execute_timed` Err→Failed, full `classify_command_outcome` truth table (incl. fail-closed on missing exit_code)
- [x] Lever 3: gate rejects ungrounded Confirmed/Revised, accepts grounded, exempts non-published nodes, reports Pending before Ungrounded
- [x] Production-path integration test (`evidence_pipeline.rs`): a real failed `execute_command` surfaces `outcome=Failed`, and an ungrounded finding cannot pass `gate_for_report` — both halves of #184 proven on the non-test path
- [x] Full `pentest-core` + `pentest-tools` + `pentest-ui` suites green; existing tests unchanged

## Deferred (by design, follow-ups)

- Lever 2 applied to `execute_command` + rustscan/ffuf/masscan (the port/dir-scan finding vectors). The remaining ~20 scanner/OSINT wrappers should get the same treatment in a follow-up — kept out of this PR to stay reviewable.
- Chat-narration grounding (the scope limit above).

### DIFF
--- a/MANUAL_TEST_THREE_AGENT_PIPELINE.md
+++ b/MANUAL_TEST_THREE_AGENT_PIPELINE.md
@@ -12,11 +12,19 @@ git fetch fork feature/three-agent-pipeline
 git checkout feature/three-agent-pipeline
 
 # 2. Build
-cargo build
+#    The `inject_test_evidence` tool used below is a dev/test-only fabrication
+#    path (pick#184) and is NOT registered unless you opt in at build time.
+#    Enable the feature so the steps that call it work:
+cargo build --features pentest-tools/inject-test-evidence
 
 # 3. Verify environment
 cat .env | grep -E "STRIKE48_HOST|TENANT_ID"
 ```
+
+> **Note (pick#184):** `inject_test_evidence` is only registered when the
+> connector is built with the `inject-test-evidence` feature. A default/release
+> build will not expose it (the tool would reply "Unknown tool"). The real-tool
+> test suites below (nmap, service banner, whatweb) do not need the feature.
 
 ## Quick Smoke Test (5 minutes)
 
--- a/crates/core/src/orchestrator.rs
+++ b/crates/core/src/orchestrator.rs
@@ -168,6 +168,17 @@ pub enum GateError {
     /// highlight the offenders.
     #[error("{} evidence node(s) are still pending validation: {}", .pending_ids.len(), .pending_ids.join(", "))]
     PendingNodes { pending_ids: Vec<String> },
+
+    /// At least one publishable finding lacks [`Provenance`] tying it to real
+    /// tool output. The report must never present a finding the agent asserted
+    /// without a grounding tool result, so the gate fails closed (pick#184).
+    /// The offending node IDs are included so the UI can name them.
+    ///
+    /// The remedy is to ground the finding in a real tool run — never to
+    /// attach synthetic provenance to satisfy the gate, which would defeat the
+    /// guardrail.
+    #[error("{} finding(s) lack tool-output provenance and cannot be published (ungrounded): {}", .ids.len(), .ids.join(", "))]
+    UngroundedFindings { ids: Vec<String> },
 }
 
 /// Build a [`ValidatedFindingsManifest`] from the current evidence graph.
@@ -193,6 +204,24 @@ pub fn gate_for_report(
         .collect();
     if !pending_ids.is_empty() {
         return Err(GateError::PendingNodes { pending_ids });
+    }
+
+    // Fail closed on ungrounded findings (pick#184). A finding that will be
+    // published (Confirmed/Revised) must carry `Provenance` tying it to real
+    // tool output; otherwise it is a claim with no backing tool result and must
+    // never reach the report. We anchor on `is_publishable_finding()` rather
+    // than a `node_type` string (which is free-form) so this catches every
+    // finding headed for the report. InfoOnly / FalsePositive / context nodes
+    // are intentionally exempt — they are not published findings. This check
+    // follows the Pending gate: an un-adjudicated node is reported as Pending
+    // first, since its provenance is not yet the operator's concern.
+    let ungrounded: Vec<String> = nodes
+        .iter()
+        .filter(|n| n.is_publishable_finding() && n.provenance.is_none())
+        .map(|n| n.id.clone())
+        .collect();
+    if !ungrounded.is_empty() {
+        return Err(GateError::UngroundedFindings { ids: ungrounded });
     }
 
     let findings: Vec<ManifestFinding> = nodes
@@ -459,6 +488,19 @@ mod tests {
         EngagementInfo::new("10.0.0.0/24", ts()).with_completed_at(ts())
     }
 
+    /// Provenance a real finding-producing tool would attach. Every finding
+    /// node in production carries one (verified across all evidence producers),
+    /// so test fixtures for publishable findings attach it too — otherwise the
+    /// fail-closed grounding gate (pick#184) would reject them.
+    fn provenance() -> Provenance {
+        Provenance::new(
+            "nmap",
+            "7.95",
+            ProbeCommand::from_exact("nmap -sV 10.0.0.1"),
+            "Nmap scan report",
+        )
+    }
+
     fn confirmed_finding(id: &str, sev: Severity) -> EvidenceNode {
         let mut n = EvidenceNode::new(
             id,
@@ -468,7 +510,8 @@ mod tests {
             "10.0.0.1",
             sev,
             "initial rationale",
-        );
+        )
+        .with_provenance(provenance());
         n.apply_validator_decision(sev, "validator confirmed");
         n
     }
@@ -482,7 +525,8 @@ mod tests {
             "10.0.0.1",
             from,
             "initial rationale",
-        );
+        )
+        .with_provenance(provenance());
         n.apply_validator_decision(to, "severity adjusted after reproducing");
         n
     }
@@ -561,6 +605,7 @@ mod tests {
             GateError::PendingNodes { pending_ids } => {
                 assert_eq!(pending_ids, vec!["p1", "p2"]);
             }
+            other => panic!("expected PendingNodes, got {other:?}"),
         }
     }
 
@@ -636,19 +681,77 @@ mod tests {
 
     #[test]
     fn manifest_preserves_provenance_for_publishable_findings() {
-        let mut n = confirmed_finding("c1", Severity::High);
-        n.provenance = Some(Provenance::new(
-            "nmap",
-            "7.95",
-            ProbeCommand::from_exact("nmap -sV 10.0.0.1"),
-            "Nmap scan report",
-        ));
+        // confirmed_finding already carries provenance (as every real finding
+        // does); assert the gate passes it through into the manifest.
+        let n = confirmed_finding("c1", Severity::High);
         let manifest = gate_for_report(&[n], engagement()).unwrap();
         let prov = manifest.findings[0]
             .provenance
             .as_ref()
             .expect("provenance preserved in manifest");
         assert_eq!(prov.underlying_tool, "nmap");
+    }
+
+    // --- Fail-closed grounding gate (pick#184 Lever 3) --------------------
+
+    #[test]
+    fn gate_rejects_publishable_finding_without_provenance() {
+        // A Confirmed finding with no provenance is an ungrounded claim — the
+        // agent asserted it with no backing tool result. The gate must refuse.
+        let mut n = confirmed_finding("ungrounded", Severity::High);
+        n.provenance = None;
+        let err = gate_for_report(&[n], engagement()).unwrap_err();
+        match err {
+            GateError::UngroundedFindings { ids } => {
+                assert_eq!(ids, vec!["ungrounded".to_string()]);
+            }
+            other => panic!("expected UngroundedFindings, got {other:?}"),
+        }
+    }
+
+    #[test]
+    fn gate_rejects_revised_finding_without_provenance() {
+        // Revised is publishable too, so the grounding requirement applies.
+        let mut n = revised_finding("rev", Severity::High, Severity::Medium);
+        n.provenance = None;
+        let err = gate_for_report(&[n], engagement()).unwrap_err();
+        assert!(matches!(err, GateError::UngroundedFindings { .. }));
+    }
+
+    #[test]
+    fn gate_accepts_publishable_finding_with_provenance() {
+        // The happy path: a grounded, adjudicated finding publishes.
+        let manifest =
+            gate_for_report(&[confirmed_finding("c1", Severity::High)], engagement()).unwrap();
+        assert_eq!(manifest.findings.len(), 1);
+    }
+
+    #[test]
+    fn gate_ignores_missing_provenance_on_non_published_nodes() {
+        // InfoOnly and FalsePositive nodes are not published findings, so the
+        // grounding requirement does not apply — even without provenance the
+        // gate must not reject them (info_only/false_positive build no prov).
+        let nodes = [info_only("i1"), false_positive("fp1")];
+        let manifest =
+            gate_for_report(&nodes, engagement()).expect("non-published nodes need no provenance");
+        assert_eq!(manifest.findings.len(), 0);
+        assert_eq!(manifest.context_nodes.len(), 1);
+        assert_eq!(manifest.counts.false_positives, 1);
+    }
+
+    #[test]
+    fn gate_reports_pending_before_ungrounded() {
+        // A still-Pending node and an ungrounded published finding together:
+        // the Pending gate fires first (provenance is not yet the operator's
+        // concern for an un-adjudicated node).
+        let mut ungrounded = confirmed_finding("c1", Severity::High);
+        ungrounded.provenance = None;
+        let nodes = [pending("p1"), ungrounded];
+        let err = gate_for_report(&nodes, engagement()).unwrap_err();
+        assert!(
+            matches!(err, GateError::PendingNodes { .. }),
+            "Pending must be reported before UngroundedFindings, got {err:?}"
+        );
     }
 
     #[test]
--- a/crates/core/src/tools.rs
+++ b/crates/core/src/tools.rs
@@ -459,6 +459,36 @@ impl ToolSchema {
     }
 }
 
+/// Whether a tool actually completed a probe, distinct from whether it
+/// "succeeded".
+///
+/// `success` is a legacy boolean that conflates two very different things: a
+/// tool that ran and legitimately found nothing (`success:true`, empty data)
+/// vs. one that failed to probe at all (which several wrappers *also* reported
+/// as `success:true`). That ambiguity is the fabrication substrate #184
+/// closes: the model cannot tell "ran, found nothing" from "never ran," so it
+/// fills the gap. `ToolOutcome` makes the distinction explicit and trustworthy.
+///
+/// Serialized `snake_case` and defaulted to [`ToolOutcome::Ran`] so payloads
+/// from older connectors (which omit the field) deserialize unchanged over the
+/// Matrix wire — no lockstep backend deploy required.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Default, Serialize, Deserialize)]
+#[serde(rename_all = "snake_case")]
+pub enum ToolOutcome {
+    /// The tool ran to completion; `data` is trustworthy. A zero-finding scan
+    /// that genuinely ran is still `Ran` (a truthful empty result).
+    #[default]
+    Ran,
+    /// The tool did not complete a real probe (nonzero exit, timeout,
+    /// unparseable output, crash). `data` is NOT trustworthy — the model and
+    /// the report gate must treat it as "no data," never as a finding.
+    Failed,
+    /// A precondition was not met (missing dependency, unsupported platform),
+    /// so the probe never started. Distinct from `Failed`: nothing went wrong,
+    /// the work simply did not run.
+    Skipped,
+}
+
 /// Result from a tool execution.
 ///
 /// `provenance` is `Option` because not every tool produces a finding —
@@ -466,6 +496,11 @@ impl ToolSchema {
 /// Tools that produce findings (scanners, probes, exploits) must attach a
 /// `Provenance` so the Report Agent can render a reproducible evidence
 /// block. See [`crate::provenance`] and GitHub issue #52.
+///
+/// `outcome` distinguishes "ran" from "failed"/"skipped" so a functional
+/// failure can never masquerade as a legitimate empty result (#184). It is
+/// additive and defaulted for wire compatibility; `success` is retained
+/// unchanged.
 #[derive(Debug, Clone, Serialize, Deserialize)]
 pub struct ToolResult {
     pub success: bool,
@@ -474,6 +509,9 @@ pub struct ToolResult {
     pub duration_ms: u64,
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub provenance: Option<Provenance>,
+    /// Whether the tool actually completed a probe. See [`ToolOutcome`].
+    #[serde(default)]
+    pub outcome: ToolOutcome,
 }
 
 impl ToolResult {
@@ -485,6 +523,7 @@ impl ToolResult {
             error: None,
             duration_ms: 0,
             provenance: None,
+            outcome: ToolOutcome::Ran,
         }
     }
 
@@ -496,10 +535,15 @@ impl ToolResult {
             error: None,
             duration_ms,
             provenance: None,
+            outcome: ToolOutcome::Ran,
         }
     }
 
-    /// Create an error result
+    /// Create an error result.
+    ///
+    /// An error means the tool did not complete a real probe, so the outcome
+    /// is [`ToolOutcome::Failed`] — the structured "this did not run" signal
+    /// the model and report gate rely on (#184).
     pub fn error(message: impl Into<String>) -> Self {
         Self {
             success: false,
@@ -507,6 +551,7 @@ impl ToolResult {
             error: Some(message.into()),
             duration_ms: 0,
             provenance: None,
+            outcome: ToolOutcome::Failed,
         }
     }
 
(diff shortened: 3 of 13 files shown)


ngdevo:
PR #240 — Fail-closed grounding guardrail (issue #184)

## PR #240 Review — Fail-closed grounding guardrail (issue #184)

Reviewed the full diff (13 files, +634/-33). Overall: this is a well-designed, defense-in-depth patch. The three
levers are independently valuable, tests are thorough, and the author correctly acknowledges scope limits in the
PR description. Below are findings by severity for both the Rust and security aspects.

---

### CRITICAL
None. The security-critical assertion — `inject_test_evidence` absent from the default registry — is directly
tested in the config that CI actually runs (`crates/tools/src/lib.rs`,
`inject_test_evidence_absent_from_default_registry`).

---

### HIGH

**H1 — Scanner-wrapper regression risk (rustscan/ffuf/masscan)**
`crates/tools/src/external/{rustscan,ffuf,masscan}.rs` — the check changed from
`exit_code != 0 && stdout.is_empty()` to `exit_code != 0`. RustScan in particular is known to return nonzero exit
codes on partial/warning conditions where the port list on stdout is still meaningful (e.g., unreachable-host
warnings, ulimit adjustments, its internal nmap exec exit propagation). Under the new rule those runs become
`Failed`, discarding real output. This is the *correct* fail-closed choice for the guardrail, but you should:
- Enumerate real-world exit codes for each of the three tools before merging (or in an immediate follow-up), and
- Consider allowlisting specific known-benign nonzero codes rather than a blanket `!= 0`.
Otherwise expect a spike in `Failed` outcomes on legitimate scans and false negatives in coverage.

**H2 — Wire-compat requires no back-fill of legacy connectors, but success remains authoritative for them**
`crates/core/src/tools.rs` — `#[serde(default)]` on `outcome` correctly deserializes older payloads as `Ran`. That
means the fabrication substrate still exists for any old connector version in the fleet until it's upgraded: it
will send `success: true` on a functional failure and the new gate/model will trust it. The PR body doesn't spell
this out. Recommend a version bump on the connector protocol or a monitoring signal that alerts when an
"outcome-less" payload is received from a supposedly-upgraded connector.

**H3 — Provenance is unforgeable to a friendly caller, not to a hostile one**
`crates/core/src/orchestrator.rs` — `gate_for_report` only checks `provenance.is_some()`. A `Provenance` is
constructed with plain strings; there is no cryptographic binding between the provenance struct and an actual
tool invocation. If any code path can push an `EvidenceNode` with a hand-built `Provenance::new("nmap", ...)`,
Lever 3 is silently bypassed. Lever 1 removes the *known* offender (`inject_test_evidence`), and every real
evidence-producer today is trusted — but this is only an assertion by inspection. Two concrete asks:
1. Add a `#[deny(dead_code)]`-style tripwire or an architectural test enumerating all callers of
   `EvidenceNode::with_provenance` / any function that writes to the graph. Fail CI on new call sites without
   review.
2. Long-term: consider making `Provenance` carry a token derived from the actual tool execution (e.g., an HMAC
   over `(tool_name, params_hash, run_id)` computed inside `execute_timed_with_provenance`) so a hand-rolled
   `Provenance::new(...)` cannot pass the gate.

**H4 — Chat-narration path still allows fabrication (acknowledged, but calling out explicitly for the risk log)**
`crates/ui/src/components/chat_panel/constants.rs` — the grounding rule is now in the prompt, but it is only a
prompt. The report is trustworthy after this PR; the *conversation surface the operator reads first* is not. For
a "no-hallucination" product this is the more visible failure mode. Track as a P1 follow-up (per the PR
description) and consider a UI-side redaction of finding-shaped assertions the transcript makes that aren't in
the graph. This is a defense-in-depth gap, not a correctness bug.

---

### MEDIUM

**M1 — `execute_command` Err-propagation from `execute_timed_with_provenance`**
`crates/tools/src/execute_command.rs:~135` — after the refactor:
`let run = execute_timed_with_provenance(|| async move { … }).await?;`. The `?` propagates any inner `Err` out of
`execute()`. This is only safe if `execute_timed_with_provenance` mirrors `execute_timed`'s contract of wrapping
Err into `Ok(ToolResult::error(...))`. The PR tests `execute_timed`'s Err arm
(`execute_timed_err_arm_yields_failed_outcome`) but does *not* test `execute_timed_with_provenance`'s Err arm. If
those helpers diverge, an inner Err escapes classification. Add an equivalent test for
`execute_timed_with_provenance`, or replace `?` with an explicit `unwrap_or_else` that stamps `Failed`.

**M2 — `Skipped` skips timing and provenance entirely**
`crates/tools/src/execute_command.rs:~64` — the early-return for unsupported platforms returns
`ToolResult::skipped(...)` with `duration_ms: 0` and no provenance. That's semantically correct (nothing ran),
but the code path completely bypasses `execute_timed_with_provenance`, so any cross-cutting behavior that helper
adds (metrics, tracing spans, structured logging) is skipped too. If any downstream observer expects every tool
invocation to emit a span, this silently loses events. Confirm nothing else in the pipeline relies on that.

**M3 — `stdout_empty` when `data.stdout` is not a string**
`crates/tools/src/execute_command.rs:~168` — `unwrap_or(true)` correctly biases toward fail-closed if the field
is missing/wrong type, but combined with the `exit_code.is_none() → Failed` branch you can double-count
"malformed data shape" as `Failed`. Fine for the guardrail; worth logging one warning line when this branch
fires so a genuine data-shape regression doesn't hide as "just another failed probe."

**M4 — `with_outcome` invariant is one-directional**
`crates/core/src/tools.rs:~587` — `with_outcome(Failed|Skipped)` forces `success = false`, but the reverse
contradiction is still representable: a caller can construct `ToolResult { success: false, outcome: Ran, … }` by
hand (or via `ToolResult::error(...).with_outcome(Ran)`). Not exploitable given current callers, but the
invariant advertised in the docstring ("`success=true, outcome=Failed` is unrepresentable through this builder")
isn't the full invariant. Consider making `ToolOutcome` fields private and forcing all construction through the
builder, or adding a `debug_assert!(success == matches!(outcome, Ran))` in a validation helper.

**M5 — `false_positive` / `info_only` exemption in the gate**
`crates/core/src/orchestrator.rs:~215` — nodes that aren't `is_publishable_finding()` skip the grounding check
entirely. Correct today, but if `is_publishable_finding()`'s definition ever widens (or a new node type is added
and forgotten), findings could slip past the gate. Add a match-exhaustive test that iterates every
`EvidenceNodeType` variant and asserts each is either "publishable → requires provenance" or "not publishable →
exempt." Prevents silent regressions when new variants are added.

**M6 — No CI guard against `--features inject-test-evidence` in release**
`crates/tools/Cargo.toml` — the feature is off by default, but nothing prevents a future workflow, Dockerfile,
or `cargo install` invocation from turning it on. Add an explicit release-build CI job that runs
`cargo tree -e features -p pentest-tools` (or equivalent) and greps for `inject-test-evidence`. Or ship a
compile-time canary type gated on `#[cfg(all(not(debug_assertions), feature = "inject-test-evidence"))]` that
intentionally fails to build.

---

### LOW

**L1 — Serde field ordering / stability**
`ToolResult` now has an additional field. If any code deserializes with `deny_unknown_fields` upstream
(unlikely, but worth grepping), it would break. Quick `rg 'deny_unknown_fields'` in the tree confirms none in
this diff, but worth a spot-check.

**L2 — Error message shape change in scanner wrappers**
The wrappers now emit `"rustscan failed (exit {}): {}"` — this is a user-visible message change. Any log-based
alerting or test asserting on the old string would break. Trivial to fix if it exists.

**L3 — Test lint hygiene**
`crates/ui/tests/evidence_pipeline.rs` — the test uses `session::evidence_snapshot()` and pushes to a shared
graph. The `serial()` guard is correctly acquired only *after* the `.await` — good comment explaining why.
Consider moving the reason ("std MutexGuard is not Send across await") into a shared helper docstring so future
contributors don't accidentally invert the order.

**L4 — Docstring says "eight evidence-producer sites"; verify count**
PR body claims "verified across all 8 evidence-producer sites." Numbers in docstrings age poorly. Consider
adding an assertion or an audit test that fails if any new call site to `push_evidence` appears without a
matching provenance attachment.

**L5 — `MANUAL_TEST_THREE_AGENT_PIPELINE.md` note position**
The added note is placed after the build step and might be missed by an operator who reads top-to-bottom and
runs `cargo build` before scrolling. Move the `> **Note (pick#184):**` block above the
`cargo build --features ...` command, or make it a fenced warning at the very top.

---

### INFO / POSITIVE OBSERVATIONS

- **Independent commits per lever**: excellent for review and for revert-in-place if any single lever regresses.
- **`#[serde(default)]` + additive field**: the *right* wire-compat choice for a phased connector rollout.
- **`#[must_use]` on `with_outcome`**: prevents the silent `.with_outcome(Failed);` bug.
- **Deliberate `Ran` for nonzero-exit-with-stdout**: correctly preserves grep/diff/test semantics, which naive
  fail-closed checks routinely break.
- **Test that Pending precedes UngroundedFindings**: pins user-facing error message ordering, which is a
  fabrication-adjacent concern (don't leak "we didn't have provenance" for a node the operator hasn't even
  adjudicated yet).
- **Cargo feature (not `cfg(debug_assertions)`)**: correct — deterministically testable in both directions and
  independent of `--release`.

---

### SUMMARY RECOMMENDATIONS BEFORE MERGE

1. **Add a test for `execute_timed_with_provenance`'s Err arm** (M1) — small, high-value.
2. **Verify rustscan/ffuf/masscan real-world exit codes** and consider an allowlist (H1) — otherwise expect
   false-positive `Failed`s.
3. **Add a CI job asserting release builds do not enable `inject-test-evidence`** (M6).
4. **File follow-ups** for: H3 (provenance authenticity), H4 (chat-narration grounding), and the ~20 remaining
   scanner wrappers named in the PR body.
5. Everything else can land as-is.

The PR meaningfully closes the three named weaknesses at the code layer and is safe to merge with M1 addressed
and H1 investigated. Nice defense-in-depth work.

jtomek-strike48 (commented):
## Adversarial review — APPROVE (sound; ready pending team approval)

Reviewed the whole PR against `main` (cross-lever composition, completeness, wire-compat, test quality), not just commit-by-commit. The guardrail holds.

**Cross-lever composition is correct.** A failed tool (Lever 2) produces no provenance, so any node derived from it is caught at the gate (Lever 3); the fabrication tool is unreachable in release (Lever 1, type not even nameable). The integration test proves this end-to-end on the production path.

**Completeness checked.** Every `EvidenceNode` creation site was audited: the tool-result path is fully closed (provenance-gated), `inject_test_evidence` is feature-gated, and the remaining node-creation sites are test-only. `is_publishable_finding()` is the correct anchor — the Red Team can't bypass by manipulating `ValidationStatus` (the Validator owns that transition).

**Wire compat verified.** `ToolOutcome` is additive + `#[serde(default)]`=Ran (legacy payloads deserialize unchanged; test covers it). `GateError::UngroundedFindings` is a standard additive enum variant.

### One flagged item, investigated and dismissed as a blocker
An automated pass flagged `evidence_flows_through_buffer` (`crates/tools/src/evidence_producer.rs`) as a CRITICAL failing test (missing serial guard on the shared `PENDING_EVIDENCE` static). I could **not reproduce it**: the full `pentest-tools` lib suite passes in parallel (279 passed), the test passes in isolation, and CI's Test job is green. It is also **not touched by this PR** — the test exists unchanged on `main`. Conclusion: a transient race in the reviewer's own concurrent environment, not a defect here. It is a *latent* flake risk worth a follow-up (add the `serial()` guard the sibling evidence tests use), but not a merge blocker for #240.

### Residual items (already documented in the PR body, restating for the record)
- **Chat-narration path** is not closed — the LLM's free-text reply can still *claim* an ungrounded finding, since there's no code seam between conversational text and the operator. The report itself stays grounded (gate blocks it); the transcript remains prompt-governed. Worth a follow-up (e.g. an evidence `source` enum, or requiring every `push_evidence` to originate from a tool result).
- **~20 remaining scanner/OSINT wrappers** still parse partial output on nonzero exit. Not a security hole (the gate still blocks ungrounded findings), but a quality follow-up to extend the rustscan/ffuf/masscan treatment.

Recommend filing those two as follow-up issues. No changes required on this PR.

cchao-devo (commented):
Approve. Checked out e31b55d and verified all three levers empirically (fmt + clippy clean under connector,desktop, both feature states).

  - Lever 1: inject_test_evidence absent from the default registry (feature off) and present with the opt-in feature — both assertions pass. Confirmed it's registered on main but gone here.
  - Lever 2: ToolOutcome constructor/invariant/wire-compat (8/8) + classify_command_outcome full truth table (13/13), incl. fail-closed on missing exit_code and nonzero-with-stdout staying
  Ran.
  - Lever 3: gate 9/9 incl. Pending-before-Ungrounded ordering; nice that the ungrounded check reuses is_publishable_finding() so it covers exactly the published set.
  - Production path: evidence_pipeline integration test green.

  Strongest signal: I neutered the gate's ungrounded early-return and both the unit tests and the production-path integration test went red — the latter with a provenance: None "Fabricated
  admin panel" reaching the manifest. The tests genuinely guard the guardrail. 🚢

  Two pre-existing failures I want to flag as unrelated (both reproduce on main, not introduced here): all_tools_registered fails on autopwn_crack, and
  evidence_producer::tests::evidence_flows_through_buffer is flaky (shared global evidence buffer — non-deterministic across runs). Worth a follow-up, especially the flake, but not blocking.
  The documented scope limit (chat-narration path still prompt-governed) is the right call to defer.

cchao-devo (approved):
LGTM

PR #246, opened 2026-07-16
fix(#245): sync pacman DB before installing tools (-Sy)

### LINKED ISSUE #245: Tool install fails on stale rootfs pacman DB (pacman -S without -Sy)
## Summary

Installing a tool into the sandbox via `pacman -S <pkg>` fails when the rootfs package databases are stale/unsynced, because the installers run `pacman -S` without `-Sy`.

## Observed

With a working sandbox (proot on an AppArmor-restricted host), installing bloodhound:

```
Tool execution error: Failed to install python-bloodhound:
warning: database file for 'extra' does not exist (use '-Sy' to download)
warning: database file for 'blackarch' does not exist (use '-Sy' to download)
error: target not found: python-bloodhound
```

The sync DB files physically exist under `<rootfs>/var/lib/pacman/sync/` but pacman reports them missing/stale because the per-install command does not refresh them.

## Root cause

- The per-tool installers run `pacman -S --noconfirm <pkg>` (e.g. `crates/tools/src/installers/bloodhound.rs:71-74`, and `package.rs`) with no `-Sy`.
- Database refresh only happens in one-time rootfs provisioning (`crates/platform/src/desktop/sandbox/rootfs.rs`: `initial_pacman_sync` = `pacman -Sy`, `initial_system_update` = `pacman -Syu`). If that provisioning did not run, partially failed, or the DBs have since gone stale, later `-S` installs have nothing to resolve against.

This is independent of sandbox backend selection (#238 / #239) — the sandbox runs pacman correctly; pacman just has no fresh DB.

## Proposed fix

Options (pick during design):

1. Make tool installs sync-then-install: `pacman -Sy --noconfirm <pkg>` (or a `pacman -Sy` immediately before `-S`). Simplest; refreshes per install.
2. Ensure `ensure_ready()` guarantees a DB sync when the sync dir is empty/stale, so installs can assume a fresh DB.
3. Both: a cheap staleness check in `ensure_ready` plus `-Sy` on the install path as a backstop.

Prefer 1 (or 1+2) so a first-run or stale rootfs self-heals instead of erroring.

## Acceptance criteria

- [ ] Installing a tool into a freshly-provisioned or stale rootfs succeeds without a manual `pacman -Sy`.
- [ ] "Install all recommended" works end-to-end on a clean rootfs.
- [ ] No regression when the DB is already fresh (avoid a full `-Syu` per install).

## Context

Surfaced during #238 / #239 verification once the sandbox itself was fixed (proot selected on an AppArmor host). Separate subsystem: rootfs pacman provisioning, not backend selection.

jtomek-strike48:
## Summary

Fixes #245. Installing a tool into the sandbox failed when the rootfs pacman databases were stale/unsynced:

```
error: target not found: python-bloodhound
```

The installers ran `pacman -S <pkg>` without refreshing the package DB, so a rootfs whose sync databases were not current had nothing to resolve against.

## Change

`crates/tools/src/external/install.rs` now uses `pacman -Sy --noconfirm --needed` (refresh DB, then install) instead of a bare `-S` in both install sites:

- `ensure_tool_installed` (single) — the path used by ~50 external tool wrappers (nmap, masscan, hydra, nikto, the network/web/specialized/postexploit/wireless families, ...)
- `install_tools_batch` (batch)

`--needed` keeps re-runs idempotent (pacman skips an already-current package).

## Scope note (rebased onto main)

This branch originally also patched the four sandbox installers (`bloodhound`, `metasploit`, `package`, `zap`). Since then, `main` (#239/#253) refactored those installers to route through a shared `installers::pacman::install` helper that already runs `pacman -Sy --noconfirm --needed`. Those changes are now redundant, so this branch has been rebased and reduced to the one remaining unfixed site: `external/install.rs`. Its flags and comment are aligned with that shared helper so the crate carries a single pacman-install convention.

## Why `-Sy` and not `-Syu`

`-Sy` fixes the missing-DB error and self-heals a stale rootfs without a full system upgrade on every install. It matches the shared `installers::pacman` helper and `RootfsManager::initial_pacman_sync` (`pacman -Sy`), and the rootfs already gets a `-Syu` at provisioning, so the partial-upgrade risk is acceptable for this ephemeral pentest sandbox. A full `-Syu` per install was explicitly ruled out (slow).

## Test plan

- [x] `cargo check -p pentest-tools --features desktop --lib` clean
- [x] `cargo clippy -p pentest-tools --features desktop --all-targets -- -D warnings` clean
- [x] `cargo test -p pentest-tools --features desktop --lib` — 293 pass
- [ ] Manual: install a tool (e.g. nmap) into a sandbox with a stale rootfs DB and confirm it succeeds

### DIFF
--- a/crates/tools/src/external/install.rs
+++ b/crates/tools/src/external/install.rs
@@ -70,10 +70,15 @@ pub async fn ensure_tool_installed(
         "Installing '{}' via pacman package '{}'...",
         binary_name, pacman_package
     );
+    // `-Sy` (refresh DB, then install) rather than a bare `-S`: a rootfs whose
+    // sync databases are stale or were never synced makes `-S` fail with
+    // "database file for 'core' does not exist ... target not found". `--needed`
+    // keeps re-runs idempotent. This mirrors the `installers::pacman` helper,
+    // which documents the full rationale for the sandbox installers.
     let install = platform
         .execute_command(
             "pacman",
-            &["-S", "--noconfirm", pacman_package],
+            &["-Sy", "--noconfirm", "--needed", pacman_package],
             Duration::from_secs(300),
         )
         .await?;
@@ -137,7 +142,9 @@ pub async fn install_tools_batch(platform: &impl CommandExec, packages: &[&str])
 
     info!("Batch installing {} packages...", packages.len());
 
-    let mut args = vec!["-S", "--noconfirm"];
+    // `-Sy` (refresh DB then install) so a stale/unsynced rootfs DB doesn't make
+    // the whole batch fail with "target not found" (see ensure_tool_installed).
+    let mut args = vec!["-Sy", "--noconfirm", "--needed"];
     args.extend_from_slice(packages);
 
     let install = platform


jtomek-strike48:
Live-verification note: the `-Sy` change here works as intended — on an affected host (proot selected) the failure moved from "database file for 'core' does not exist" (stale DB, the bug this fixes) to a *different* error. So this PR is correct and necessary.

However, a **separate, deeper bug** sits underneath it: in the proot sandbox, `pacman -Sy` exits 0 but the synced databases are unreadable (`pacman -Sl` sees 0 packages), so installs still fail with `target not found`. Filed as #248 with full diagnostics. This PR is a prerequisite for working installs but is not sufficient on its own until #248 is resolved — so the "install a tool end-to-end" test-plan checkbox will stay unchecked until then. Merging this on its own is still correct (it fixes the stale-DB case).

cchao-devo (approved):
Review: Pick PR #246 — fix(#245): sync pacman DB before installing tools (-Sy)
  
  Verdict: ✅ Approve. Correct, minimal, well-justified one-line-per-site fix. It closes the stale-DB pacman -S failure and aligns the last remaining install site to the crate's
  existing -Sy --noconfirm --needed convention. No blockers.

  Lenses run: mechanical gates + manual source trace + consistency/consumer trace. Verified in pick_08/pick; pick_07 untouched, no cluster.

  Gates (on pr-246)

  - cargo fmt --all --check clean · cargo clippy -p pentest-tools --lib -- -D warnings clean · cargo check compiles.
  - pentest-tools --lib: 293 pass (matches the PR's claim) on clean runs. ⚠️  One caveat below.

  The change is correct and consistent

  - Both sites (ensure_tool_installed single-install, install_tools_batch) now use ["-Sy", "--noconfirm", "--needed", pkg] — byte-identical to the shared
  installers::pacman::install_args helper (verified at pacman.rs:27). The -Sy-not--Syu rationale, --needed idempotency, and stale-DB failure mode all match the shared helper's
  documented convention.
  - Grep confirms external/install.rs was genuinely the last bare--S install path. The only other pacman -S strings in the crate are operator-facing error hints ("run sudo pacman -S 
  ... on the host") and an unrelated evil-winrm -S flag — none are sandbox install invocations. The "single pacman-install convention" claim holds.
  - --needed is a pure safety-add (skips already-current packages); no behavior regression vs. bare -S. Arg-vector construction in the batch path (vec! + extend_from_slice) is sound.

  Two things to note (neither blocks)

  1. Pre-existing flaky tests, NOT this PR. evidence_producer::tests::{multiple_push_single_drain, evidence_flows_through_buffer} fail intermittently (~1 in 3 runs). I confirmed they
  flake identically on main — root cause is test-isolation: they share a process-global evidence buffer and drain_pending_evidence() in one test steals nodes pushed by a parallel test
  (the test's own comment acknowledges "may contain others from parallel tests"). Unrelated to install.rs, but worth a separate cleanup issue (e.g. serialize these tests with a mutex
  or use a per-test buffer) since it makes the suite non-deterministic in CI.
  2. PR is a prerequisite, not a complete fix — author already disclosed this. Jonathan's own PR comment notes the deeper #248 (pacman -Sy exits 0 but synced DBs are unreadable in the
  proot sandbox → installs still fail target not found). This PR correctly fixes the stale-DB case and is safe to merge on its own; end-to-end install won't work until #248 lands.
  That's honestly represented and the test-plan "install end-to-end" box is appropriately left unchecked.

  init-dev relevance

  Low. init-dev deploys headless with DISABLE_SANDBOX=true, so the sandbox install path this PR touches isn't exercised there (tools are expected pre-installed on host). No consumer
  impact — this is purely a sandbox-mode correctness fix.

Issue #244, opened 2026-07-16
Signal sandbox-disabled state in Tools UI; reconcile DISABLE_SANDBOX vs shell_mode


jtomek-strike48:
## Summary

When the sandbox is disabled, tool installation fails one-tool-at-a-time with per-tool errors instead of being signalled up front. Users get a wall of failures (or a confusing single failure) with no clear indication of *why* or how to proceed, and no way to tell which sandbox mode they are in.

## Observed

With `DISABLE_SANDBOX=true`, clicking install for a sandbox-installed tool returns, e.g.:

```
Tool execution error: bloodhound-python is not installed and sandbox is disabled.
Install the collector on the host with: pipx install bloodhound ...
```

And "Install all recommended" produces a long list of individual failures rather than one clear message.

## Root cause / context

There are **two independent sandbox switches that are not synchronized**, and neither is surfaced in the UI:

1. `DISABLE_SANDBOX` env var (`.env`) -> sets the process-global `USE_SANDBOX` (via `init_sandbox_from_env` in `crates/platform/src/desktop/command.rs`). Gates the **tool execution / install path** (`execute_command`, the per-tool installers under `crates/tools/src/installers/*`).
2. `shell_mode` setting (`Native`/`Proot`, settings.json) -> gates the **interactive PTY shell path** and is also mirrored into `set_use_sandbox()` by the desktop/web UI (`workspace_app.rs`, `router.rs`), but NOT by headless.

Because these are wired to different state and updated by different code, a session can be in a contradictory state (shell sandboxed, installs not) and the user has no visibility into it. There is also an ordering hazard: `is_sandbox_enabled()` runs `init_sandbox_from_env()` (a `Once`) while the UI may call `set_use_sandbox()`, so env and UI can fight over the same flag depending on call order.

The installers themselves are correct: each checks `sandbox_enabled()` and returns host-install `manual_instructions()` when off (see `bloodhound.rs`, `zap.rs`, `metasploit.rs`, `package.rs`). The gap is purely UX/visibility.

## Proposed improvements

1. **Surface sandbox state in the Tools UI.** Show whether the sandbox is enabled (and which backend was selected: bwrap/proot/none). A small status chip/banner.
2. **Proactive guard on install actions.** When the sandbox is disabled, disable or annotate the sandbox-install buttons with a tooltip ("Sandbox is off - enable it in Settings, or install this tool on the host"), instead of letting each install fail. For "Install all", show one banner rather than N per-tool errors.
[...]

Issue #241, opened 2026-07-16
Relabel Shell Mode toggle: 'Sandboxed'/'Native' instead of 'Proot'/'Native'


jtomek-strike48:
## Summary

The Shell Mode settings toggle presents two buttons, **"Native"** and **"Proot"** (`crates/ui/src/components/settings_page.rs:439-459`), with the sub-label "BlackArch proot" vs "Native shell". This leaks an implementation detail as if it were a user choice.

## Problem

The real user intent is binary: **run tools in the isolated BlackArch sandbox** vs **run on the host**. Users are not expected to know the difference between the two Linux sandbox engines (bwrap = kernel namespaces; proot = ptrace), and the app in fact chooses between them automatically at runtime (`preferred_backend_for` + `detect_backend`, and the symmetric EACCES fallback added in #239). The word "Proot" on the button therefore over-specifies: a user clicking it means "sandbox me", not "use the ptrace-based proot binary specifically".

This confusion was the root cause of the debate in #238 / #239 about whether selecting "Proot" should pin the proot backend.

## Proposed change

- Relabel the toggle **"Sandboxed" / "Native"** (or "Isolated" / "Host").
- Keep `ShellMode::{Native, Proot}` as-is internally for now, or rename `Proot -> Sandboxed` with a serde alias for backward-compatible settings deserialization (larger change; can be split out).
- Update the sub-label text accordingly ("BlackArch sandbox" instead of "BlackArch proot").

## Context

Follow-up from #239 (review discussion with cchao-devo/ngdevo). Not a blocker for that PR.

## Acceptance criteria

- [ ] Settings toggle no longer shows "Proot" as a user-facing label.
- [ ] Existing saved settings (`shell_mode = Proot`) still deserialize without loss.
- [ ] No behavioral change to backend selection.

jtomek-strike48:
Closing as completed. The relabel has already shipped on `main` (verified against `origin/main` at b9a43d9). All three acceptance criteria are satisfied in `crates/ui/src/components/settings_page.rs`:

- Toggle no longer shows "Proot" as a user-facing label: the button now reads "Sandboxed" (settings_page.rs:902) / "Native" (settings_page.rs:877), and the status sub-label reads "Sandboxed (isolated Linux)" / "Native (this host)" (settings_page.rs:859). No "Proot" or "BlackArch proot" text remains in the toggle.
[...]

PR #247, opened 2026-07-16
feat: default to sandboxed (proot) shell mode everywhere

jtomek-strike48:
## Summary

Make the sandboxed BlackArch environment (proot/bwrap) the **default** execution mode across all surfaces, so pentest tools are sandboxed out of the box instead of running on the host.

## Changes

- `ShellMode::default()`: `Native` -> `Proot`. This is the default for `AppSettings::default()` and any fresh/absent settings, so **headless** (which reads `shell_mode` straight from settings) also defaults to sandboxed.
- `apps/desktop` and `apps/web`: `default_proot` `false` -> `true`, matching `apps/mobile` (already `true`).

## Behavior

- **First-run only.** The per-app `default_proot` seed only applies when `last_config.is_none()`; it never overrides an existing user choice.
- **Existing users unaffected.** A persisted `shell_mode` deserializes to its stored value; only fresh installs pick up the new default.
- **User-reversible.** Native remains selectable in Settings.
- **On-demand rootfs download.** Selecting Proot triggers a BlackArch rootfs download the first time a shell/tool runs (`ensure_ready` -> `ensure_rootfs`). The UI already gates the Proot option behind `blackarch_downloaded` and downloads on demand — a known, handled path (mobile ships `default_proot=true`).

## Rationale

Sandboxing tool execution by default is the safer, higher-value posture for a pentest connector and matches the product intent. It also aligns desktop/web with mobile.

## Depends on / relates to

Builds on the #238/#239 sandbox-selection fixes (a working backend is now selected on AppArmor-restricted Linux hosts) and #245/#246 (tool installs sync the pacman DB). Those should land first so the default lands on a working sandbox.

Related follow-ups: #241 (relabel toggle), #244 (surface sandbox state / reconcile switches).

## Test plan

- [x] `cargo test -p pentest-core --lib` (256 pass)
- [x] `cargo check --workspace` clean
- [ ] Manual: fresh install (no prior settings) starts in Proot; existing settings with Native are preserved

### DIFF
--- a/apps/desktop/src/main.rs
+++ b/apps/desktop/src/main.rs
@@ -11,7 +11,7 @@ const DESKTOP_CONFIG: ConnectorAppConfig = ConnectorAppConfig {
     platform_name: "Desktop",
     container_class: "mobile-app",
     shell_route_mode: ShellMode::Native,
-    default_proot: false,
+    default_proot: true,
     start_liveview_server: true,
     inject_css: false,
     extra_init_messages: &[],
--- a/apps/web/src/main.rs
+++ b/apps/web/src/main.rs
@@ -12,7 +12,7 @@ const WEB_CONFIG: ConnectorAppConfig = ConnectorAppConfig {
     platform_name: "Web (Liveview)",
     container_class: "mobile-app",
     shell_route_mode: ShellMode::Native,
-    default_proot: false,
+    default_proot: true,
     start_liveview_server: false,
     inject_css: false,
     extra_init_messages: &["Tools execute on the server machine."],
--- a/crates/core/src/config.rs
+++ b/crates/core/src/config.rs
@@ -9,9 +9,10 @@ use crate::aggression::AggressionLevel;
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Serialize, Deserialize, Default)]
 pub enum ShellMode {
     /// Run commands directly on the host machine (native shell)
-    #[default]
     Native,
-    /// Run commands inside the proot BlackArch environment
+    /// Run commands inside the sandboxed BlackArch environment (bwrap/proot).
+    /// Default: the connector sandboxes tool execution out of the box.
+    #[default]
     Proot,
 }
 


cchao-devo (approved):
Review: Pick PR #247 — feat: default to sandboxed (proot) shell mode everywhere
  
  Verdict: ✅ Approve, with one non-blocking accuracy correction to the PR body. The change is small, correct, safe for init-dev, and the intent (sandbox-by-default) is sound. But the
  body overstates the migration safety — there's a narrow silent Native→Proot flip it claims can't happen. Worth a comment, not a block.

  Lenses run: mechanical gates + manual source trace + adversarial migration-safety agent, with the init-dev consumer trace done explicitly. Verified in pick_08/pick; pick_07
  untouched, no cluster.

  Gates (on pr-247)
  
  - cargo fmt --all --check clean · cargo check --workspace clean · pentest-core lib 256/256 pass. (clippy per PR body clean; check compiles the same tree.)

  init-dev consumer trace — this PR is a no-op for the cluster deployment ✅

  The headless connector init-dev deploys (apps/headless/src/main.rs) does not use connector_app(), has no default_proot/set_sandbox wiring, and never calls set_use_sandbox. Its
  tool-execution sandbox is governed solely by the platform static USE_SANDBOX = AtomicBool::new(true) + the DISABLE_SANDBOX env var (command.rs:20,28) — neither touched by this PR.
  init-dev sets pick.disableSandbox=true → DISABLE_SANDBOX=true → sandbox stays off (still correct: nmap file-caps need root, proot OOMs there). No crash-loop risk, no behavior change 
  for init-dev.

  PR-body accuracy — two claims need correcting
  
  1. "headless reads shell_mode straight from settings, so it also defaults to sandboxed." Inaccurate. Headless hardcodes ShellMode::Native for the shell-WS routes and never reads
  shell_mode into the sandbox flag. The outcome (headless sandboxes by default) is right, but via the platform static — not via shell_mode. The line should be reworded so a future
  reader doesn't assume headload honors the setting.
  2. "First-run only … never overrides an existing user choice." Not strictly true. The seed guard (connector_app.rs:255) is default_proot && shell_mode==Native && 
  last_config.is_none(). But last_config is persisted only when the user checks "remember" (:411-413), which defaults false (:695). So a desktop/web user who explicitly chose Native in
  Settings but never enabled remember/auto-connect has shell_mode=Native, last_config=None — and on upgrade the seed silently flips them to Proot + enables the sandbox, triggering a
  BlackArch rootfs download on first tool run. The guard can't distinguish "deliberately Native" from "unset default Native." (Also a narrower field-absent serde-default flip, but no
  released Pick build ever wrote a settings file lacking shell_mode, so that path is essentially theoretical.)

  Severity call

  The Claim-B flip is LOW–MEDIUM, non-blocking: it only hits desktop/web GUI users (not init-dev), Proot is the safer posture, and it's user-reversible — the harm is a surprise rootfs
  download + behavior change, not a security regression. But since the body explicitly promises the opposite, it's worth either (a) softening the claim, or (b) gating the seed on a
  dedicated first-run marker (e.g. shell_mode_seeded: bool, or reuse device-id creation) instead of last_config.is_none().

PR #249, opened 2026-07-17
Easy Mode + PLG mobile: Sage reskin, docs/share, OAuth-first connector registration, telemetry (+ nix/dx build + Android CI)

### LINKED ISSUE #278: Add Sentry telemetry to Pick (sessions + per-activity events)
Parent: Strike48/project-management#72 · Init: init/plg

## Add Sentry telemetry to Pick

Pick currently ships **no** Sentry/telemetry. Add it from scratch, mirroring StrikeHub's
`sh-core/src/sentry_init.rs` pattern (release-health sessions + business-event spans + platform
tags) so we can answer **who is using Pick and how**.

Implements the shared spec from Strike48/project-management#72 (event taxonomy + privacy/PII model).

### Scope
- [x] Add `sentry` crate + a `sentry_init`-equivalent module (DSN via `build-defaults.toml`,
      environment from build profile, `auto_session_tracking` for DAU/WAU + crash-free rate).
- [x] Tag platform context: `app.mode`, `app.platform`, `app.arch`, `app.channel` (easy/advanced),
      release version.
- [x] Emit **per-activity events** for each tool run / major action (e.g. `tool.run`,
      `network.check`, `scan.start`) per the taxonomy — names + properties from the shared spec.
- [x] Attach the anonymous/pseudonymous install id + (when authenticated) PLG identity per the
      privacy model — **no PII / no target data** (respect existing `pii-check.yml`).
- [x] Honor a telemetry opt-out.
- [x] Wire the DSN secret + environment into `release.yml`.

**Repo:** Strike48-public/pick · **Depends on:** taxonomy/privacy spec (#72 child)

### LINKED ISSUE #279: Android app — single "check the network" button
Parent: Strike48/project-management#72 · Init: init/plg

## Android app — single "check the network" button

A dead-simple Android build of Pick: one button that runs the network check and shows results.
`android-lib/` scaffolding already exists in the repo; no Android build job in `release.yml` yet.

### Scope
- [ ] `dioxus-mobile` Android target building in CI (AAB), signed via the Play upload key.
- [x] Minimal easy-mode UI: one "Check the network" button → `network.check` flow → result view.
- [x] Wire Sentry events (per taxonomy).
- [ ] Play listing assets.

**Repo:** Strike48-public/pick · **Depends on:** Google Play account, Pick Sentry

### LINKED ISSUE #280: iOS app — single "check the network" button
Parent: Strike48/project-management#72 · Init: init/plg

## iOS app — single "check the network" button

Same one-button easy-mode experience as the Android app, on iOS (`dioxus-mobile`). Just a
"check the network" button — symmetric with Android.

### Scope
- [ ] `dioxus-mobile` iOS target building + signed/notarized in CI.
- [x] Minimal easy-mode UI: one "Check the network" button → `network.check` flow → result view.
- [x] Wire Sentry events (per taxonomy).
- [ ] TestFlight/dev-mode distribution (no App Store until later per punchlist).

**Repo:** Strike48-public/pick · **Depends on:** Apple Developer account, Pick Sentry

### LINKED ISSUE #283: Easy-mode default auth to PLG (no-auth fallback + override URL)
Parent: Strike48/project-management#72 · Init: init/plg

## Easy-mode default auth to PLG (no-auth fallback + override URL)

In easy mode, if there's no auth configured, default-authenticate to the PLG tenant; allow an
override URL for other environments. (Coordinate with StrikeHub's OIDC/Matrix auth.)

### Scope
- [x] Default to the PLG host/tenant when no auth is present.
- [x] Override URL escape hatch (advanced).
- [x] Graceful state when offline / auth unavailable.

**Repo:** Strike48-public/pick

### LINKED ISSUE #284: Easy-mode share link — share screenshot/usage to X / LinkedIn / Facebook
Parent: Strike48/project-management#72 · Init: init/plg

## Easy-mode share link → screenshot/usage share to X / LinkedIn / Facebook

An easy-mode button that produces a share link; the link opens a share to X / LinkedIn / Facebook
with the user's screenshot / usage summary — a growth loop.

### Scope
- [ ] Generate a shareable link/card (screenshot + usage summary from analytics events).
- [ ] Share intents for X, LinkedIn, Facebook.
- [ ] Track share events (per taxonomy) so influence/virality is measurable.

**Repo:** Strike48-public/pick · **Depends on:** Pick Sentry

josh-devo:
## What

Closes #279
Closes #278 
Closes #280 
Closes #283 
Closes #284 
Closes https://github.com/Strike48/project-management/issues/219

This branch started as the Android/iOS **buildability + CI** work and grew into the full **Easy Mode / PLG** product surface. The mobile surface is delivered via the **Dioxus** app (`apps/mobile`) on iOS and Android — the same Dioxus app is the desktop/advanced surface. (An earlier Crux native-shell exploration — SwiftUI + Jetpack Compose over a shared Rust Crux core — was reverted; it survives only as `docs/superpowers/` design notes.) It covers:

### 1. Mobile build + CI
- `just build-android` produces an installable APK on NixOS with no manual patching; dioxus/dx bumped to 0.7; Android CI re-enabled (nix-based, direct cache.nixos.org substitution).
- iOS + Android built from the single Dioxus `apps/mobile` crate (`just build-ios` / `just build-android`); no separate native shells.

### 2. Easy Mode (self-serve mobile UX)
- Reskin to the approved **Sage** design system (semantic `--em-*` CSS vars, IBM Plex, light/dark parity), co-brand top bar (Strike48 "S" badge + "Pick").
- Simplified chat: "Scan My Network" one-tap **plus a freeform message input** from the start screen; no expert chips; sage tool-call cards; shared Rust markdown renderer; agent activity status line.
- Navigation **drawer** (platform-native: iOS slide-over, Android `ModalNavigationDrawer`) opened by a left hamburger, holding New chat, Reports, Settings, recent chats, and Log out.
- Documents + Share: in-app markdown report viewer (own full-screen screen), public `/s/` share links, native OS share sheet + social intents (X/LinkedIn/Facebook); share/open uses the `?preview=1` URL so links render standalone.
- **Mermaid diagram** rendering (WKWebView / Android WebView) from the shared markdown parser.
- Settings screen with a runtime telemetry (usage-analytics) toggle; Log out clears token + session.
- iOS Keychain + Android Keystore secure token storage; auth restore on relaunch (drops expired tokens). iOS ad-hoc signed with a keychain-access-groups entitlement so the sim can persist.
- Rename display name to **Pick**.

### 3. PLG connector registration (OAuth-first)
- In easy mode, sign in via in-app OAuth, exchange the JWT for a tenant-scoped OTT (`/api/connectors/pre-approve`), stage it for the SDK, and register the connector **approved under the user's personal tenant** — replacing the tokenless->pending path.
- The Dioxus app registers a connector with **both `Tool` and `App` behaviors** (`pick_connector.rs`): `Tool` so the agent can execute device tools during a scan, and `App` so the connector serves its Dioxus UI (`AppManifest` + app pages) over the liveview socket. (The reverted crux shells were Tool-only — no App surface over the wire.)

### 4. Telemetry / observability (#278)
Sentry, wired through the shared **Rust core** (`crates/core/src/telemetry.rs`) so it covers all Dioxus surfaces (desktop + mobile):
- Usage traces + Rust-panic capture. Runtime opt-out that **fully closes the client** (no events *or* release-health sessions) and re-inits on re-enable. User-facing UI actions (`scan.send`, `message.send`, `reports.list`, `sign_in`, `document.open`, `share_link.create`, `conversation.load`, ...) are timed spans with `ok`/`error` outcome; `scan.start` is tagged `source=easy_mode_button`. Tool runs are timed spans that **continue a distributed trace** from `sentry_trace`/`baggage` headers if the backend forwards them. Agent backend errors are captured as **Issues** (`capture_agent_error`, classified `stream_error`/`token_limit`/`upstream`).
- **Design decisions:** activity is emitted as **traces, not Issues** (`capture_message` would file everything as an error); DSN/env injected at **build time** from the ambient env via `option_env!` (`PICK_SENTRY_DSN` + `STRIKE48_SENTRY_ENV`, defaulting dev builds to the `dev` environment) so nothing is committed to this public repo; everything gated on the runtime telemetry opt-out.
- *(The separate native-SDK view-crash layer — sentry-cocoa / sentry-android — was part of the reverted native-shell work and is not in this branch.)*

## Tickets

- Epic: **Strike48/project-management#72** — PLG refinement.
- **#283** (pick) — Easy-mode default PLG auth. done
- **#278** (pick) — Sentry telemetry. done (per-activity event *names* provisional, pending taxonomy spec)
- **#284** (pick) — Easy-mode share -> social. Social-intent + doc-share building blocks done; usage-summary card + share-event analytics remain gated on the taxonomy.
- **project-management#101** — usage-event taxonomy / PII model (external spec).

## Verified (this session)
- **iOS (Dioxus, `apps/mobile`):** Easy Mode + PLG functional on the simulator — sign in, connector registers under the personal tenant, scan runs, agent calls `document_write` (shareable report created), in-app viewer + share, drawer nav, settings/telemetry toggle. Rust Sentry init confirmed (`environment=dev`). (Keychain token persistence needs a real signing identity on the sim — tracked separately.)
- **Android (Dioxus, `apps/mobile`):** Easy Mode functional on the x86_64 emulator — same flow end-to-end; token persists across relaunch (Keystore); Rust Sentry init confirmed.
- **Desktop (Dioxus):** Easy Mode + change-server + report viewer/zoom verified on Linux against the plg dev backend.
- Batch host tools (`port_scan` subnet/hosts, `service_banner` targets) collapse the scan from ~22 serial tool calls to a handful.
- `cargo clippy -D warnings` clean across the workspace (core / platform / tools / ui); core + middleware unit tests pass.

## Known upstream (matrix) gaps
- PLG personal-tenant provisioning on the dev cluster does not yet set the tenant realm (defaults NULL -> connector realm resolves to a nonexistent slug), grant the owner an admin role, or run tenant policy provisioning. Worked around via RPC for testing; matrix-side fixes.
- **Public share links:** the deployed studio image serves `/s/:token` behind a login; the fix that serves `public`-scope links (and `?preview=1`) anonymously is the *next* commit on the matrix branch and not yet deployed — the connector side is wired and works once it ships.
- **Distributed trace nesting:** tool spans join the conversation trace once the backend forwards `sentry_trace` (+ `baggage`) in the tool-request `metadata` map (`tool_manager.ex`); our side reads it already. Until then, tool spans are standalone-but-timed.

## Follow-ups (not in this PR)
- **Durable auth:** the underlying Studio session token is short-lived and non-refreshable, so once it *expires* the user is sent back to sign-in. Token **persistence across relaunch works** (Keychain/Keystore, verified iOS + Android) — the remaining work is exchanging it for a longer-lived Keycloak JWT via `POST /api/app/token/exchange` to avoid the eventual re-login.
- **GraphQL `conversationEvents` subscription** to replace the 800ms poll (matches Dioxus).
macos-desktop-dioxus-05-settings
macos-desktop-dioxus-04-document
macos-desktop-dioxus-03-reports
macos-desktop-dioxus-02-drawer
macos-desktop-dioxus-01-home
android-dioxus-07-settings
android-dioxus-08-signin
ios-dioxus-07-settings
ios-dioxus-06-document
ios-dioxus-05-reports
ios-dioxus-04-drawer
ios-dioxus-03-home
ios-dioxus-01-signin
ios-dioxus-02-oauth-login
linux-desktop-dioxus-05-settings
linux-desktop-dioxus-04-document
linux-desktop-dioxus-03-reports
linux-desktop-dioxus-01-home
linux-desktop-dioxus-02-drawer
android-dioxus-05-reports
android-dioxus-06-document
android-dioxus-04-home-empty
android-dioxus-02-drawer
android-dioxus-01


jtomek-strike48 (changes requested):
## Review method

Checked out `355d9ca` into an isolated worktree, base `63d511f`, and ran the real gates there rather than reasoning about the diff. Everything below states what was run and what happened.

Gates I ran: `cargo test --workspace --locked --features pentest-platform/desktop-pcap` passes (exit 0, ~1000 tests, zero failures). `cargo clippy --workspace -- -D warnings` is clean. Branch is 0 commits behind `main`.

## Verified sound

Leading with this because most of the engineering here holds up under adversarial checking:

- **`insecure-tls` gating is genuinely careful.** A dedicated cargo feature, deliberately not `debug_assertions` (which is off for `release-ffi`), enabled only by the sim/emulator justfile recipes. The arm64 device recipe and every CI and release path omit it. `grep` confirms no committed build bakes it.
- **No memory-safety defect at the FFI boundary.** All null checks, `Box::into_raw`/`from_raw` pairing, `PickBuf` capacity round-trip, both shells' free discipline, and the `Send`/`Sync` justifications check out. `markdown.rs` is panic-free across 17 adversarial inputs. Absence of `catch_unwind` aborts rather than causing UB on rustc 1.94.1.
- **Telemetry has real PII discipline.** `send_default_pii: false`, pseudonymous install id, and every span property traces to a fixed literal. No credential or target data reaches Sentry. The Rust runtime toggle genuinely closes the client.
- **Mutation testing confirms the suite is load-bearing where it exists.** Neutering the implementation turned the guarding test RED for `markdown.rs`, `pre_approval.rs` (0600 and clear guarded independently), `util.rs::param_u16_opt`, the `port_scan` CIDR helpers, `realm_from_token`, and the `start_paused` OAuth race tests.
- `run_js`'s filesystem sandbox correctly reuses `workspace::resolve_path`, with real timeout and memory bounds. `secure_store` returns errors rather than falling back silently. The `docs/superpowers/` spec and plan docs follow the existing convention already on `main`.

## Blocking

**1. `apps/ios-crux/PickCrux/OAuthManager.swift:48` - full access token written to the iOS system log.**
`NSLog("[PickCrux] OAuth callback URL: \(callbackURL?.absoluteString ...))"` logs the whole callback URL, and `extractAccessToken` at `:66-83` reads the token from exactly that URL's query or fragment. The token lands in the unified log, readable via `log show` and captured in sysdiagnose bundles. Line 55 immediately below logs only `token.count`, so the length-only convention was known. Fix: log the length, or nothing.

**2. `apps/android-crux/.../AndroidManifest.xml:49` with `crates/core/src/matrix/auth.rs:335` - forged OAuth callback can hijack connector registration.**
The callback activity is `android:exported="true"` with category `BROWSABLE`. Grepping all five new or changed OAuth files (`auth.rs`, `ios/oauth.rs`, both `OAuthCallbackActivity.kt`, `OAuthManager.swift`) for `state`, `nonce`, `pkce`, `code_challenge` returns zero hits. `deliver_native_oauth_callback` caches the token before checking for a waiting login, and its `None` arm returns `true`, so a token is adopted with no sign-in in flight. Any zero-permission app can `startActivity` a forged callback and cause the victim's connector, which exposes `execute_command`, `run_js` and nmap, to register under the attacker's tenant. Fix: generate and validate a `state` parameter, and reject callbacks with no matching in-flight request.

**3. `crates/crux-core/src/markdown.rs:65` - markdown tables render as nothing.**
Verified by executing the real parser: `"| Host | Port |\n|---|---|\n| 10.0.0.1 | 22 |"` returns `[]`, zero blocks. `"- parent\n  - child"` returns a single `ListItem` containing only `"child"`; the parent text is gone. Line 65 sets `Options::ENABLE_TABLES`, so pulldown-cmark emits `Tag::Table`/`TableHead`/`TableRow`/`TableCell`, but no match arm handles any of them and the text falls through unmatched. A findings table is the most common shape an agent emits in a scan report, and this is the shared renderer behind iOS chat, Android chat, and the in-app report viewer, so the user sees a blank gap where results should be. None of the 12 markdown tests cover it.

**4. `crates/crux-core/src/update.rs:87` - a poll in flight across logout repopulates the cleared session.**
`Event::Logout` (`:483-497`) clears `conversation_id`, `messages`, docs and history, but a `PollConversation` issued beforehand resolves afterwards and the `Delta` arm assigns `model.messages = delta.messages` with no comparison against `model.conversation_id`. On a shared device the previous user's scan findings render into the next user's session. Compounding it, `:154` does `conversation_id.clone().unwrap_or_default()`, so once `conversation_id` is `None` the poll re-arms with the empty string every 800ms indefinitely. Fix: tag deltas with their conversation id and drop non-matching ones; stop polling when `conversation_id` is `None`.

**5. `crates/crux-ffi/src/lib.rs:427` - iOS logout cannot clear the token.**
`ContentView.swift:148` `logout()` calls `core.setToken("")`; `CoreBridge.swift:80-83` does `Array(token.utf8).withUnsafeBufferPointer`, and an empty Swift array yields a nil `baseAddress`, so `pick_set_token` receives null and returns early at the `is_null()` guard. The core keeps the old token while the UI shows the sign-in screen. `pick_core_new`'s reader at `:358-361` explicitly handles this case ("An empty (zero-length) input is legitimate; a null pointer is not"), which is what makes this the bug rather than the convention. The Rust early return and the Swift call shape are confirmed by reading; the nil-`baseAddress` step is documented Swift behaviour, not executed here. Fix: accept null-with-zero-length as the empty string, matching `pick_core_new`.

**6. `crates/crux-ffi/src/lib.rs:212` - the tool connector survives logout under the previous user's tenant.**
`connector_started.swap(true, SeqCst)` latches for the life of the process; the only reset is inside the detached task's async-prep failure arm. There is no `pick_stop_connector` export and `Event::Logout` has no FFI counterpart. After user A signs out and B signs in on the same device, A's connector stays live on A's tenant and keeps accepting tool-execution requests, while B's token adoption silently declines to start one. Fix: add an explicit stop path that aborts the task and clears the flag on logout.

**7. `crates/crux-core/src/update.rs:82` - failure leaves an animated status line running forever.**
`Event::StartScan` sets `model.activity = AgentActivity::Thinking` (`:20`). The scan-failure branch (`:82-84`) and the sign-in failure branch (`:164-166`) set `scan_active = false` and `error`, but never reset `activity`. `ChatList.swift:30` renders `TypingIndicator` whenever `agentActivity != .idle`, so the user sees a permanent "Thinking..." animation beside the error card with no escape. The sibling notice branch at `:103-107` does clear it, which shows this is an omission. All 30 crux-core tests pass with the bug present.

**8. `crates/tools/src/run_js.rs:314` - `host.httpFetch` has no URL validation.**
`grep -c validate_url` in `run_js.rs` returns 0, while seven sibling tools (nikto, zap, ffuf, dirb, gobuster, metasploit, ad/common) call `pentest_core::url_validation::validate_url`. An agent-authored payload calling `host.httpFetch('http://169.254.169.254/latest/meta-data/iam/security-credentials/')` reaches the metadata service and returns the body into JS, which can exfiltrate it through a second `httpFetch`. `RunJsTool` is registered unconditionally for all platforms (`lib.rs:247`). The filesystem side is properly guarded, so the module doc's "same guards the native tools use" claim at `:14` is false for the network path.

**9. Required `Format` check fails.**
Reproduced locally: `cargo fmt --all -- --check` exits 1 with three diffs, all in `crates/platform/src/ios/oauth.rs` (lines 77, 113, 134), matching the CI log. `Check` and `Clippy` pass because they run on ubuntu and macos and never compile the `cfg(target_os = "ios")` module, while `cargo fmt` parses every file target-agnostically. One `cargo fmt --all` fixes it.

## Should-fix

**Credential and session handling**
- `crates/core/src/matrix/pre_approval.rs:65` - the OTT is written with default permissions and only chmod'ed to 0600 afterwards. Verified by executing the same sequence under umask 002: the file lands at 0664 inside a 0775 directory before the chmod. The existing test asserts only the final mode. Use `OpenOptions::new().mode(0o600)` and set the directory to 0700.
- `crates/crux-ffi/src/lib.rs:242` - the crux mobile path stages the single-use OTT but never calls `clear_staged_ott`, on success or failure. Calls exist only in `crates/ui` (the Dioxus front-end), yet `pre_approval.rs:77-78` documents that it is cleared on both paths.
- `crates/core/src/connector_registration.rs:23` - `derive_api_url("https://host", use_tls=false)` returns `http://host`, verified by reimplementing the function. That URL then carries the user's bearer JWT in `pre_approve`. An explicit `https` input should not be downgraded.
- `crates/crux-core/src/update.rs:483` - `Event::Logout` clears eleven fields but not `tool_calls`, so the prior session's nmap invocations and target hosts stay on screen. `Event::NewChat` has the same gap.

**Silent failures**
- `crates/platform/src/common/mdns.rs:45` - the result of `set_read_timeout` is discarded. The loop at `:57-67` relies on each `recv_from` returning within 250ms to re-check the deadline; if the setsockopt failed the socket stays blocking and `discover()` can hang indefinitely, with no enclosing timeout.
- `crates/platform/src/common/ssdp.rs:25,41` and `mdns.rs:38,49` - bind and send failures return `Ok(vec![])`. The doc comment says this is deliberate, but for a pentest report "the probe never ran" and "nothing is on the network" must be distinguishable. Suggest a structured result recording the probe as skipped.
- `crates/core/src/telemetry.rs:203` - `sentry::init`'s guard is never checked with `is_enabled()`, yet `ENABLED` is set true and "telemetry initialized" is logged. A rejected DSN yields a disabled client rather than an error, so the UI claims telemetry is on while nothing is delivered. `flush()` is documented as required on backgrounding and has no caller.
- `crates/ui/src/components/documents_panel.rs:195` - `if let Ok(list) = ...` with no else arm drops list errors and retries forever, while the sibling at `:91-99` surfaces the same failure as "Couldn't load reports".

**LLM-facing input shapes**
- `crates/tools/src/util.rs:57` - `param_bool` accepts only real JSON booleans. Verified by execution: `param_bool({"append":"true"}, false)` returns `false` and `param_bool({"recursive":1}, false)` returns `false`, while `param_u64({"timeout_ms":"5000"})` returns 5000. Lines 44-47 document precisely this hazard for numbers but `param_bool` did not get the same treatment, across 20+ call sites. It also has no test.

**Test coverage**
- `crates/crux-ffi/src/lib.rs:575` - `fake_core()` calls the private `PickCore::build` and does its own `Box::into_raw`, bypassing `pick_core_new`. Making that constructor unconditionally return null leaves all 4 tests green, as does no-oping `pick_set_notify`. Five of seven exported symbols are exercised; the untested two include the first call every shell makes.
- `crates/crux-core/src/update.rs:91` - the replace-not-extend fix carries a comment describing the duplicate-render bug it fixed, but the only delta test starts from `..Default::default()`, so `messages` is empty and extend equals replace. Reverting to `.extend()` keeps all 30 tests green. Seed the model with an existing message.
- `crates/core/src/matrix/auth.rs:201` - the eight OAuth tests are behind the non-default `browser-auth` feature. The macOS lane's package-scoped command never enables it, so they compile to zero tests there and run on Linux only via incidental feature unification. Add the feature to the invocation.
- `crates/tools/src/service_banner.rs` - zero tests for +154 lines of new batch logic, including `as_str().unwrap_or("")` on a missing host. Its sibling `port_scan.rs` does have `collect_hosts` tests.
- `crates/tools/src/run_js.rs:440` - the traversal test hardcodes a two-level `../..` escape, so its strength depends on `TMPDIR` depth; with the guard removed it leaked `/etc/passwd` under `TMPDIR=/tmp`. `fs_write` and `fs_list` have no traversal test. Assert on an absolute path.
- No tests for roughly 6,000 lines of new Swift and Kotlin. No test directory, no XCTest target, no `testImplementation`; the three new mobile CI jobs are compile gates only. Every behaviour listed under "Verified (this session)" rests on manual testing and will regress silently.

**Maintainability and disclosure**
- Generated bindings are committed four times in cross-language form: 11 Swift files plus a stray `build.gradle.kts` under `apps/android-crux/.../java/`, and 17 Kotlin files under `apps/ios-crux/generated/`. The cross-language copies are stale, differing by 188 Swift and 212 Kotlin lines including shifted bincode variant tags (`ConnectResult` at 13 versus 16). Not a live wire break: I traced `settings.gradle.kts` sourceSets and the pbxproj, and each build consumes the correct fresh copy of its own language. But four copies of a positional wire contract with no CI regeneration check will drift. Delete the cross-language copies and add a regenerate-and-diff step.
- `run_js` is absent from the PR description. A 482-line arbitrary-JavaScript execution tool with outbound HTTP and filesystem access, registered unconditionally into the live registry, is the highest-risk addition in the diff and where finding 8 lives. A reviewer following the body's four sections would never look at it. Please add a section covering the tool, its sandbox bounds, and its threat model.

## Low

- Telemetry is enabled by default with no first-run consent prompt, and a mid-session opt-out only reaches the native SDKs on next launch. The implementation is careful and leaks nothing, so this is a product decision rather than a defect, but a pentest connector phoning home on first launch deserves an explicit maintainer call.
- `crates/crux-core/tests/ffi_bridge.rs:19` asserts only that byte buffers are non-empty; gutting the entire ViewModel projection still passes. Decode and assert on concrete fields.

## Suggested split

The engineering quality here is high and most of my sound-verification list is substantial. My concern is the review surface, not the work: 247 files and 224 commits spanning build tooling, a new shared-core architecture, two native shells, an auth flow, telemetry, and an undisclosed code-execution tool is more than one unit can be reviewed as, and the findings above cluster in the parts that got the least attention per line.

Suggest landing in this order, each with its own CI signal:

1. Mobile build and CI (nix flake, dx 0.7.9, Android and iOS crux build jobs). Mechanical, independently verifiable, and unblocks the rest.
2. Crux core plus shells, with the markdown table fix and the logout and session-lifecycle fixes (findings 3, 4, 5, 6, 7).
3. PLG OAuth-first registration, with the credential fixes (findings 1, 2, and the OTT items).
4. Telemetry.
5. `run_js` on its own, with the URL validation fix and a written threat model.

Happy to re-review any of these individually. Blocking items are 1 through 9; item 9 is a single `cargo fmt --all`.

jtomek-strike48:
Heads-up on one item from my review that has since gotten worse, so the fix is slightly larger than I described.

**Blocking item 9 (`Format` check)** — when I reviewed, `cargo fmt --all -- --check` produced 3 diffs, all confined to `crates/platform/src/ios/oauth.rs`. The current run ([30146740179](https://github.com/Strike48-public/pick/actions/runs/30146740179), 2026-07-25 06:00) has **20 diffs across 5 files**:

```
crates/platform/src/ios/oauth.rs
crates/ui/src/auth_flow.rs
crates/ui/src/components/chat_panel/mod.rs
crates/ui/src/connector_app.rs
crates/ui/src/lib.rs
```

Still one `cargo fmt --all` to fix. Worth noting *why* it keeps slipping through: `Check` and `Clippy` stay green because they run on ubuntu and macos and never compile the `cfg(target_os = "ios")` module, while `cargo fmt` parses every file target-agnostically. The `crates/ui` diffs are new since my review, so this is drifting as the branch grows rather than being a one-time miss — a `cargo fmt --all` in your pre-commit hook would stop it recurring.

The other 8 blocking findings from my review stand as written; nothing I've since disproved. No need to re-request review until you want another pass.

jtomek-strike48:
## Re-review - pick#249 (Easy Mode + PLG mobile), current head `ba8bb7f`

**Method:** checked out the PR head into an isolated worktree, base `84d0b8e8` (`main`). This is a re-review against my `CHANGES_REQUESTED` from 2026-07-25 (at `355d9ca`, still in history - no force-push, +89 commits since). Two specialist passes ran the empirical gates: Rust core tests **343/343 pass** (build verified), and a dedicated security pass verified the OTT/tenant/token-storage/telemetry surface. I then re-checked each prior BLOCKING finding against the current head myself. States run vs. read per item.

### Verdict: still **Changes requested.** Real, creditable progress on the PLG/telemetry surface - but the majority of the prior blockers are still open on the current head, it's a draft, and it's conflicting with `main`. Can't greenlight yet; here's the exact state so it's a short list to close.

### Genuinely resolved since the last review - nice work
- **OTT / tenant registration surface is sound** (security pass, verified): `tenant_id` is server-scoped (a user can't inject another tenant), the OTT is cleared on success/error/sign-out, token storage is Keychain/Keystore (never plaintext), expired tokens are dropped on restore. No secrets in the diff; no DSN committed.
- **Telemetry is correct** (verified): runtime opt-out fully drops the client and stops all egress; `send_default_pii=false`; span properties trace to fixed literals - no target/credential data reaches Sentry. The "fully closes the client and re-inits" claim holds.
- **Polling->subscription migration** (`a596953` and the `conversationEvents` work) removes the old 800ms poll path and reshapes the logout/delta handling that finding #4 was about.
- Core + middleware tests pass (343/343, verified).

### Still BLOCKING on the current head (verified this pass)
1. **iOS OAuth token logged to syslog - NOT fixed.** `apps/ios-crux/PickCrux/OAuthManager.swift:48` still `NSLog`s the full callback URL (`callbackURL?.absoluteString`), and `:50` extracts the token from that same URL. Line `:56` right below logs only `token.count`, so the safe pattern is right there. The token lands in the unified log (`log show`, sysdiagnose). Fix: log length or nothing. (read)
2. **Forged OAuth callback can hijack connector registration - NOT fixed.** Grep across all OAuth files (`auth.rs`, `ios/oauth.rs`, both `OAuthCallbackActivity.kt`, `OAuthManager.swift`) for `state`/`nonce`/`pkce`/`code_challenge`/`code_verifier` still returns zero real CSRF tokens. `deliver_native_oauth_callback` (`crates/core/src/matrix/auth.rs:317`) caches the token from the unvalidated callback URL **first**, before any in-flight-login check, and the Android callback activity is `exported=true` + `BROWSABLE`. A zero-permission app can still `startActivity` a forged callback and register the victim's connector (which exposes `execute_command`/`run_js`/nmap) under an attacker-influenced flow. Fix: generate + validate a `state` param; reject callbacks with no matching in-flight request. (read/grep)
3. **Markdown tables render as nothing - NOT fixed.** `crates/crux-core/src/markdown.rs:65` sets `ENABLE_TABLES`, but there is still **no `Tag::Table`/`TableHead`/`TableRow`/`TableCell` match arm** - table events fall through unmatched and produce zero blocks. This is the shared renderer behind iOS chat, Android chat, and the in-app report viewer, and a findings table is the most common shape a scan report emits. (read)
4. **SSRF via `host.httpFetch` - NOT fixed.** `crates/tools/src/run_js.rs` still has `validate_url` count = 0, while seven sibling tools call `pentest_core::url_validation::validate_url`. An agent-authored payload can `host.httpFetch('http://169.254.169.254/latest/meta-data/...')` and exfiltrate the response. `RunJsTool` is registered unconditionally. (read/grep)
5. **iOS logout can't clear the token - NOT fixed.** `ContentView.swift:148` `logout()` calls `core.setToken("")`; `pick_set_token` (`crates/crux-ffi/src/lib.rs:427`) still early-returns on a null pointer with no null-with-zero-length special case, so the empty Swift array (nil `baseAddress`) is dropped and the core keeps the old token. `pick_core_new` explicitly handles this case, which is what makes it a bug not a convention. (read)
6. **Connector survives logout under the previous user's tenant - NOT fixed.** There is still **no `pick_stop_connector` export**; `connector_started` only resets inside the async-prep failure arm, and `Event::Logout` doesn't touch it. On a shared device, user A's connector stays live on A's tenant after B signs in. Fix: add an explicit stop path that aborts the task and clears the flag on logout. (read)
7. **Failure leaves the "Thinking..." animation running forever - NOT fixed.** `crates/crux-core/src/update.rs` scan-failure (`:82`) and sign-in-failure (`:164`) arms set `scan_active = false` but never reset `model.activity`; only the notice arm (`:106`) does. The typing indicator renders whenever `activity != Idle`, so the user sees a permanent spinner next to the error. (read)
8. **Required `Format` check still fails.** `cargo fmt --all -- --check` exits non-zero at head - the prior `ios/oauth.rs` diffs were fixed, but new unformatted code was added (`crates/core/src/matrix/phoenix_sub.rs:271,297,465,505,519,540`, `report_meta.rs:80,87,103,181`). One `cargo fmt --all` clears it. (ran)

### Still worth closing (should-fix, from the prior list - spot-checked, still applicable)
- `pre_approval.rs` OTT file created then chmod'd 0600 (race window on Unix; Windows unprotected) - the security pass flagged the Windows side as LOW given single-use + short-lived + sandboxed, but the create-then-chmod race is cheap to close with `OpenOptions.mode(0o600)`.
- `derive_api_url` still downgrades an explicit `https` input when `use_tls=false`, and (Rust pass) has no tests for the `grpc://`/`grpcs://` scheme branches it advertises.
- `Event::Logout` clears many fields but not `tool_calls`, so the prior session's nmap invocations/targets stay on screen.
- OAuth tests behind the non-default `browser-auth` feature don't run on the macOS package-scoped lane (Lens 2).
- ~6,000 lines of new Swift/Kotlin have no tests; the mobile CI jobs are compile gates only.
- `run_js` (482-line arbitrary-JS tool with outbound HTTP + FS) is still absent from the PR description - please add a section covering its sandbox bounds + threat model (finding #4 lives there).

### Process
- **Draft + conflicting with `main`.** GitHub won't let this be approved/merged while it's a draft and `DIRTY`. Rebase onto `main`, close the blockers above, un-draft, and re-request. Given the security-relevant blockers (#1, #2, #4), I'd want a fresh security pass on the OAuth-callback + `run_js` paths specifically before an approve.

Happy to re-run the empirical gates as soon as these are pushed - the fastest path to green is #8 (`cargo fmt`) + #1/#3/#5/#7 (small, localized) first, then the callback CSRF (#2) and SSRF (#4) which need the most care.

jtomek-strike48:
Filed the security items from the re-review as tracked issues so they're not lost when this draft moves:

- #339 - OAuth callback CSRF (no `state`/PKCE); forged callback can hijack connector registration [P0] (was BLOCKING #2)
- #340 - `run_js` `host.httpFetch` SSRF, no URL validation [P0] (was BLOCKING #8)
- #341 - iOS OAuth token written to the unified log via `NSLog` [P1] (was BLOCKING #1)
- #342 - logout leaves the in-core token set and the tool connector live on the prior user's tenant [P1] (was BLOCKING #5 + #6)

Each carries the file:line evidence and acceptance criteria. The remaining blockers from the review (markdown tables render blank #3, stuck "Thinking..." on failure #7, `cargo fmt` #8) are UX/hygiene and stay in the PR thread rather than as separate security issues.

jtomek-strike48 (approved):
## Hermes Review

**Verdict: LGTM**

Huge PR (242 files, +41k/−3k) delivering Easy Mode/PLG mobile, OAuth-first connector registration, secure token storage, telemetry, and mobile build/CI. Triage focus was the security-sensitive surfaces (OAuth, token storage, pre-approval OTT, the new `http_request` tool, telemetry DSN handling). Overall the security posture is deliberate and well-tested — secrets baked at build time, OTT written 0600, bearer never downgraded to http, Keychain/Keystore for tokens, SSRF gate on the HTTP tool. Nothing blocking.

### Warnings
- **crates/core/src/matrix/auth.rs (native OAuth callback)** — The iOS path accepts a callback token when `state` is *absent* (server didn't echo it), relying on "a login is in flight + the token arrived on our custom scheme" as the binding. That's a documented, reasonable relaxation, but custom-scheme registration isn't exclusive on iOS. Confirm the auth session is `ASWebAuthenticationSession` (which binds the callback to the initiating app) on every native path that reaches this code — if any path uses a plain scheme handler, the missing-state window is exploitable. When `state` *is* present it's validated correctly and forged/no-login-in-flight callbacks are rejected — good.

### Suggestions
- **crates/tools/src/http_request.rs:~118** — `danger_accept_invalid_certs(true)` is hardcoded on. Documented as intentional for probing lab/self-signed targets and gated by `validate_url(target_validation_mode())`, so acceptable. Consider making it a `verify_tls` opt-out param (default off) so the agent can still make trust-sensitive requests when it wants them — minor, not blocking.
- **crates/tools/src/port_scan.rs:~216** — `results.into_iter().next().unwrap()` in the single-target branch is safe today (one target ⇒ one result) but fragile if the collection logic changes. A guarded match or `expect("single-target scan always yields one result")` documents the invariant.

### Looks Good
- `pre_approval.rs` / `connector_registration.rs`: OTT written at 0600 (created *at* mode, no permissive window), single-use with clear-on-success/failure, `derive_api_url` never downgrades an explicit https/wss to http — all with regression tests including the `matrix_wss_url`-absent studio.strike48.com case.
- `secure_store.rs` + `ios/keychain.rs`: clean backend abstraction; Keychain items use `AccessibleAfterFirstUnlockThisDeviceOnly`, never synced/iCloud, delete-then-add upsert. Desktop falls back to settings file only when no backend registered — correctly documented as single-user-desktop-only.
- `telemetry.rs`: DSN via `option_env!("PICK_SENTRY_DSN")` (build-time only, deliberately not the SDK's magic `SENTRY_DSN`), env kill-switch, runtime opt-out fully closes the client, production is opt-in only, traces sample rate clamped 0..=1. Privacy model (no PII/targets/args) stated and enforced via enum/count properties.
- `social_share.rs`: query params encoded via `reqwest::Url`, no hand-rolled escaping.
- Strong unit-test coverage throughout; PR reports `cargo clippy -D warnings` clean and core/middleware tests passing.

jtomek-strike48 (changes requested):
## Review: verified empirically against head 7bf848f (worktree + real gates), not diff-reasoning

Big, careful piece of work, and a lot of it holds up under direct testing. Flagging one merge-gating regression and the red Android CI first, then the batch.

### Verified sound (checked directly, so these are not concerns)
- Auth/token: token never logged; loopback binds 127.0.0.1 only; native (iOS/Android) OAuth uses a 128-bit CSPRNG `state` verified on callback; secure store is fail-closed with read-back-after-write; logout clears both persisted and in-memory; OTT file created 0600; the new auth guards are mutation-tested and genuine.
- Telemetry: no target IPs, tokens, or tool arguments reach Sentry (traced every sink); no DSN committed (`option_env!`); Gitleaks/PII/Dependency-Audit green.
- Streaming: cumulative-vs-delta handling correct; no conversation-id TOCTOU; reconnect/stall-watchdog resilient.
- AuthFlow reducer is a total function with no reachable dead-end, no double-fire, no reentrancy. serde defaults complete. Headless is unaffected (the +3 is a `telemetry::flush()` no-op). config (48) and auth_flow (11) unit tests run in the required `Test` job and pass (ran them at Rust 1.92 with the CI feature set).

### BLOCKING - merging as-is reverts the markdown XSS fix from #365/#367
This branch predates `c161dc2` (#365/#367), and that commit is the markdown sanitizer. It touches the exact two files this PR rewrites, which is why the PR is DIRTY:
- `crates/core/src/rendering.rs` here is 341 lines with no `MarkdownSanitizer`/scheme-allowlist; on `main` it is 593 lines with it.
- `crates/ui/src/components/chat_panel/render.rs:231 render_markdown` -> `push_html` -> `dangerous_inner_html` (lines 52, 78), unsanitized. That is the highest-risk sink #367 closed (chat reflects agent/tool/target output).

A "take the branch's version" conflict resolution silently reintroduces script/onerror/javascript: injection into the desktop webview. Fix: rebase onto current `main`, keep `MarkdownSanitizer`, and route the new chart-fence rendering through `sanitize_markdown_event` so the two renderers cannot drift.

### BLOCKING for the "Android CI re-enabled" claim - Android Build is red
The Android Build job is new in this PR (`ci.yml:213`). `just fetch-proot` pins `proot_version := "5.1.107.89"` and `curl -sL` (no `--fail`) it from Termux `pool/main`, which only keeps the latest build. The version rotated, the URL 404s, the HTML error page is saved as the .deb, and `ar x` fails: `ar: ../proot_x86_64.deb: file format not recognized`. Fix: `curl -fsSL`, and vendor/checksum the .deb or resolve the version dynamically.

### SHOULD-FIX
- Supply-chain integrity: `rootfs.rs:726`, `docker.rs:42`, `arch.rs:36` download rootfs/packages with no checksum; the ALARM aarch64 rootfs is fetched over HTTP; docker `strap.sh` curl is missing `--fail`. Add a sha256 param to `download_file`, self-host/HTTPS the rootfs, and use `curl -fsSLf`.
- Telemetry opt-out does not persist in easy mode: `easy_mode.rs:422` writes the toggle via a detached `load_settings()->save`, so a later signal save reverts `telemetry_enabled` to true next launch (the expert toggle at `connector_app.rs:1522` does it correctly through the signal). With telemetry default-on (`config.rs:860`), the user's opt-out silently does not stick.
- Settings clobber wipes the saved endpoint/tenant: `connector_app.rs:685` (remember) and `:772` (OTT self-heal) persist via `settings.peek().clone()` without updating the signal, so a later `settings.write()` save clobbers `last_config`/`auto_connect`/`tenant_id`. Regresses the pick#223 tenant-UUID persistence. Same class fixed for `easy_mode`/`matrix_api_url`; finish it for these fields.
- OAuth callback page XSS: `auth.rs:853` interpolates `matrix_url` into `var MATRIX_URL = '{matrix_url}';` unescaped; a `'` breaks out, and that page handles the access token. `serde_json`-encode it.
- Desktop loopback OAuth omits CSRF `state`: the loopback authorize URL (`auth.rs:~1053`) has no `state`, and `/callback`+`/token` do not verify one, while both native flows do (and the comment at 469-475 calls the loopback nonce more load-bearing). A cross-site page or local process can hit `http://localhost:PORT/token?access_token=...` to fixate a token. Round-trip and verify `state` on the desktop path too.

### MEDIUM / LOW
- Frontmatter YAML: `report_meta.rs:64` calls `serde_yml::from_str` (0.0.12) with no size/complexity cap on agent-authored frontmatter (operator-local DoS on a crafted report). Also, malformed frontmatter degrades to `Some(Default)`, which `badge()` renders as a false "clean" badge. Cap input size and treat parse failure distinctly from clean.
- Bearer token in URL query: `phoenix_sub.rs:195` puts the token in the WS join URL (`?token=`), and the loopback uses `/token?access_token=`. URLs land in proxy/server logs (RFC 6750 sec 2.3). Prefer a header/subprotocol for the WS token.
- Unbounded event channel `subscription.rs:129` (add backpressure); `unwrap_or_default()` swallows frame-serialize errors `subscription.rs:383`; sandbox integration tests are all `#[ignore]d` (confirm a periodic job runs them).

### Checked and NOT issues (so they do not get re-raised)
- `wsl.rs:509` is not a command injection: `wsl.exe` is invoked via an argv array (not a shell), and `-- /bin/bash -c <cmd>` runs inside the sandbox by design; the pacman-lock-heal prepend grants nothing new.
- Telemetry activity is emitted as traces, not Issues, as documented; only agent errors use `capture_message`, which is intended.

### Scope note
242 files / +41k spanning Android CI, iOS build, Easy Mode, PLG OAuth, telemetry, a WSL installer, streaming, and vendoring dioxus-liveview. Hard to review as one unit; a rebase onto current `main` (preserving #367) is a prerequisite regardless.

jtomek-strike48:
**Bug: in Easy Mode the connector never selects an agent, so "Scan My Network" is a no-op and the chat shows "Select an agent to begin".**

Reproduced against a dev tenant that has more than 10 agents.

Root-cause chain:
- `ChatPanel` (`crates/ui/src/components/chat_panel/mod.rs`) auto-selects the connector's persona by exact-name match against the result of `list_agents()`.
- `LIST_AGENTS_QUERY` (`crates/core/src/matrix/client.rs:428`) is `agents(filter: { isEnabled: true })` with no pagination and no name/connector scoping. The server returns only ~10 agents (alphabetically ordered), so the connector's own `pentest-connector` / `pentest-connector-validator` / `pentest-connector-report` personas (which sort later) are truncated off the list.
- The exact-name auto-select then misses, so the code falls through to `create_agent(...)`, which fails with "A persona with that name already exists" (they do exist, just off the returned page).
- The create-failure arm (~L771) only logs a warning and never sets `selected_agent`, leaving it `None`. That single `None` drives both the dead scan button and the empty-state text.

Evidence (instrumented build):
- `ListAgents` returned exactly 10 agents, none matching the connector.
- All three `CreateAgent` calls returned the "already exists" error.
- `selected_agent` was never set, yet `chat_ready` fired.
- `agents` rejects a `first:` argument ("Unknown argument \"first\" on field \"agents\""), and `filter` only exposes `isEnabled`, so there is no client-side way to page past the cap or fetch the connector's agent by name.

Suggested fixes (server-side needed):
- Make `createAgent` idempotent (return the existing agent on a name conflict). Smallest change; the connector code then works unmodified.
- Or add a scoped/paginated agents query (support `first:`, or a `filter` by name / connector key) and have the connector fetch its own persona by name.

Connector-side safety net (depends on the above): on a create "already exists" error, look up and select the existing agent instead of leaving `selected_agent = None`.

jtomek-strike48:
## Re-review - pick#249, head `2e6a90e` vs base `main` (`c18c8a7`)

**Method:** checked out the PR head into an isolated worktree over `main`; ran the Format gate, pulled the CI job logs, probed the Termux package pool, and read the changed files at head. States run-vs-read per item below. This supersedes my 2026-07-29 `CHANGES_REQUESTED` (at `ba8bb7f`), which was written against the **Crux** architecture that has since been reverted.

### Headline: much healthier than last pass. The security blockers are closed and the proot lane is fixed. The two things left are not code defects - the PR description no longer matches the code, and the diff is too large to review as one unit.

---

### Verified sound (nice work)

- **OAuth callback CSRF - fixed** (was BLOCKING #2 / issue #339). `crates/core/src/matrix/auth.rs` `deliver_native_oauth_callback` now rejects when no login is in flight, validates `state` against `NATIVE_OAUTH_STATE`, and caches the token **only after** validation - the exact ordering the prior review flagged. Guard tests exist (`deliver_rejects_callback_with_no_inflight_login`, `deliver_rejects_state_mismatch`). The documented relaxation (accept when the server omits `state`, on in-flight + custom-scheme binding) is a reasonable, commented trade-off. *(read + grep)*
- **run_js SSRF - surface removed** (was BLOCKING #4 / issue #340). `run_js.rs` is gone; the surviving fetch tool `crates/tools/src/http_request.rs:102` routes through `validate_url(&url, target_validation_mode(), None)?`. *(read + grep)*
- **Format gate - passes.** `cargo fmt --all -- --check` exits 0 locally and in CI (was BLOCKING #8). *(ran)*
- **Check / Clippy / Format green** on the current head's CI run; Test and Android Build still in progress at time of writing.

### Tracked security issues #339-342 - reconciliation

These were filed against the reverted Crux code, so their file:line evidence is stale. Re-verified against the current locations:

- **#339 (P0, OAuth CSRF)** - resolved (see above; guarded with tests).
- **#340 (P0, run_js SSRF)** - resolved; the vulnerable tool was removed and the replacement validates URLs.
- **#341 (P1, iOS token in unified log)** - the Swift `NSLog` is gone. iOS OAuth moved to `crates/platform/src/ios/oauth.rs`, which parses the callback `absoluteString` (`:196`) but does not log the callback URL or the token. Verified by read only - no on-device `log show` check.
- **#342 (P1, logout credential bleed)** - `clear_matrix_token()` clears secure-store + in-memory token (`crates/ui/src/session.rs:184-188`), and `crates/ui/src/connector_app.rs:1083` has an explicit easy-mode logout handler that shuts the connector down. Verified by read only - not exercised on a shared-device runtime.

Recommend wiring `Closes #339 #340 #341 #342` into this PR so they auto-close on merge, and a short device/runtime pass on #341/#342 before then since they are only read-verified.

### Android Build - fixed

The earlier failure (`ar: proot_x86_64.deb: file format not recognized`) was a pinned proot version dropped from the rolling Termux pool. `aee06ca` bumped it to 5.1.107.91 and added a fail-loud path; all four pinned Termux packages (proot 5.1.107.91, libtalloc 2.4.3, busybox 1.38.0-1, libandroid-shmem 0.7) now match the live pool, so the fetch step passes. *(verified against the pool)* Worth noting the fragility is structural - pinning against `termux-main/pool/main` (a rolling repo that keeps only current versions) will break again at the next Termux bump. A pinned/vendored mirror (as we already do for the openat2 proot release) would end the recurrence.

---

### Blocking-for-merge, but not code

**1. The PR description no longer matches the code.** The body describes the mobile surface as "native crux shells - a SwiftUI (iOS) and Jetpack Compose (Android) app over a shared Rust Crux core (`crates/crux-*`, `apps/ios-crux`, `apps/android-crux`)" with "Mermaid via WKWebView", etc. None of that exists at head - Crux survives only under `docs/superpowers/`. The mobile surface is **Dioxus** (`apps/mobile`), and the markdown renderer is `crates/core/src/rendering.rs` / `crates/ui/.../render.rs`. Anyone reviewing or merging gets the wrong mental model. A paste-ready corrected opener:

> This branch started as Android/iOS buildability + CI and grew into the Easy Mode / PLG product surface. The mobile surface is delivered via the **Dioxus** app (`apps/mobile`) on iOS and Android - the earlier Crux native-shell exploration was reverted (it survives only as `docs/superpowers/` design notes). Covers: mobile build + CI (`just build-android`, Android CI re-enabled), Easy Mode UX (Sage reskin, one-tap "Scan My Network" + freeform input, drawer nav, in-app report viewer + share), OAuth-first PLG connector registration, and Sentry telemetry with a runtime opt-out.

The "Verified" section's "iOS crux" / "Android crux" bullets should be reworded to the Dioxus surfaces (the screenshots are already mostly `*-dioxus-*`; a few stale `android-crux-*` shots remain).

**2. Scope: unreviewable as one unit.** 40,715 additions / 2,342 deletions / 242 files / 100 commits, spanning mobile CI, arch-aware sandbox, telemetry, a matrix auth/subscription rewrite, and docs. Because of the Crux -> Dioxus pivot, my prior two review rounds are superseded and do not transfer. Strongly recommend either splitting (mobile-CI, sandbox, telemetry, auth as separate PRs) or scheduling a fresh full review of the current shape before merge.

### Remaining verification gaps (honesty)

- Mobile behavioral coverage: `apps/mobile` (Dioxus) + `android-lib` (Kotlin) CI jobs are compile/build gates only - no behavioral tests confirmed.
- I relied on CI for Check/Clippy/Test/macOS/iOS rather than re-running the multi-minute builds locally; I ran only the Format gate locally.
- #341/#342 and the logout/connector-teardown paths were verified by reading the code, not by exercising them on a device.
- The 2026-08-14 "Scan My Network is a no-op / Select an agent to begin" report is a server-side agents-pagination cap (`LIST_AGENTS_QUERY`, no `first:`/scoping); the real fix is matrix-side and outside this PR.

jtomek-strike48 (approved):
Approving for the desktop surface (Linux/macOS/Windows).

This supersedes my prior `CHANGES_REQUESTED`, which was written against the reverted Crux architecture. Re-verified at head `2e6a90e`:

- Security blockers resolved: OAuth callback CSRF now validates `state` and caches the token only after validation, with guard tests (#339); the `run_js` SSRF surface was removed and `http_request` routes through `validate_url` (#340); iOS token is no longer logged (#341); logout clears token + credentials + staged OTT and shuts the connector down (#342).
- The two should-fixes I re-checked this pass are clear: `derive_api_url` no longer downgrades an explicit secure scheme and now has a regression test; the stale-`tool_calls`-across-logout concern does not apply in the Dioxus app (chat state is component-local and unmounts on logout).
- CI is green on all lanes.

Mobile (iOS/Android) is explicitly deferred and tracked in #387 - it is feature-gated, does not affect desktop builds or the independent desktop release jobs, and its advisory CI lanes do not gate merge.

Non-blocking follow-up for the author: the PR description still describes the reverted Crux/native-shell architecture and should be updated to the Dioxus reality before or shortly after merge (paste-ready summary is in my review comment above).

PR #253, opened 2026-07-17
fix: exclude AD suite from default installs; show install time estimate

jtomek-strike48:
## Summary

Two related fixes to the Settings -> Tools install flow.

**1. AD suite no longer installed by a default install.**
"Install all recommended" sweeps every dependency whose `recommended` flag is
true. That flag defaults to `true`, so `bloodhound`, `certipy`, `kerbrute`, and
`netexec` were being installed by a plain default install despite being
specialized AD-engagement tooling. They now carry `.recommended(false)`,
matching the existing Metasploit/Burp precedent. Per-tool Install still installs
them on demand (`install_by_binary` deliberately ignores `recommended`).

**2. Install progress now shows how long it will take.**
The install UI previously rendered only an indeterminate animated bar. The
command layer runs installs to completion without streaming byte-level progress,
so a true percentage is not feasible; instead we surface a labeled time
estimate:

- `catalog::estimated_install_secs(method, binary)` returns a soft per-mechanism
  default (pacman 90s, apt 60s, custom 120s) with per-binary overrides for
  heavyweight installers (metasploit 360s, zap 150s, webwright 180s), exposed as
  `CatalogItem::estimated_secs`.
- A 1s ticker fills a determinate bar toward the estimate, capped at 95% until
  the install actually completes, with a `0:42 elapsed - ~2:00 expected` label.
  Bulk install estimates the sum of pending recommended entries. Tools without
  an estimate keep the indeterminate bar.

Estimates are deliberately soft hints (mirror speed / package size vary); the
bar is tuned to read slightly slow rather than pin at 100% and spin.

## Test plan

- [x] `cargo test --workspace --locked --features pentest-platform/desktop-pcap`
      - catalog 14/14 (incl. new `estimated_secs_defaults_by_method_and_overrides_by_binary`,
        `ad_suite_is_not_in_recommended_default_set`)
      - settings_page 4/4 (`format_duration`, three `install_progress` cases)
- [x] Live-verified in the web (liveview) app against a local Strike48 studio:
      AD suite shows individual Install buttons (not auto-installed); a BloodHound
      install showed the elapsed clock advancing 0:01 -> 0:02 against "~2:00 expected"
      with the determinate bar filling.

## Screenshot

Install in progress, showing the elapsed clock and estimate (and the AD suite now offered as individual opt-in installs rather than swept into the default set):

install progress bar

### DIFF
--- a/crates/tools/src/catalog.rs
+++ b/crates/tools/src/catalog.rs
@@ -66,7 +66,48 @@ pub struct CatalogEntry {
     pub state: InstallState,
 }
 
+/// Rough expected install duration, in seconds, for a dependency. Used only to
+/// drive the UI's elapsed-vs-estimate progress bar — it is a soft hint, never a
+/// guarantee. The command layer runs installs to completion without streaming
+/// byte-level progress, so a true percentage is not available; a labeled
+/// estimate is the honest signal we can give.
+///
+/// Values are deliberately generous (a bar that fills slightly slow reads
+/// better than one that pins at 100% and keeps spinning). The default covers a
+/// single `pacman`/`apt` package fetch on a warm mirror; heavyweight custom
+/// installers override by `id`.
+pub fn estimated_install_secs(method: &InstallMethod, binary_name: &str) -> u32 {
+    // Sensible per-mechanism defaults.
+    let base = match method {
+        // Manual installs are operator-driven; no automated timer applies.
+        InstallMethod::Manual { .. } => 0,
+        // A single package plus the -Sy DB refresh on the sandbox rootfs.
+        InstallMethod::Pacman => 90,
+        InstallMethod::AptHost => 60,
+        // Bespoke installers vary widely; refined by binary below.
+        InstallMethod::Custom { .. } => 120,
+    };
+
+    // Per-tool overrides for the installers whose real-world duration is far
+    // from the mechanism default. Keyed by the catalog binary name.
+    match binary_name {
+        // Metasploit pulls a very large package set + Ruby gems.
+        "msfconsole" => 360,
+        // ZAP downloads a sizable Java app bundle.
+        "zaproxy" | "zap" => 150,
+        // webwright installs a Python env plus a Playwright Chromium download.
+        "webwright" => 180,
+        _ => base,
+    }
+}
+
 impl CatalogEntry {
+    /// Rough expected install duration in seconds. See
+    /// [`estimated_install_secs`]. `0` means no automated install (manual).
+    pub fn estimated_install_secs(&self) -> u32 {
+        estimated_install_secs(&self.install_method, &self.binary_name)
+    }
+
     /// Whether this entry can be installed automatically in the current mode.
     pub fn is_auto_installable(&self) -> bool {
         match &self.install_method {
@@ -357,6 +398,10 @@ pub struct CatalogItem {
     pub manual_instructions: Option<String>,
     /// Tools that declared this dependency.
     pub used_by: Vec<String>,
+    /// Rough expected install duration in seconds, for the UI's
+    /// elapsed-vs-estimate progress bar. `0` for manual-only entries. Soft hint.
+    #[serde(default)]
+    pub estimated_secs: u32,
 }
 
 impl From<&CatalogEntry> for CatalogItem {
@@ -381,6 +426,7 @@ impl From<&CatalogEntry> for CatalogItem {
             auto_installable: e.is_auto_installable(),
             manual_instructions: e.manual_instructions(),
             used_by: e.used_by.clone(),
+            estimated_secs: e.estimated_install_secs(),
         }
     }
 }
@@ -633,6 +679,7 @@ mod tests {
             auto_installable: true,
             manual_instructions: None,
             used_by: vec!["zap".into()],
+            estimated_secs: 150,
         };
         let json = serde_json::to_string(&item).unwrap();
         let back: CatalogItem = serde_json::from_str(&json).unwrap();
@@ -685,6 +732,75 @@ mod tests {
         deduped.dedup();
         assert_eq!(cats.len(), deduped.len(), "categories must be distinct");
         assert!(cats.iter().any(|c| c == "active_directory"));
+    }
+
+    #[test]
+    fn estimated_secs_defaults_by_method_and_overrides_by_binary() {
+        // Mechanism defaults.
+        assert_eq!(
+            estimated_install_secs(&InstallMethod::Pacman, "nmap"),
+            90,
+            "pacman default"
+        );
+        assert_eq!(
+            estimated_install_secs(&InstallMethod::AptHost, "some-apt-tool"),
+            60,
+            "apt default"
+        );
+        assert_eq!(
+            estimated_install_secs(
+                &InstallMethod::Custom {
+                    id: "certipy".into()
+                },
+                "certipy"
+            ),
+            120,
+            "custom default"
+        );
+        // Manual has no automated timer.
+        assert_eq!(
+            estimated_install_secs(
+                &InstallMethod::Manual {
+                    url: None,
+                    instructions: "x".into()
+                },
+                "burpsuite"
+            ),
+            0,
+            "manual = 0"
+        );
+        // Heavyweight per-binary overrides beat the mechanism default.
+        assert_eq!(
+            estimated_install_secs(
+                &InstallMethod::Custom {
+                    id: "metasploit".into()
+                },
+                "msfconsole"
+            ),
+            360,
+            "metasploit override"
+        );
+        assert!(
+            estimated_install_secs(&InstallMethod::Custom { id: "zap".into() }, "zaproxy")
+                > estimated_install_secs(&InstallMethod::Pacman, "nmap"),
+            "zap should estimate longer than a plain pacman package"
+        );
+    }
+
+    #[tokio::test]
+    async fn ad_suite_is_not_in_recommended_default_set() {
+        // Specialized AD-engagement tooling must not be swept in by the bulk
+        // "install all recommended" action (the operator installs it on demand).
+        let catalog = build_catalog().await;
+        for bin in ["bloodhound-python", "certipy", "kerbrute", "nxc"] {
+            let entry = catalog.iter().find(|e| e.binary_name == bin);
+            if let Some(entry) = entry {
+                assert!(
+                    !entry.recommended,
+                    "{bin} must be opt-in, not part of the recommended default set"
+                );
+            }
+        }
     }
 
     #[tokio::test]
--- a/crates/tools/src/external/ad/bloodhound.rs
+++ b/crates/tools/src/external/ad/bloodhound.rs
@@ -63,7 +63,10 @@ impl PentestTool for BloodHoundTool {
                     "BloodHound AD data collector (bloodhound-python ingestor)",
                 )
                 .custom_installer("bloodhound")
-                .category(ToolCategory::ActiveDirectory),
+                .category(ToolCategory::ActiveDirectory)
+                // Specialized AD-engagement tooling: installed on explicit
+                // operator action, not swept in by "install all recommended".
+                .recommended(false),
             )
             .param(ToolParam::required(
                 "domain",
--- a/crates/tools/src/external/ad/certipy.rs
+++ b/crates/tools/src/external/ad/certipy.rs
@@ -52,7 +52,10 @@ impl PentestTool for CertipyTool {
                     "Certipy - Active Directory Certificate Services (AD CS) attack tool",
                 )
                 .custom_installer("certipy")
-                .category(ToolCategory::ActiveDirectory),
+                .category(ToolCategory::ActiveDirectory)
+                // Specialized AD-engagement tooling: installed on explicit
+                // operator action, not swept in by "install all recommended".
+                .recommended(false),
             )
             .param(ToolParam::required(
                 "target",
--- a/crates/tools/src/external/ad/kerbrute.rs
+++ b/crates/tools/src/external/ad/kerbrute.rs
@@ -50,7 +50,10 @@ impl PentestTool for KerbruteTool {
                     "Kerbrute - Kerberos pre-auth username enumeration and password spraying",
                 )
                 .custom_installer("kerbrute")
-                .category(ToolCategory::ActiveDirectory),
+                .category(ToolCategory::ActiveDirectory)
+                // Specialized AD-engagement tooling: installed on explicit
+                // operator action, not swept in by "install all recommended".
+                .recommended(false),
             )
             .param(ToolParam::required(
                 "mode",
--- a/crates/tools/src/external/ad/netexec.rs
+++ b/crates/tools/src/external/ad/netexec.rs
@@ -52,7 +52,10 @@ impl PentestTool for NetExecTool {
                     "NetExec (nxc) - network service exploitation across SMB/WinRM/LDAP/MSSQL/SSH",
                 )
                 .custom_installer("netexec")
-                .category(ToolCategory::ActiveDirectory),
+                .category(ToolCategory::ActiveDirectory)
+                // Specialized AD-engagement tooling: installed on explicit
+                // operator action, not swept in by "install all recommended".
+                .recommended(false),
             )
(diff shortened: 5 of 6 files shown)


ngdevo:
## PR #253 Review — fix: exclude AD suite from default installs; show install time estimate

Reviewed the full diff (6 files, +287/-14): the catalog changes (`recommended(false)` on 4 AD tools + new
`estimated_install_secs` API), the settings-page UI progress bar, and the accompanying tests. Cross-referenced
the pre-PR `main` catalog to verify (a) the `install_by_binary` code path — which deliberately ignores
`recommended` — is unchanged, (b) `install_all_recommended` filters by `entry.recommended`, and (c) the
`CatalogItem` wire shape treatment of the new field. Findings below by severity.

Two independent changes bundled here. Both are small, safe, and well-tested. Bundling is defensible — they
share the settings-page install UI surface — but they could just as easily have shipped separately.

---

### CRITICAL
None.

---

### HIGH

**H1 — `estimated_secs: 0` for `Manual` renders "0:00 elapsed · ~0:00 expected" if a manual install ever
enters the "installing" branch**
`estimated_install_secs` returns `0` for `InstallMethod::Manual { .. }`. In `install_progress`, `0` triggers
the indeterminate branch (correct). But the settings page conditionally renders the "Installing... {msg}"
block only when `installing() == Some(item.binary_name.clone())` — which the manual path shouldn't reach if
`auto_installable == false`. Verified in the pre-PR catalog: `Manual { .. } => false` for
`is_auto_installable`, so the button that sets `installing` isn't rendered for manual entries. This is
correct today.

The concern is durability: the invariant "`estimated_secs == 0` iff `auto_installable == false`" is not
enforced anywhere. If a future refactor makes a `Manual` entry auto-installable (or a `Custom` installer
returns 0 seconds for some reason), the UI shows the paradox of "Installing…" plus "0:00 expected." Not a
security concern, but the UX degrades silently.

Fix (small): either
- `debug_assert!` in `install_progress` that `elapsed_secs == 0 || estimate_secs > 0` at first tick, OR
- Guard in the UI: `if installing() == Some(bin) && item.auto_installable { ... }`, OR
- Add a compile-time test asserting no `auto_installable && estimated_secs == 0` combination exists in the
catalog.

The third is the strongest — it catches regressions at test time rather than in the UI.

**H2 — Ticker effect can spawn duplicate tasks if `installing` changes rapidly**
`crates/ui/src/components/settings_page.rs:~110` — the `use_effect` reads `installing()` reactively; when the
value changes, the effect re-runs and spawns a new async task. The old task's `if installing() !=
started_for { break; }` guard is correct, but there is a race window:
- T0: install A starts; effect spawns task-A with `started_for = Some("A")`.
- T1: install A finishes; `installing.set(None)`.
- T2: task-A wakes from its 1s sleep, sees `installing() == None != Some("A")`, breaks. Good.
- T2': *simultaneously* the effect re-runs with `active == false` and returns early. Good.

BUT if the user starts install B before task-A's sleep expires:
- T1: user starts install B; `installing.set(Some("B"))`; effect re-runs; task-B is spawned with
`started_for = Some("B")`.
- T2: task-A wakes, sees `Some("B") != Some("A")`, breaks. Good.
- Between T1 and T2, two tickers are alive. Task-A won't do harm (it'll break on next wake), but the elapsed
counter is set by both tasks in the overlap window — task-A's last-second increment could clobber the
reset-to-0 that the click handler did for install B.

Concretely: the click handler for B sets `install_elapsed.set(0)` synchronously; task-A wakes ~1s later,
computes `install_elapsed() + 1` and writes it. If A's wake happens between the click handler running and
task-B's first tick, the counter is bumped to 1 based on the stale (pre-reset) elapsed. Small user-visible
glitch: the elapsed counter for B may start at 1 instead of 0 on rapid re-start.

Fix: gate the increment on `installing() == started_for` (already gated for the break, but not for the
`install_elapsed.set(...)` — re-check before writing). One extra line.

---

### MEDIUM

**M1 — `estimated_install_secs` uses a match on `binary_name` string — case-sensitive and typo-prone**
The overrides are `"msfconsole" | "zaproxy" | "zap" | "webwright"`. If a future entry's `binary_name` differs
in case or spelling (e.g., `"ZAP"`, `"Metasploit"`), the override silently doesn't apply and the mechanism
default is used. That produces a noisy-but-not-broken bar (2:00 for a 6:00 metasploit install), which is the
"reads slightly slow" mode the PR says it wants — but by accident, not by design.

Better shape: co-locate the estimate with the tool definition, similar to how `recommended(false)` is now
attached via builder in each tool file. E.g., `.estimated_install_secs(360)` in `metasploit.rs`. Same
pattern the AD-suite change follows. Removes the string-match seam entirely.

**M2 — `install_progress` cap at 0.95 is not covered by a docstring on the constant**
`let fraction = (elapsed_secs as f64 / estimate_secs as f64).min(0.95);` — the magic number `0.95` matches
the PR body's "capped at 95%" but a reader of the function has to infer why. The test names the intent
(`install_progress_caps_below_full_when_overrunning`) but the function itself doesn't. Extract:
```rust
const MAX_PROGRESS_UNTIL_COMPLETE: f64 = 0.95;
Or add a one-line comment referencing the "never claim 100% before actually complete" invariant. Minor
readability ask, not a bug.

M3 — install_progress returns a String allocation on every render tick (1Hz × N visible items)
The formatted label is allocated every second while an install runs. Cheap in absolute terms (a Vec<u8>
per tick) but wasteful in Dioxus's re-render model — the string is compared against the previous render's
value by the framework and only diffed if changed, so this is not a correctness issue. Just noting: if
progress ever moves to sub-second updates or many concurrent installs render at once, prefer computing the
label lazily or memoizing.

M4 — Bulk-install estimate sums estimated_secs but doesn't account for parallelism
The bulk estimate is items.iter().filter(...).map(|i| i.estimated_secs).sum(). If install_all_recommended
runs installs sequentially, the sum is right. If it runs any of them concurrently (verified against
crates/tools/src/catalog.rs: the pre-PR code shows sequential iteration in install_all_recommended, so
this is correct today), the sum overestimates. Not a bug today; would become one silently if the installer
ever parallelizes.

Add a comment stating the assumption ("bulk installs run sequentially per install_all_recommended; if that
changes, this sum needs to change too") so a future refactor can't silently break the estimate. Or, better,
compute the estimate inside install_all_recommended's progress sink so the two definitions can't drift.

M5 — Only the AD list is tested for !recommended; positive assertions for the recommended set are
missing
ad_suite_is_not_in_recommended_default_set correctly asserts the four AD binaries are excluded. But there
is no complementary test asserting the expected recommended set is present (nmap, wireshark,
whatever-the-default-is). If a future refactor accidentally sets recommended(false) on a legitimately
default tool, no test would catch it. Add a companion test:
assert!(catalog.iter().any(|e| e.binary_name == "nmap" && e.recommended),
      "nmap must remain in the recommended default set");
Or, more robustly: pin the entire recommended-default set as a sorted list and assert it matches.

M6 — if let Some(entry) = entry { ... } in ad_suite_is_not_in_recommended_default_set silently passes
if the entry is missing
let entry = catalog.iter().find(|e| e.binary_name == bin);
if let Some(entry) = entry {
  assert!(!entry.recommended, "{bin} must be opt-in");
}
If any of the four AD binaries is removed from the catalog entirely, the test still passes without warning.
The comment "specialized AD-engagement tooling" reads as an invariant, but the assertion is conditional on
presence. Prefer:
let entry = catalog.iter().find(|e| e.binary_name == bin)
  .unwrap_or_else(|| panic!("AD tool {bin} not found in catalog"));
assert!(!entry.recommended, ...);
Trades permissiveness for regression-visibility. If AD tools are deliberately removed later, the test
should be updated; you don't want it silently passing.

---
LOW

L1 — \u{00b7} middle-dot in the label is fine but not portable across all terminals
format!("{elapsed_label} elapsed \u{00b7} ~{expected} expected") — renders ·. Displays correctly in
Dioxus/web UI; if any log line ever writes this string to a limited-charset terminal (SSH into a legacy
box), it becomes a mojibake. Not a real concern for a desktop UI label; noting for completeness.

L2 — format_duration doesn't handle overflow
u32 seconds → hours as u32. u32::MAX / 3600 ≈ 1.2M hours — nowhere near reachable. Not a bug. Just
naming: could use Duration from std::time for symmetry with other tokio/std interfaces, but that's a
style preference.

L3 — Test names use snake_case with method-and-behavior separated by "and"
estimated_secs_defaults_by_method_and_overrides_by_binary — long but readable. Consider splitting into
two tests (estimated_secs_defaults_by_method and estimated_secs_overrides_by_binary) so a failure in one
half doesn't obscure the other's assertions. Small.

L4 — #[tokio::test] on ad_suite_is_not_in_recommended_default_set because build_catalog is async
Fine, but the test is otherwise pure. If build_catalog().await ever gains I/O side effects (currently just
walks a static registry), the test becomes non-hermetic. Note it as an assumption in a comment.

L5 — No test for install_progress(0, 0) — the zero-elapsed no-estimate case
Edge case: an install just started, no ticks yet, no estimate. Should return (0.0, "0:00 elapsed").
Trivial to add. Would also pin the format-of-zero behavior against future format_duration regressions.

L6 — No test for install_progress with elapsed == estimate (exact match)
(elapsed / estimate).min(0.95) — at exact match, returns 1.0.min(0.95) = 0.95. The
_caps_below_full_when_overrunning test asserts this for elapsed > estimate but not for elapsed == estimate. Add a case pinning the boundary.

L7 — Comment "matching the existing Metasploit/Burp precedent" isn't verifiable from this diff
The four AD-suite files add .recommended(false). The precedent is claimed but not cited by file path.
Cheap to link: metasploit.rs:LN / burp.rs:LN. Helps a future maintainer verify the pattern hasn't
drifted.

L8 — Doc for estimated_secs: u32 field on CatalogItem says "Soft hint" — good — but #[serde(default)]
makes older UI clients silently deserialize this as 0
Correct for wire compat (an older UI without this field gets 0 → indeterminate bar, same as pre-PR
behavior). Worth a one-line comment noting the intent (// serde default => older UI shows indeterminate bar) so a future reviewer doesn't remove the attribute and break rollout.

---
INFO / POSITIVE OBSERVATIONS

- Right precedent match: the AD-suite .recommended(false) addition mirrors the existing
Metasploit/Burp Suite treatment. Consistent, easy for reviewers to check, and the invariant
"specialized/dual-use → opt-in" is now documented in the entry itself.
- install_by_binary deliberately ignores recommended: per-tool install still works. Verified in the
pre-PR catalog at install_by_binary(...) — no change needed. Good separation of "bulk auto" vs "explicit
per-tool" install policies.
- Honest UX for progress: the PR body correctly acknowledges "true percentage is not feasible" and
chooses a labeled time estimate. This is the right call — a fake percentage would erode operator trust
when it hits 100% and keeps spinning.
- .min(0.95) cap: prevents the exact bug the PR body names ("reads slightly slow rather than pin at
100% and spin"). Textbook good defensive UI code.
- #[serde(default)] on the new field: correct wire-compat choice. Older UI clients keep working
(indeterminate bar); newer ones benefit from the estimate. No lockstep upgrade required.
- format_duration tests cover the H:MM:SS boundary (3661 seconds → "1:01:01"). Small but proves the
overflow path.
- Single-ticker guarantee: the effect's started_for sentinel is the right primitive for "at most one
ticker per install." H2 is a race in the increment, not the sentinel logic itself.
- Zero-visible-blast-radius for the AD-suite change: an operator who wants BloodHound still gets it via
the per-tool Install button. Only the bulk sweep behavior changes. That's the minimum-blast-radius fix
for a "we're pulling too much by default" bug.

---
SECURITY-FOCUSED SUMMARY

The AD-suite exclusion is a small but meaningful hardening for the default install posture. An operator on a
non-AD engagement no longer has BloodHound / Certipy / Kerbrute / NetExec sitting on disk (attack surface,
credential-scraping targets, EDR triggers) just because they clicked "install recommended."

┌─────────────────────────────────────┬───────────────────────────────────┬──────────────────────────────────────────┐
│               Concern               │              Before               │                  After                   │
├─────────────────────────────────────┼───────────────────────────────────┼──────────────────────────────────────────┤
│ Default-install pulls specialized   │ Present — bloodhound-python,      │ Removed. Operator installs on demand     │
├─────────────────────────────────────┼───────────────────────────────────┼──────────────────────────────────────────┤
│ AD tooling                          │ certipy, kerbrute, nxc all in     │ via per-tool Install (which correctly    │
├─────────────────────────────────────┼───────────────────────────────────┼──────────────────────────────────────────┤
│                                     │ recommended default set           │ ignores recommended).                    │
├─────────────────────────────────────┼───────────────────────────────────┼──────────────────────────────────────────┤
│ Progress bar as a mis-timing        │ Indeterminate spinner — no        │ Determinate bar with elapsed/expected    │
├─────────────────────────────────────┼───────────────────────────────────┼──────────────────────────────────────────┤
│ signal ("stuck" vs "still running") │ operator signal for run duration  │ label. Reduces the "did it hang?" false- │
├─────────────────────────────────────┼───────────────────────────────────┼──────────────────────────────────────────┤
│                                     │                                   │ alarm class. No new attack surface.      │
├─────────────────────────────────────┼───────────────────────────────────┼──────────────────────────────────────────┤
│ Per-tool Install for AD tools       │ Worked (install_by_binary ignores │ Unchanged. AD tools remain usable when   │
├─────────────────────────────────────┼───────────────────────────────────┼──────────────────────────────────────────┤
│                                     │ recommended)                      │ the engagement calls for them.           │
├─────────────────────────────────────┼───────────────────────────────────┼──────────────────────────────────────────┤
│ Attack surface reduction on host    │ Broader (four extra binaries      │ Narrower default. Real reduction in      │
├─────────────────────────────────────┼───────────────────────────────────┼──────────────────────────────────────────┤
│                                     │ on disk by default)               │ credential-scanner and Kerberos-attack   │
├─────────────────────────────────────┼───────────────────────────────────┼──────────────────────────────────────────┤
│                                     │                                   │ tooling on non-AD engagements.           │
├─────────────────────────────────────┼───────────────────────────────────┼──────────────────────────────────────────┤
│ Progress-bar-as-side-channel        │ N/A                               │ Estimated seconds is a soft hint         │
├─────────────────────────────────────┼───────────────────────────────────┼──────────────────────────────────────────┤
│                                     │                                   │ derived from binary_name — not user      │
├─────────────────────────────────────┼───────────────────────────────────┼──────────────────────────────────────────┤
│                                     │                                   │ input, no injection surface.             │
└─────────────────────────────────────┴───────────────────────────────────┴──────────────────────────────────────────┘

Net: this is a UX/hardening change with no offensive-tool policy weakening. The AD-suite exclusion is the
more important half — it's a real posture improvement.

---
RECOMMENDATIONS BEFORE MERGE

1. Address H2 (duplicate-ticker race): guard the install_elapsed.set(...) write on
installing() == started_for. One-line fix.
2. Address H1 (Manual-with-installing paradox): add a compile-time test asserting no
auto_installable && estimated_secs == 0 combination exists. Cheap regression guard.
3. Consider M1 (co-locate estimate with tool definition via builder): matches the pattern the
AD-suite change itself uses. If deferred, add a test enumerating the four override strings and asserting
each matches an entry in the catalog (guards against typo-drift).
4. Address M5/M6 together: pin the recommended default set and assert AD tools are present in the
catalog before asserting !recommended. Prevents silent regressions on both sides.
5. Optional: extract the 0.95 cap constant (M2), add the missing install_progress edge cases (L5, L6).

Everything else can land as follow-ups. Correct fix, correct precedent, safe scope, comprehensive tests.
Ready to merge with H1 and H2 addressed.
```

cchao-devo (approved):
Approve. Verified locally — fmt/clippy clean (with the liveview feature set), catalog 14/14 and settings_page 4/4 matching your results.

  - AD-suite exclusion is correctly scoped: .recommended(false) drops the four tools from the install_all_recommended sweep (filters on !recommended) while per-tool install still works
  (deliberately doesn't gate on recommended, per the catalog.rs comment). Consistent with the Metasploit/Burp precedent, and the test iterates the real catalog by binary name. 👍
  - Estimate UI is clean: pure estimated_install_secs/install_progress/format_duration, fraction capped at 0.95 so the bar never claims early completion, and the ticker effect guards
  against concurrent tickers (installing() != started_for break). Soft-hint framing is honest given no byte-level progress is available.

  LGTM. 🚢

PR #254, opened 2026-07-17
fix: harden install-progress ticker and strengthen catalog tests

jtomek-strike48:
## Summary

Follow-up to #253 (merged), applying three fixes surfaced by a post-merge multi-agent review pass. #253 was squash-merged while the review was in flight, so these hardening changes land as a separate PR.

**1. Install-progress ticker race (settings_page.rs).**
The elapsed-clock ticker compared a captured `installing()` snapshot, leaving a narrow window where restarting an install within ~1s of the previous one could let a stale ticker advance the clock. Added a monotonic `install_generation` bumped on each install start; the ticker captures its generation and exits as soon as a newer session begins. The clock reset now lives in the ticker effect so it is atomic with the spawn rather than racing the onclick handlers. Impact was cosmetic (clock ticking slightly fast in a rare transition), never a crash or stuck state — this removes the window entirely and corrects the stale comment.

**2. AD-suite guard test could pass vacuously (catalog.rs).**
`ad_suite_is_not_in_recommended_default_set` used `if let Some(entry)` with no existence assertion, so a rename/refactor dropping a binary from the catalog would silently retire the regression guard for the primary #253 fix. Now asserts each binary is present before checking its `recommended` flag.

**3. Missing `webwright` estimate-override test (catalog.rs).**
The estimator test covered the metasploit (360s) and zap (150s) overrides but not webwright (180s), so a refactor could regress it to the Custom default undetected. Added the assertion.

## Test plan

- [x] `cargo test --workspace --locked --features pentest-platform/desktop-pcap` — catalog 14/14, settings_page 4/4

### DIFF
--- a/crates/tools/src/catalog.rs
+++ b/crates/tools/src/catalog.rs
@@ -836,6 +836,16 @@ mod tests {
             360,
             "metasploit override"
         );
+        assert_eq!(
+            estimated_install_secs(
+                &InstallMethod::Custom {
+                    id: "webwright".into()
+                },
+                "webwright"
+            ),
+            180,
+            "webwright override (Python env + Playwright Chromium download)"
+        );
         assert!(
             estimated_install_secs(&InstallMethod::Custom { id: "zap".into() }, "zaproxy")
                 > estimated_install_secs(&InstallMethod::Pacman, "nmap"),
@@ -849,13 +859,17 @@ mod tests {
         // "install all recommended" action (the operator installs it on demand).
         let catalog = build_catalog().await;
         for bin in ["bloodhound-python", "certipy", "kerbrute", "nxc"] {
-            let entry = catalog.iter().find(|e| e.binary_name == bin);
-            if let Some(entry) = entry {
-                assert!(
-                    !entry.recommended,
-                    "{bin} must be opt-in, not part of the recommended default set"
-                );
-            }
+            // Assert the entry EXISTS before checking its flag: without this the
+            // test would pass vacuously if a rename/refactor dropped the binary
+            // from the catalog, silently retiring the regression guard.
+            let entry = catalog
+                .iter()
+                .find(|e| e.binary_name == bin)
+                .unwrap_or_else(|| panic!("{bin} should be present in the catalog"));
+            assert!(
+                !entry.recommended,
+                "{bin} must be opt-in, not part of the recommended default set"
+            );
         }
     }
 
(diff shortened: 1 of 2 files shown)


cchao-devo (approved):
Review: Pick PR #254 — "harden install-progress ticker and strengthen catalog tests"

  Verdict: ✅ Approve. Small, correct, well-scoped hardening follow-up to #253. No blocking findings. One low-severity note worth leaving as a comment for the author.

  Lenses run: mechanical gates (fmt/clippy/test) + manual source trace + 3 review agents (silent-failure/concurrency, test-coverage, code-review/conventions). Reviewed in my scratch
  checkout pick_08/pick — your pick_07 untouched, no cluster involved.

  Mechanical gates (run locally on pr-254)

  - cargo fmt --all --check — clean
  - cargo clippy --workspace --locked --features pentest-platform/desktop-pcap -- -D warnings — clean
  - cargo test -p pentest-tools — PR's own tests pass (ad_suite_is_not_in_recommended_default_set ok, estimator/webwright ok)
  - One failing test, all_tools_registered (wifi_scan_detailed not in capabilities), is in pentest-core — a crate this PR doesn't touch. I checked out origin/main and it fails there 
  too, so it's the known pre-existing failure, not this PR.

  The substantive fix (ticker race) — correct & idiomatic

  I traced the Dioxus reactivity myself and all three agents concur:
  - Both — and only — install-start sites bump install_generation (:357 __all__, :458 single tool), each immediately before installing.set(Some(...)). Stop sites (:385, :479) correctly
  don't bump.
  - The effect subscribes to both installing and install_generation, so a same-tool restart (the original bug — installing value looks unchanged) now re-triggers via the generation
  change. Stale ticker breaks on its next wake before incrementing (:135), so no double-count and no wrong tick.
  - install_elapsed.set(0) moved into the effect is genuinely atomic-with-spawn; the effect does not subscribe to install_elapsed (the read is inside the spawned async block), so no
  self-trigger storm. The rewritten doc-comment is accurate — it correctly retires the old, false "at most one ticker" claim.

  Test hardening — one real, one shallow

  - #2 (AD-suite existence assertion): real. Verified all four binaries (bloodhound-python, certipy, kerbrute, nxc) are genuinely in the catalog fixture (each declares
  .recommended(false)), so the new unwrap_or_else(panic!) closes the vacuous-pass hole rather than spuriously panicking.
  - #3 (webwright estimator assertion): shallow but harmless. Correct as a pure-function test, but the "webwright" => 180 arm is unreachable via the real catalog — webwright's deps are
  python3/playwright, no entry has binary_name == "webwright". Pre-existing (from #253), neither caused nor fixed here.

PR #258, opened 2026-07-17
chore: gitignore .env backup copies

jtomek-strike48:
## Summary

`.env` is gitignored, but ad-hoc backup copies like `.env.bak-238` are not, so a `git add -A` (or an editor's autosave copy) could commit a secret-bearing copy of `.env`. Those slip past the `.env` rule and would only be caught by the gitleaks full-history scan, after which a history rewrite is needed to remove them.

This adds `.env.bak`, `.env.bak-*`, and `.env.*.bak` to the tracked `.gitignore` so the protection is shared across all clones rather than living in each operator's local `.git/info/exclude`.

Scoped intentionally to env-backup patterns only. Personal build dirs (`target-check/`) and `.claude/` scratch stay clone-local, since `/target/` and `.claude/settings.local.json` are already the repo convention.

## Test plan

- [x] `git check-ignore` confirms `.env.bak-238`, `.env.bak`, `.env.production.bak` are now ignored
- [x] `.env`, `.env.local`, `.env.*.local` still ignored (unchanged behavior)
- [x] No tracked files are newly ignored (diff is additive, 4 lines)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -37,6 +37,10 @@ logs/
 .env
 .env.local
 .env.*.local
+# Backups/copies of .env - secret-bearing, never commit (e.g. .env.bak-238).
+.env.bak
+.env.bak-*
+.env.*.bak
 
 # AI tooling local overlays — private, operator-specific, never committed.
 # See CLAUDE.md "Local Overrides" section.


cchao-devo (approved):
Review: Pick PR #258 — chore: gitignore .env backup copies

  Verdict: ✅ Approve. Correct, minimal, well-scoped defensive change. It closes a real gap (secret-bearing .env.bak-* copies slipping past the .env rule into git add -A), the patterns
  match exactly what's intended without over-matching, and no tracked file is affected. Verified in pick_08/pick.

  - Patterns correct & anchored: .env.bak, .env.bak-*, .env.*.bak catch the backup shapes (incl. the .env.bak-238 example from the body) while leaving .env.example/.env.sample/.envbak
  committable.
  - Safely additive: the only tracked .env* files (.env.example, .envrc) are not matched → the "no tracked files newly ignored" claim confirmed.
  - Good comment on the added lines explaining why; scoping rationale in the body (env-backups only, leaving target-check//.claude/ clone-local) is sound and consistent with existing repo
  convention.

PR #267, opened 2026-07-18
fix(#266): web UI resolves a free port instead of hard-binding 3000

### LINKED ISSUE #266: Web/liveview server: hardcoded ports collide silently with other local Dioxus apps (add PORT env + auto-increment + loud logging)
## Problem

Pick's local servers bind **hardcoded ports** with no override:

- `apps/web/src/main.rs:34` -> `0.0.0.0:3000`, bound via `.expect("failed to bind TCP listener")` (panics on collision).
- `crates/ui/src/liveview_server.rs:121` -> `LIVEVIEW_PORT = 3030` (desktop/headless internal liveview server).

When another local process already owns the port, one of two bad things happens:

1. **Hard panic** - `pentest-web` dies immediately on `.expect()` with a raw bind error.
2. **Silent wrong-app (worse)** - if the other app is *also* a Dioxus liveview app (real example: **4n6 Nexus** / `nexus-web` on :3000), then `http://localhost:3000/app` serves *Nexus's* shell, which has the identical `<title>Dioxus Liveview App</title>` and dioxus websocket glue. The operator points a browser at what they think is Pick, sees a Dioxus app, and cannot tell they are looking at the wrong process. This actively wasted debugging time.

Running Pick and 4n6 Nexus (or two Pick instances) on the same machine is a normal DFIR/pentest workstation scenario, so this will recur.

## Proposed fix

Introduce a small, predictable port-resolution ladder for the web app (and apply the same pattern to the 3030 liveview server):

1. **`PORT` env var** - explicit operator control (`PORT=3005 pentest-web`).
2. Fall back to the current **default (3000 / 3030)**.
3. If the chosen port is taken, **auto-scan upward** to the next free port (3001, 3002, ...).
4. **Log the chosen bind loudly**, e.g. `Pick web UI available at http://localhost:3001` - so the operator always knows where it landed (auto-increment without logging just moves the "which port?" confusion).
5. Replace the `.expect()` panic with a clear, actionable error if binding is truly impossible.

### Why this shape (not blind auto-increment alone)
- `PORT` env gives determinism when the operator wants a fixed port (scripts, reverse proxies).
- Auto-increment gives zero-config coexistence with Nexus / a second Pick.
- Loud logging is the piece that prevents the silent-wrong-app failure mode: the URL is always printed.

## Scope

- `apps/web/src/main.rs`: read `PORT`, default 3000, auto-scan on collision, log final URL, drop `.expect()`.
- `crates/ui/src/liveview_server.rs`: make `LIVEVIEW_PORT` (3030) similarly overridable/collision-tolerant; audit `ServerConfig.port` callers.
- `apps/mobile/src/main.rs` references the local Axum port (line ~37) - verify it discovers the actual chosen port rather than assuming a constant.
- Consider a shared helper (`bind_with_fallback(preferred: u16) -> (TcpListener, u16)`) so web/liveview share one implementation.

## Acceptance criteria

- [ ] `PORT` env var overrides the web bind port; default unchanged when unset.
- [ ] On collision, the server binds the next free port instead of panicking.
- [ ] The actual bound URL is logged at startup (INFO).
- [ ] Pick and another local Dioxus app (e.g. 4n6 Nexus on :3000) can run simultaneously without manual intervention.
- [ ] The 3030 liveview server follows the same pattern; mobile port discovery verified.
- [ ] fmt/clippy(-D warnings)/test green at workspace scope (--features pentest-platform/desktop-pcap).

## Notes

Discovered while trying to live-test the #259 Settings Tools grid: `nexus-web` (4n6 Nexus) held :3000, so the browser showed Nexus's Dioxus shell and Pick appeared "not running." The hardcoded `:3000` + panic-on-bind + look-alike shell combination is the root cause.

## Recommended agent
devops-engineer (port/bind handling) or build-error-resolver; rust-reviewer for the shared helper.

jtomek-strike48:
## Summary

The web app hard-bound `0.0.0.0:3000` via `.expect()`. On a port collision it either **panicked**, or - when another local Dioxus app already held 3000 (real case: **4n6 Nexus** / `nexus-web`) - the operator saw that app's near-identical `<title>Dioxus Liveview App</title>` shell with **no signal they were on the wrong process**. Running Pick alongside another local Dioxus app is a normal DFIR/pentest-workstation scenario, so this recurred.

Closes #266.

## What changed (`apps/web/src/main.rs`)

- **`PORT` env var** for explicit operator control (default 3000); a malformed `PORT` warns and falls back rather than silently masquerading as the default.
- **Auto-increment** to the next free port on `AddrInUse`, bounded to 20 attempts so a misconfigured host can't spin forever. Only `AddrInUse` retries; any other bind error is fatal.
- **Loud logging** of the actual bound URL: `Pick web UI available at http://localhost:<port>` - this is the piece that prevents the silent-wrong-app confusion.
- **No more panic**: a clear error + `exit(1)` if no port is free.

## Scope note (investigation finding)

The issue proposed also fixing the `3030` liveview server. On inspection that constant (`LIVEVIEW_PORT` in `crates/ui/src/liveview_server.rs`) is **vestigial** - that server binds a PID-based IPC socket (Unix socket / named pipe), never a TCP port, so it has no collision to fix and already avoids conflicts across simultaneous connectors. `apps/web` was the only fixed TCP bind in the codebase (other binds use ephemeral `:0`). Scope narrowed accordingly.

## Verification

- `cargo check -p pentest-web`, `cargo clippy -p pentest-web -- -D warnings`, `cargo fmt --check`: all clean.
- **Live behavioral test** (the actual scenario from #266): with `nexus-web` holding `:3000`, launched `pentest-web` and observed:
  ```
  Port 3000 in use, trying next
  Port 3000 was in use; bound 3001 instead (set PORT to pin a specific port)
  Pick web UI available at http://localhost:3001
  ```
  `ss` confirmed `pentest-web` on 3001 and `nexus-web` still on 3000 - both running simultaneously, no panic.

## Test plan

- [x] With nothing on the preferred port, binds it directly with no fallback (verified via known-free `PORT=3999` -> bound 3999, no warning; literal 3000 not tested since another local app holds it).
- [x] With 3000 occupied (4n6 Nexus), binds 3001 and logs the collision + chosen port.
- [x] `PORT=3456` binds 3456 (verified).
- [x] `PORT=garbage` warns and falls back (then 3000 busy -> bound 3001).

---
**Manual test run (2026-07-18):** all four scenarios verified against the built binary. Testing also surfaced that `PORT=0` (OS-assigned ephemeral) logged `localhost:0` instead of the real port; fixed in `b09ecbe` by reading the bound port from `listener.local_addr()` (now logs e.g. `:44005`). `PORT=0` re-verified after the fix.

### DIFF
--- a/apps/web/src/main.rs
+++ b/apps/web/src/main.rs
@@ -37,7 +37,13 @@ fn preferred_port() -> u16 {
 /// retry; any other bind error is fatal and returned as-is.
 async fn bind_with_fallback(preferred: u16) -> std::io::Result<(TcpListener, u16)> {
     for offset in 0..MAX_PORT_ATTEMPTS {
-        let port = preferred.saturating_add(offset);
+        // Stop rather than re-probe: near the u16 ceiling `preferred + offset`
+        // would saturate and re-target the same top port, so a busy port high
+        // in the range would spin uselessly. Skipping keeps every attempt a
+        // distinct port.
+        let Some(port) = preferred.checked_add(offset) else {
+            break;
+        };
         let addr = SocketAddr::from(([0, 0, 0, 0], port));
         match TcpListener::bind(&addr).await {
             Ok(listener) => {
--- a/apps/web/src/main.rs
+++ b/apps/web/src/main.rs
@@ -41,12 +41,17 @@ async fn bind_with_fallback(preferred: u16) -> std::io::Result<(TcpListener, u16
         let addr = SocketAddr::from(([0, 0, 0, 0], port));
         match TcpListener::bind(&addr).await {
             Ok(listener) => {
+                // Read the port back from the socket rather than trusting
+                // `port`: with PORT=0 the OS assigns an ephemeral port, and the
+                // caller must log the real one (not 0). Falls back to `port`
+                // only if local_addr() somehow fails.
+                let bound = listener.local_addr().map(|a| a.port()).unwrap_or(port);
                 if offset > 0 {
                     tracing::warn!(
-                        "Port {preferred} was in use; bound {port} instead (set PORT to pin a specific port)"
+                        "Port {preferred} was in use; bound {bound} instead (set PORT to pin a specific port)"
                     );
                 }
-                return Ok((listener, port));
+                return Ok((listener, bound));
             }
             Err(e) if e.kind() == std::io::ErrorKind::AddrInUse => {
                 tracing::debug!("Port {port} in use, trying next");
--- a/apps/web/src/main.rs
+++ b/apps/web/src/main.rs
@@ -4,9 +4,65 @@ use axum::Router;
 use dioxus::prelude::*;
 use dioxus_liveview::LiveviewRouter;
 use std::net::SocketAddr;
+use tokio::net::TcpListener;
 
 use pentest_core::config::ShellMode;
 use pentest_ui::{connector_app, mobile_css, theme_css, utils_css, ConnectorAppConfig};
+
+/// Default port for the web UI when `PORT` is unset.
+const DEFAULT_PORT: u16 = 3000;
+
+/// How many consecutive ports to try before giving up. Keeps auto-increment
+/// bounded so a misconfigured host can't spin forever.
+const MAX_PORT_ATTEMPTS: u16 = 20;
+
+/// Resolve the starting port from the `PORT` env var, falling back to
+/// [`DEFAULT_PORT`]. A malformed value is rejected loudly rather than silently
+/// ignored, so a typo'd `PORT` surfaces instead of masquerading as the default.
+fn preferred_port() -> u16 {
+    match std::env::var("PORT") {
+        Ok(raw) => raw.trim().parse().unwrap_or_else(|_| {
+            tracing::warn!(
+                "PORT='{raw}' is not a valid port number; falling back to {DEFAULT_PORT}"
+            );
+            DEFAULT_PORT
+        }),
+        Err(_) => DEFAULT_PORT,
+    }
+}
+
+/// Bind the first free port at or above `preferred`, scanning up to
+/// [`MAX_PORT_ATTEMPTS`] ports. Returns the listener plus the port actually
+/// bound so the caller can report the real URL. Only `AddrInUse` triggers a
+/// retry; any other bind error is fatal and returned as-is.
+async fn bind_with_fallback(preferred: u16) -> std::io::Result<(TcpListener, u16)> {
+    for offset in 0..MAX_PORT_ATTEMPTS {
+        let port = preferred.saturating_add(offset);
+        let addr = SocketAddr::from(([0, 0, 0, 0], port));
+        match TcpListener::bind(&addr).await {
+            Ok(listener) => {
+                if offset > 0 {
+                    tracing::warn!(
+                        "Port {preferred} was in use; bound {port} instead (set PORT to pin a specific port)"
+                    );
+                }
+                return Ok((listener, port));
+            }
+            Err(e) if e.kind() == std::io::ErrorKind::AddrInUse => {
+                tracing::debug!("Port {port} in use, trying next");
+                continue;
+            }
+            Err(e) => return Err(e),
+        }
+    }
+    Err(std::io::Error::new(
+        std::io::ErrorKind::AddrInUse,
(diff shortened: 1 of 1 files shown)


cchao-devo (approved):
Review: Pick PR #267 — fix(#266): web UI resolves a free port instead of hard-binding 3000
  
  Verdict: ✅ Approve. Correct, well-scoped fix for a real operator-confusion bug (silently serving another local Dioxus app's identical shell on a :3000 collision). Clean gates, sound
  logic, good error discrimination. No blockers — two minor non-blocking edge notes.

  Lenses run: mechanical gates (fmt/clippy/check) + manual control-flow trace + empirical edge-case check (saturating_add) + test-convention check. Verified in pick_08/pick; pick_07
  untouched, no cluster.

  Gates (on pr-267)
  
  - cargo fmt --all --check clean · cargo clippy -p pentest-web -- -D warnings clean · cargo check -p pentest-web clean

  Logic is correct

  - bind_with_fallback: retries only on AddrInUse (e.kind() == ErrorKind::AddrInUse); any other bind error returns immediately (fatal) — correct discrimination, a permission/address error
  won't get masked as a port scan. Bounded to MAX_PORT_ATTEMPTS=20 so a misconfigured host can't spin forever; exhaustion returns a clear AddrInUse error → exit(1) with a readable message
  (no panic). ✓
  - local_addr() readback (main.rs:48) correctly fixes the mid-PR PORT=0 bug: with an OS-ephemeral port it logs the real bound port, not 0. Falls back to port only if local_addr() fails. ✓
  - preferred_port: malformed PORT warns and falls back to 3000 rather than silently masquerading — matches the PR's stated intent. ✓
  - Bind-before-build + single listener: bind moved ahead of app construction so collisions fail fast; the old redundant second TcpListener::bind + .expect() is removed; the one listener
  flows straight into axum::serve. No stale addr/:3000 references remain. ✓
  - Loud success log (Pick web UI available at http://localhost:{port}) is the actual fix for the silent-wrong-app confusion. ✓

  Two minor non-blocking edge notes

  1. saturating_add degrades near the u16 ceiling. Confirmed empirically: PORT=65535 makes all 20 attempts target the same port 65535 (1 distinct); PORT=65530 yields only 6 distinct. So at
  the extreme top of the range the "scan 20 ports" intent collapses to "retry the same port." It's safe (saturating avoids an overflow panic — the right call), just less useful there. A
  realistic default (3000) scans 20 distinct ports correctly. Optional: cap the start so preferred + MAX_ATTEMPTS can't exceed 65535, or log when saturation kicks in. Very low priority.
  2. No unit test for preferred_port/bind_with_fallback. But apps/web is a binary-only crate (no tests/, no test modules, functions not in a lib), so this is consistent with the crate's
  convention rather than a dropped test — and the author did a thorough live behavioral test (all 4 scenarios incl. the real nexus-web collision + PORT=0 after the fix). Acceptable; if
  these functions ever move to a lib, they'd be trivially unit-testable (pure logic + a :0-bind test).

  Scope-narrowing claim — sound

  The PR's decision to skip the 3030/LIVEVIEW_PORT server is justified: I've seen in prior reviews that the headless/liveview server binds a PID-based Unix socket, not a TCP port, so
  there's no collision to fix there. apps/web was indeed the only fixed TCP bind. Consistent with what I know of the codebase.

  init-dev relevance

  None — this is the desktop/local web app (pentest-web on :3000); init-dev deploys the headless connector, which doesn't use this binary. No consumer impact.

jtomek-strike48:
Addressed the `saturating_add` edge note (commit `1a552f9`): switched to `checked_add` + `break`, so near the u16 ceiling the scan stops instead of re-probing the same saturated port. Each attempt is now a distinct port; no change for realistic defaults (3000 still scans 20 distinct ports). Verified: `PORT=65534` binds cleanly, default->3001 and `PORT=3007` regressions intact; clippy `-D warnings` + fmt clean.

The other note (no unit test for the port helpers) is left as-is per the review: `apps/web` is a binary-only crate with no test module, so this matches the crate's convention, and the logic was behaviorally verified across all scenarios.

Remaining before merge: this branch will go green once the flaky-test fix (#272) lands on main and this rebases onto it - the current red `Test` check is the pre-existing #269 webwright flake, unrelated to this diff.

PR #270, opened 2026-07-18
feat(#261): curated tool category taxonomy + re-tag all wrapped tools

jtomek-strike48:
## Summary

Child A of epic #250. Expands the tool taxonomy so the Settings Tools grid (from #259) and the read-only Tools page group tools into meaningful, BlackArch-aligned categories instead of dumping ~80 tools into "Other".

Closes #261.

### Enum (`crates/core/src/tools.rs`)
Added 6 curated categories - `WebDiscovery`, `Proxy`, `Exploitation`, `Sniffing`, `Crypto`, `Utilities` - alongside the existing Network, Web, ActiveDirectory, Credentials, PostExploit, Wireless, Recon, Forensics.

### Catalog (`crates/tools/src/catalog.rs`)
- **`category_key()`** - one exhaustive match from `ToolCategory` to its stable serde key, shared by the catalog and the overview. Replaces a hand-written match in `overview_category` that had already gone stale, and makes a future missing key a **compile error**.
- **Hide runtime deps** - `python3` / `playwright` are filtered from the catalog UI (`HIDDEN_CATALOG_BINARIES`). They stay real dependencies for webwright's installer; they're just not operator-facing "tools" (per maintainer decision).

### Re-tagging (~80 files)
Every `ExternalDependency::new(...)` now declares a `.category(...)`. Also corrected 3 that predated the taxonomy: burp/zap `Web` -> `Proxy`, metasploit `PostExploit` -> `Exploitation`. Final distribution (no tool left in "Other"):

| Category | # | Category | # |
|---|---|---|---|
| Web (scan/exploit) | 14 | Credentials | 5 |
| Recon & OSINT | 14 | Sniffing | 3 |
| Active Directory | 14 | Proxy / PostExploit / Exploitation / Utilities | 2 each |
| Network | 13 | Wireless / Forensics / Crypto | 1 each |
| Web Discovery | 12 | | |

### UI (`crates/ui/src/components/tool_category.rs`)
Friendly labels ("Recon & OSINT", "Web - Content Discovery", "TLS / Crypto", ...) and icons for every new key, shared by both surfaces.

## How the ~80 re-taggings were done
The mechanical `.category(...)` insertion + import handling was fanned out across 6 parallel subagents over disjoint file sets, each self-verifying with `cargo check`. I then verified the whole set centrally: 0 files untagged, category distribution matches the intended mapping, and diffs contain only category/import/rustfmt-reflow changes (no logic/param edits) - spot-checked several files.

## Verification
[...]

PR #272, opened 2026-07-18
fix(#269): serialize webwright evidence tests to stop CI flakiness

### LINKED ISSUE #269: Flaky tests: webwright_integration evidence tests fail under full-workspace parallelism (shared evidence buffer)
## Problem

Two tests in `crates/tools/tests/webwright_integration.rs` fail nondeterministically when the whole workspace test suite runs in parallel:

- `evidence_ingestion_produces_typed_nodes` (asserts 8 nodes, sometimes gets 3)
- `findings_ingestion_maps_all_severity_levels` (asserts 5, sometimes gets 10 - exactly double)

```
assertion `left == right` failed: Should produce 8 evidence nodes, got 3
assertion `left == right` failed   left: 10  right: 5
```

## Root cause (shared global state)

Both tests call `drain_pending_evidence()`, which drains a **process-global** pending-evidence buffer. When other tests that also push evidence run concurrently (in the same or another test binary sharing the global), counts bleed across tests - hence "got 3" (drained early by another test) and "got 10" (another test's 5 nodes still queued, doubling the count). The doubling/halving pattern is the classic shared-mutable-global race.

## Evidence it is pre-existing and isolation-related (A/B verified)

- **Single-threaded**: `cargo test --test webwright_integration -- --test-threads=1 <both tests>` -> **pass**.
- **That binary alone, parallel, 3x**: all **pass** (7/7 each run).
- **Full workspace parallel** (`cargo test --workspace --features pentest-platform/desktop-pcap`): the 2 tests **FAIL**.
- **Same full-workspace run on clean `main`** (no local changes): **also FAILS identically** (`5 passed; 2 failed`).

So this reproduces on `main` and is unrelated to any specific feature branch. It surfaced while validating an unrelated change (#261 tool-category re-tagging), whose diff does not touch evidence ingestion.

## Suggested fix

- Give the pending-evidence buffer test isolation: either a per-test instance/handle instead of a process global, or serialize the affected tests (e.g. a shared `#[serial]` guard via the `serial_test` crate, or gate them behind `#[ignore]` + a dedicated single-threaded CI step like the existing evidence-buffer tests in `evidence_producer.rs`).
- Audit other `drain_pending_evidence()` callers in the test suite for the same coupling.

## Recommended agent
test-engineer (test isolation) or rust-reviewer.

jtomek-strike48:
## Summary

Two tests in `crates/tools/tests/webwright_integration.rs` -  `evidence_ingestion_produces_typed_nodes` and `findings_ingestion_maps_all_severity_levels` - fail nondeterministically under the full-workspace parallel test run, red-lighting unrelated PRs (it just failed #267's CI; would hit #270 and every future PR).

Closes #269.

## Root cause

Both tests drain the **process-global** `PENDING_EVIDENCE` buffer (`evidence_producer.rs`), ingest a fixed set of nodes, then assert on the drained count. When they run concurrently with each other or any other test that pushes evidence, nodes bleed across tests - so the count comes out wrong (observed: got 1, 3, or 10 vs expected 8/5). The varying number is the signature of a shared-state race, not a logic bug. A/B-verified: fails identically on clean `main`; passes single-threaded.

## Fix

Add a test-local `EVIDENCE_LOCK: Mutex<()>` and hold it across each test's full drain-ingest-drain sequence, so each test sees only its own evidence.

- **Why not `#[ignore]`** (the pattern the 3 evidence-buffer unit tests use): CI has **no `--ignored` step**, so ignoring these would silently drop their coverage. Serializing keeps them running in normal CI.
- **Poison-tolerant** (`into_inner()` on a poisoned lock) so a panic in one guarded test doesn't cascade-fail the others.
- Only the 2 non-ignored buffer tests are guarded. The two sandbox-gated `#[ignore]`'d tests that also touch the buffer `.await` across their buffer ops (holding a std Mutex across await = deadlock/`await_holding_lock` risk) and don't run in normal CI, so they're intentionally left unguarded.

## Verification

- `cargo test --workspace --features pentest-platform/desktop-pcap`: **exit 0, 0 failures** (previously 2 failed) - this is the exact CI condition that failed.
- `webwright_integration` binary: green across 3 repeated parallel runs.
- `cargo clippy --workspace --tests -- -D warnings`: clean (no `await_holding_lock`).
- `cargo fmt --all --check`: clean.

## Impact on other PRs

Once merged, #267 and #270 rebase onto this and go green deterministically. This removes the flake for all future PRs.

## Test plan
- [x] Full workspace test passes (exit 0).
- [x] webwright_integration passes across repeated parallel runs.
- [x] clippy `-D warnings` clean (no await-holding-lock).
- [ ] CI Test job passes on this PR.

### DIFF
--- a/crates/tools/tests/webwright_integration.rs
+++ b/crates/tools/tests/webwright_integration.rs
@@ -14,6 +14,20 @@ use pentest_tools::webwright::evidence::{ingest_webwright_evidence, ingest_webwr
 use pentest_tools::webwright::workspace::WebwrightWorkspace;
 use pentest_tools::WebwrightTool;
 use serde_json::json;
+use std::sync::{Mutex, MutexGuard};
+
+/// Serializes the tests that touch the process-global `PENDING_EVIDENCE`
+/// buffer (via `push_evidence`/`drain_pending_evidence`). Without this, tests
+/// that drain-then-assert-count race each other: one test's pushed nodes land
+/// in another's drain, so counts come out wrong nondeterministically (the
+/// classic doubling/halving). Holding this lock for the whole
+/// drain-ingest-drain sequence makes each such test see only its own evidence.
+/// Poison-tolerant: a panic in one test must not cascade-fail the others.
+static EVIDENCE_LOCK: Mutex<()> = Mutex::new(());
+
+fn lock_evidence() -> MutexGuard<'static, ()> {
+    EVIDENCE_LOCK.lock().unwrap_or_else(|e| e.into_inner())
+}
 
 /// Test that execute mode runs a Python script and captures output.
 #[tokio::test]
@@ -396,6 +410,7 @@ async fn workspace_creates_and_collects_artifacts() {
 fn evidence_ingestion_produces_typed_nodes() {
     use pentest_core::provenance::ProbeCommand;
 
+    let _guard = lock_evidence();
     let _ = drain_pending_evidence();
 
     let artifacts = json!({
@@ -488,6 +503,7 @@ fn findings_ingestion_maps_all_severity_levels() {
     use pentest_core::export::Severity;
     use pentest_core::provenance::ProbeCommand;
 
+    let _guard = lock_evidence();
     let _ = drain_pending_evidence();
 
     let findings = json!([


cchao-devo (approved):
Review: Pick PR #272 — fix(#269): serialize webwright evidence tests to stop CI flakiness
  
  Verdict: ✅ Approve. Correct, minimal, well-reasoned test-isolation fix. I reproduced the flake on base and confirmed the fix eliminates it. No blockers. This is the fix for the exact
  flakiness I'd flagged as pre-existing in my #246 and #270 reviews — good to see it closed at the source.

  Lenses run: mechanical gates (fmt + clippy --tests -D warnings) + manual correctness trace (sync-vs-async lock safety, race-partner enumeration) + empirical A/B reproduction of the
  flake. Verified in pick_08/pick; pick_07 untouched, no cluster.

  Empirical proof (the key evidence)

  - Base (pre-fix), 2 tests run concurrently (--test-threads=8): flaked 1/10 — both tests failed together (the shared-buffer race).
  - Fix (pr-272), same condition: 0/15 failures.
  - Plus 6/6 green on the full binary in default parallel mode. The EVIDENCE_LOCK serialization demonstrably closes the race.

  Correctness trace — the subtle parts hold

  - Std Mutex is safe here: both guarded tests are plain sync #[test] (not #[tokio::test]), and there's no .await between lock_evidence() and the drain/assert. So no await_holding_lock /
  cross-await-deadlock risk. Confirmed by clippy --tests -D warnings passing clean.
  - Race partners fully covered: in this binary, exactly 4 tests touch PENDING_EVIDENCE — the 2 non-ignored sync tests (now both hold the shared lock → serialized against each other) and 2
  #[ignore]'d async tests (skipped in CI; require bwrap sandbox; they .await across buffer ops so a std Mutex would be unsafe → correctly left unguarded). Since the 2 CI-running tests are
  the only concurrent buffer users in this binary, guarding both closes the race.
  - Cross-binary is a non-issue: other pushers (crates/ui/tests/evidence_pipeline.rs, session.rs, evidence_producer.rs unit tests) run in separate test processes, each with its own
  PENDING_EVIDENCE static — no cross-binary bleed possible.
  - Poison-tolerant (unwrap_or_else(|e| e.into_inner())) so a panic in one guarded test doesn't cascade-fail the other — sensible for test code.

  Judgment calls — all correct

  - Mutex over #[ignore]: right call — CI has no --ignored step, so ignoring these would silently drop their coverage (exactly the gap I'd worry about). Serializing keeps them running.
  - The comment accurately explains the doubling/halving signature and the guarded-scope reasoning.

  Note

  The deeper design smell — a process-global mutable PENDING_EVIDENCE buffer that tests must coordinate around — remains (this PR pragmatically serializes rather than refactoring to
  injectable state). That's the right scope for a CI-unblock fix; a longer-term "thread the buffer through instead of a global" refactor would be a separate change, not warranted here.

  init-dev relevance

  None — test-only change.

PR #275, opened 2026-07-19
feat(#265): uninstall tools from the Settings panel (trash-can + confirm)

jtomek-strike48:
## Summary

Child E of epic #250. Adds an operator-driven **uninstall** for installed tools — the trash-can + confirm-dialog the issue specced — completing the install/uninstall loop in the redesigned Tools panel.

Closes #265.

### Backend (`crates/tools`)
- `pacman::uninstall` -> `pacman -Rns --noconfirm <pkg>` (package + now-unneeded deps + config files; `-s` preserves deps other installed packages still need). "target not found" is treated as success (idempotent).
- `catalog::is_uninstallable` gates by mode: `Pacman` in-sandbox, `AptHost` in native; **never** `Manual` (we didn't install it) or `Custom` (no standard teardown). `uninstall_entry` / `uninstall_by_binary` mirror the install path; apt-host removal via `apt-get remove -y`.
- `CatalogItem.uninstallable` flag so the UI knows when to show the trash-can. Reuses `validate_package_name` on both removal paths.

### UI (`crates/ui`)
- Trash-can top-right of installed cards (only when `uninstallable`), opening a confirm dialog that **names the package** and **warns about sibling tools that share the dependency** (the `used_by` blast radius). Destructive-red confirm button; per-card "Removing..." state; one-op-at-a-time guard; overlay-click and Cancel both dismiss.

### Security review (destructive commands, per the issue + CLAUDE.md)
Ran a security-reviewer pass focused on the `pacman -Rns` / `apt-get remove` construction. Verdict: **safe to merge** — package names come from hardcoded schemas (not user input), pass `validate_package_name` (rejects shell metacharacters), commands use argv (`execute_command(cmd, &[args])`) not a shell string, and Manual/Custom tools are provably unremovable (guarded in `uninstall_entry`, verified by test).

Two findings actioned:
- **Applied now**: added an explicit leading-dash reject to `validate_package_name`, so a name like `-Syu` / `--purge` can't be parsed as a pacman/apt-get flag (argument injection). Not currently reachable, but defense-in-depth on the widened `apt-get remove` surface (where `--purge` would escalate to config deletion). Test extended to cover it.
- **Deferred (pre-existing, filed as #274)**: the sandbox command path builds a shell string and escapes only the args, not the command name. Not exploitable today (static `cmd`, validated names) and affects the install path equally, so out of scope here.

## Verification
[...]

Issue #263, opened 2026-07-17
Child C (#250): live BlackArch full tier from pacman -Sg / -Si (sandbox-gated)


jtomek-strike48:
Child of #250. The headline capability: browse/install all ~3000 BlackArch tools, never hardcoded.

## Problem

We hardcode package names and a flat wrapped-tool list. The maintainer wants the full BlackArch suite (we run BlackArch), which grows over time. Hardcoding cannot track ~3000 rolling packages.

## Scope

- Read categories live from `pacman -Sg` (~50 BlackArch groups) and per-package Name/Description from `pacman -Si`, inside the sandbox. Never hardcode names (avoids the name-drift bug class, e.g. python-bloodhound -> bloodhound-ce-python).
- Render under a second-tier **"All BlackArch"** section, below the curated **Featured** section.
- **Sandbox-gated**: when no Linux sandbox is provisioned (or DB unsynced), show a clear empty-state ("Provision the sandbox to browse the full BlackArch catalog"), not an error.
- **Badge wrapped tools**: packages that have a Rust wrapper (agent-orchestrable, scope+provenance) are visually distinguished from raw packages (install-only, reachable via execute_command).
- Cache the group/package listing per session; refresh on demand.

## Dependencies / risks

- Gated by pacman-in-sandbox reliability: **#248** (openat2 / DBs unreadable) and **#245** (-Sy sync). May need those green first.
- Cross-platform: works wherever a Linux sandbox exists (Linux proot/bwrap, Windows WSL2, macOS Docker) - not Linux-host-only.

## Acceptance criteria

- [ ] Categories derived from `pacman -Sg`; package names/descriptions from the package DB - none hardcoded.
- [ ] Full tier renders only when the sandbox is up; clear empty-state otherwise.
- [ ] Wrapped (integrated) tools are badged distinctly from raw packages.
- [ ] Listing is cached and refreshable; no blocking the UI thread.

## Recommended agent
api-architect / devops-engineer (sandbox + pacman) + rust-reviewer.

Depends on Child A + B (#250); gated by #248, #245.

Issue #252, opened 2026-07-17
Productionize openat2-capable proot: CI build + checksummed hosted artifact


jtomek-strike48:
## Context

#248 root cause: bundled proot v5.3.0 cannot translate `openat2`, so libalpm can't read pacman DBs. Verified fix: an openat2-capable proot (built from termux/proot). The fix is proven and installed **locally on one dev host**, but is **not durable** — `download_proot` (`crates/platform/src/desktop/sandbox/proot.rs`) still points at the broken upstream `proot-me v5.3.0` and re-fetches it on any fresh install / cleared binary / new host.

## Work

1. **CI workflow** (`.github/workflows`) to build the openat2-capable proot **static** for `x86_64` + `aarch64`:
   - Source: termux/proot at a pinned commit (it has the openat2 -> openat rewrite; upstream proot-me does not).
   - Build glibc-static in Debian (musl chokes on `struct rlimit64`).
   - Two source patches: strip the lld-only `--rosegment` linker flag; provide a minimal `linux/ashmem.h` so the Android-only `ashmem_memfd` extension compiles (do NOT delete the object — `cli/proot.c` references `ashmem_memfd_callback`).
   - Full recipe verified and recorded during the spike.
2. **Host the artifact** on a `Strike48-public/pick` GitHub release (decided: public release, matches how proot is fetched today).
3. **Update the downloader**: point `PROOT_DOWNLOAD_URLS` at the release **and add SHA-256 verification** — the current downloader has NONE (plain `reqwest::get` -> write -> chmod 0755), which is a supply-chain gap regardless.
4. **Fold in `XferCommand = /usr/bin/curl -L -f -o %o %u`** in `rootfs.rs::fix_pacman_config` — pacman 7's internal parallel downloader is a silent no-op under proot; curl makes DB downloads deterministic.
5. **CI smoke test**: run the `openat2` probe (syscall 437 on a rootfs file must return a fd, not ENOENT) so a proot regression can never ship. This is also an early-warning for future new-syscall gaps as BlackArch's rolling libalpm evolves.
6. **Re-test #238** interactive PTY shell against the new binary (this branch swaps the proot it selects).

## Acceptance criteria

- [ ] CI produces checksummed static proot binaries for x86_64 + aarch64.
- [ ] `download_proot` fetches from the Strike48 release and verifies SHA-256, failing closed on mismatch.
- [ ] Fresh sandbox provisioning installs tools end-to-end with no manual binary swap.
[...]

PR #285, opened 2026-07-20
ci(#252): workflow to build openat2-capable static proot

### LINKED ISSUE #248: proot sandbox: pacman -Sy succeeds but databases unreadable -> all tool installs fail (target not found)
## Summary

Inside the proot BlackArch sandbox, `pacman -Sy` reports success but the synced package databases are never usable: `pacman -Sl` / `-Ss` / `-S <pkg>` all behave as if the databases are empty or missing. As a result, **no tool can be installed via pacman** even though the sandbox itself now works (proot selected, commands run). Installs fail with `error: target not found: <pkg>`.

This is downstream of #238/#239 (sandbox selection) and #245/#246 (which added `pacman -Sy` before install). The `-Sy` now runs and exits 0 — but the DB it produces is not readable, so the install still fails. Different, deeper layer: rootfs pacman DB provisioning under proot.

## Reproduction (on the affected host)

Run any tool install in Proot mode, e.g. BloodHound:

```
Tool execution error: Failed to install python-bloodhound: error: target not found: python-bloodhound
```

Directly in the sandbox:

```
$ proot -0 -r <rootfs> ... /bin/bash -lc 'pacman -Sy; pacman -Sl core | wc -l'
:: Synchronizing package databases...
 core downloading...
 extra downloading...
 blackarch downloading...
(exit 0)
0        <-- pacman sees ZERO packages
```

## Diagnostics gathered

- **The .db files exist on disk and are valid**: `core.db` (129 KB), `extra.db` (~8.7 MB), `blackarch.db` (~2.1 MB), all valid gzip tarballs containing real package `desc` entries (`acl-2.4.0-1/`, `0ad-...`, `0d1n-...`).
- **pacman claims the DBs are unreadable** even though they exist — `--debug` shows, for a file `ls`/`cat` can read fine:
  ```
  debug: registering sync database 'core'
  debug: "/var/lib/pacman/sync/core.db" is not readable: No such file or directory
  ```
- **`-Sy` exits 0 but writes nothing usable**: after `pacman -Syy` (exit 0) the sync dir mtimes are unchanged and no `.part` files remain.
- **Not a proot fops/path issue**: `cat /var/lib/pacman/sync/core.db` inside proot succeeds (reads 129057 bytes); plain `mv` in that dir works.
- **Not the pre-existing provisioned files**: syncing to a *fresh* empty `--dbpath /tmp/freshdb` reproduces identically (downloads "succeed", 0 packages readable).
- **Env is otherwise correct**: `uname -m` = x86_64, DBPath `/var/lib/pacman/`, repos core/extra/blackarch all registered, `Architecture = auto`, pacman v7.1.0 / libalpm v16.0.1, `ALPM_DB_VERSION` = 9.

So: the download step under proot produces no usable database, from a clean slate, despite exit 0. This looks like a proot + pacman/libalpm download-and-commit interaction (the downloaded DB is not landing where/how libalpm can then open it), not a stale-file or path-translation problem.

## Leads for whoever picks this up

- `RootfsManager::run_in_rootfs` (`crates/platform/src/desktop/sandbox/rootfs.rs`) tries `arch-chroot -> chroot -> proot` in order. On this host `arch-chroot` is absent and `chroot` needs root, so provisioning likely fell through to proot. Databases created/synced via one access path may not be consumable via another — worth confirming which path actually ran at provisioning vs. at install time.
- Compare a `pacman -Sy` under proot vs. under bwrap/chroot on a host where those work, to isolate whether this is proot-specific.
- Check whether libalpm's download-to-`.part`-then-commit step completes under proot (the `.part` tempfile + rename/fsync path), and whether `DownloadUser`/sandbox options in `pacman.conf` interact with proot here.
- Consider shipping pre-synced databases with the rootfs, or performing the DB sync at provisioning time through a path proven to work, so install-time `-Sy` is not required under proot.

## Impact

Blocks all pacman-based tool installation in the proot sandbox on affected hosts (e.g. Ubuntu 24.04 desktops where bwrap is unavailable and proot is the selected backend). The sandbox shell works; only package installation is affected.

## Acceptance criteria

- [ ] In the proot sandbox, `pacman -Sy` produces databases that `pacman -Sl` can read (non-zero package count).
- [ ] Installing a tool (e.g. bloodhound) into the proot sandbox succeeds end-to-end.
- [ ] Works from a clean rootfs (no pre-provisioned sync DBs required), or the rootfs ships pre-synced DBs that proot can read.

## Context

Found during #238/#239/#246 live verification once the sandbox backend was correctly selected (proot on an AppArmor-restricted Ubuntu 24.04 host). Separate subsystem from backend selection and from the `-Sy` install-arg fix.

jtomek-strike48:
## Summary

First step of #252, which productionizes the fix for **#248** (the sandbox bug blocking all pacman tool installs — and, downstream, #263's live BlackArch tier and #275's uninstall end-to-end test).

**#248 root cause** (already diagnosed in #252): the bundled `proot-me v5.3.0` can't translate the `openat2` syscall, so libalpm (pacman 7) can't read the sync DBs under the sandbox — every install fails `target not found` even though the `.db` files are on disk. The verified fix is a static, **openat2-capable** proot built from `termux/proot`.

This PR adds the workflow that **builds and hosts that binary**, so the downloader can later point at it.

## What's here
- **`.github/workflows/build-proot.yml`** — matrix build (x86_64 + aarch64) in a `debian:bookworm` container (aarch64 via QEMU). x86_64 runs an **openat2 smoke test**: a probe compiled and run *under the freshly built proot* must open a file via syscall 437, or the build fails — so a proot that reintroduces #248 can never ship. Per-artifact SHA-256 recorded. A `publish` job (gated on a `proot-openat2-v*` tag or a `release_tag` input) uploads the binaries + checksums to a GitHub release.
- **`.github/proot-build/build.sh`** — the recipe recorded during the #248 spike (glibc-static; strip the lld-only `--rosegment` flag; stub `linux/ashmem.h`), as a committed script bind-mounted into the container.
- **`.github/proot-build/{ashmem.h, openat2_probe.c}`** — the ashmem stub + smoke-test probe as committed files.

### A deliberate robustness choice
The build recipe is a **committed script + support files**, not a nested heredoc inside a docker `bash -c '...'` inline block. Multi-layer YAML+shell quoting around nested heredocs silently corrupts the payload — I hit exactly that while drafting this, so I moved the C header and probe out to real files. (Verified: `openat2_probe.c` compiles/links, `build.sh` passes `bash -n`, the workflow is valid YAML with zero heredocs.)

## What this PR does NOT do (honest scope)
- It does not repoint `download_proot` or add SHA-256 verification yet (that's #252 steps 3-4) — those are done **after** a release exists with a known checksum, so they can be verified rather than shipped against a placeholder.
- I **cannot** run the actual cross-compile or publish the release from here (needs the CI runner + a release). This PR is the workflow that lets CI do it.

## Action required to make it usable
1. ~~Pin `TERMUX_PROOT_REF`~~ **DONE** — pinned to termux/proot @ `6c09638` (full SHA), verified at pin time to carry the openat2 handling (references `openat2` across 13 files incl. the `src/syscall/enter.c` rewrite + per-arch `sysnums-*.h` tables). If the spike used a different commit, bump + re-verify.
2. Merge, then run the workflow (dispatch with a `release_tag`, or push a `proot-openat2-v1` tag) so it builds + publishes.
3. Follow-up PR: repoint `download_proot` at the release + SHA-256 verify + `XferCommand=curl`.

## Test plan
- [x] `TERMUX_PROOT_REF` pinned to a verified openat2-capable commit (termux/proot @ `6c09638`).
- [ ] Workflow dispatched; x86_64 + aarch64 build green.
- [ ] openat2 smoke test passes (proot translates syscall 437).
- [ ] Release published with both binaries + `.sha256` files.
- [ ] (follow-up) downloader repointed + checksum-verified; fresh sandbox installs a tool end-to-end (closes #248).

Part of #252. I cannot validate the build end-to-end locally (no CI runner / static libc on this host); the workflow YAML, the build script (`bash -n`), and the probe (compiles) are validated as far as possible without a run.

### DIFF
--- a/.github/proot-build/ashmem.h
+++ b/.github/proot-build/ashmem.h
@@ -1,0 +1,20 @@
+/* Minimal linux/ashmem.h stub for building termux/proot on glibc/Debian.
+ *
+ * termux/proot's Android-only `ashmem_memfd` extension #includes
+ * <linux/ashmem.h>, which does not exist in mainline glibc headers. We stub it
+ * (rather than delete the object) because cli/proot.c references
+ * ashmem_memfd_callback. See .github/workflows/build-proot.yml and #248/#252.
+ */
+#ifndef _LINUX_ASHMEM_H
+#define _LINUX_ASHMEM_H
+
+#include <linux/limits.h>
+#include <linux/ioctl.h>
+
+#define ASHMEM_NAME_LEN 256
+#define ASHMEM_NAME_DEF "dev/ashmem"
+
+#define ASHMEM_SET_NAME _IOW(0x77, 1, char[ASHMEM_NAME_LEN])
+#define ASHMEM_SET_SIZE _IOW(0x77, 3, size_t)
+
+#endif /* _LINUX_ASHMEM_H */
--- a/.github/proot-build/build.sh
+++ b/.github/proot-build/build.sh
@@ -1,0 +1,39 @@
+#!/usr/bin/env bash
+# Build a static, openat2-capable proot from termux/proot (#248/#252).
+#
+# Runs INSIDE a debian:bookworm container (see build-proot.yml), with the repo
+# bind-mounted at /work. Reads $TERMUX_PROOT_REF and $ARCH from the env.
+# Recipe verified A/B during the #248 spike:
+#   - glibc-static (musl fails on `struct rlimit64`)
+#   - strip the lld-only `--rosegment` linker flag (GNU ld rejects it)
+#   - stub <linux/ashmem.h> so the Android-only ashmem_memfd extension compiles
+#     (do NOT delete the object — cli/proot.c references ashmem_memfd_callback)
+set -euo pipefail
+
+: "${TERMUX_PROOT_REF:?TERMUX_PROOT_REF must be set}"
+: "${ARCH:?ARCH must be set}"
+
+apt-get update
+apt-get install -y --no-install-recommends \
+  git ca-certificates build-essential libc6-dev \
+  pkg-config uthash-dev libtalloc-dev python3 gzip file
+
+git clone https://github.com/termux/proot /src
+cd /src
+git checkout "$TERMUX_PROOT_REF"
+
+# Patch 1: strip the lld-only --rosegment flag (GNU ld rejects it).
+if grep -rl -- "--rosegment" . >/dev/null 2>&1; then
+  grep -rl -- "--rosegment" . | xargs sed -i "s/-Wl,--rosegment//g"
+fi
+
+# Patch 2: install the committed linux/ashmem.h stub.
+mkdir -p /usr/include/linux
+cp /work/.github/proot-build/ashmem.h /usr/include/linux/ashmem.h
+
+# Static glibc build.
+make -C src LDFLAGS="-static" V=1
+/src/src/proot --version
+
+cp /src/src/proot "/work/proot-${ARCH}"
+file "/work/proot-${ARCH}"
--- a/.github/proot-build/openat2_probe.c
+++ b/.github/proot-build/openat2_probe.c
@@ -1,0 +1,30 @@
+/* openat2 smoke test for the freshly built proot (#248/#252).
+ *
+ * #248: the old proot could not translate the openat2 syscall, so libalpm's
+ * pacman-DB reads failed under the sandbox. This probe opens a file via
+ * openat2 (syscall 437). Run UNDER the new proot it must return a valid fd; if
+ * proot cannot translate openat2 it returns an error and the CI build fails,
+ * so a regression can never ship. Compiled and invoked by
+ * .github/workflows/build-proot.yml.
+ */
+#define _GNU_SOURCE
+#include <fcntl.h>
+#include <stdio.h>
+#include <string.h>
+#include <unistd.h>
+#include <sys/syscall.h>
+#include <linux/openat2.h>
+
+int main(void) {
+    struct open_how how;
+    memset(&how, 0, sizeof(how));
+    how.flags = O_RDONLY;
+
+    long fd = syscall(SYS_openat2, AT_FDCWD, "/etc/hostname", &how, sizeof(how));
+    if (fd < 0) {
+        perror("openat2");
+        return 2;
+    }
+    printf("openat2 ok, fd=%ld\n", fd);
+    return 0;
+}
--- a/.github/workflows/build-proot.yml
+++ b/.github/workflows/build-proot.yml
@@ -1,0 +1,159 @@
+# Build an openat2-capable static `proot` for the BlackArch sandbox.
+#
+# Why this exists (#248 / #252): the upstream proot-me v5.3.0 binary the
+# connector downloads cannot translate the `openat2` syscall. Modern libalpm
+# (pacman 7) opens the sync databases with `openat2`, so under that proot every
+# DB read fails with a bogus "not readable: No such file or directory" and
+# ALL pacman-based tool installs fail with "target not found" — even though the
+# .db files are on disk and readable via cat.
+#
+# termux/proot carries an `openat2 -> openat` rewrite that upstream proot-me
+# lacks. Building that statically (glibc, not musl — musl chokes on
+# `struct rlimit64`) produces a proot whose DB reads work, which fixes #248.
+#
+# This workflow builds that binary for x86_64 + aarch64, verifies it can
+# translate openat2 (smoke test), checksums it, and — on a tagged/dispatched
+# run — publishes it to a GitHub release the connector's downloader points at.
+#
+# Manual, not push-triggered: building proot is infrastructure, run
+# deliberately (workflow_dispatch) or when a `proot-openat2-vN` tag is pushed.
+name: Build proot
+
+on:
+  workflow_dispatch:
+    inputs:
+      release_tag:
+        description: "Release tag to publish under (e.g. proot-openat2-v1). Blank = build + smoke-test only, no publish."
+        required: false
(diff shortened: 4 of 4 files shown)


cchao-devo (approved):
Review: Pick PR #285 — ci(#252): workflow to build openat2-capable static proot

  Verdict: ✅ Approve (with one substantive smoke-test note the author should weigh). Clean, well-documented infra PR that unblocks #248. Locally-runnable gates pass; the parts that need a
  CI runner are honestly scoped as un-runnable-here. No blockers — the note below is a "make the guarantee real" improvement, not a correctness defect in what ships.

  Lenses run: shellcheck + shfmt + bash -n (build.sh), YAML validity, C-probe API review, workflow-logic trace, smoke-test soundness analysis, action-pinning convention check, supply-chain
  (pinned ref). Verified in pick_08/pick; no product code, no cluster.

  Gates

  - shellcheck -s bash build.sh clean · shfmt -i 2 -ci -bn clean · bash -n clean (matches repo bash convention)
  - Workflow YAML valid (ruby parse)
  - openat2_probe.c uses the correct ABI (SYS_openat2, struct open_how, linux/openat2.h, how.flags) — can't compile on macOS (linux/openat2.h is Linux-only; that's expected, CI builds on
  Linux)
  - Not runnable locally (by design): the cross-compile/publish need a CI runner + static glibc — the author is upfront about this.

  What's well done

  - Robustness choice is right: committed build.sh + ashmem.h + openat2_probe.c bind-mounted into the container, instead of nested heredocs in a docker bash -c. The author hit
  heredoc-quoting corruption and moved to real files — exactly the correct call.
  - Supply-chain hygiene: TERMUX_PROOT_REF pinned by full SHA (6c09638…) with a comment on why (reproducibility + force-push resistance) and what was verified at pin time. Good.
  - Smoke-test gating is sound: set -euo pipefail + running the probe bare means a non-zero exit (probe returns 2 on openat2 failure) fails the step → fails the build. A hard openat2
  failure can't ship silently.
  - Build recipe captures the real #248-spike findings (glibc-static not musl, strip lld-only --rosegment, stub linux/ashmem.h without deleting the object since cli/proot.c refs
  ashmem_memfd_callback). Idempotent patch guards (grep -rl … >/dev/null before sed).
  - actions/checkout@v5 and softprops/action-gh-release@v1 both match existing repo convention (checkout@v5 used 24× elsewhere; gh-release@v1 used 2×).

  Substantive note (worth the author's judgment — not a blocker)

  The smoke test may not catch the exact #248 signature, because it runs against an identity rootfs (-r /). #248 isn't "openat2 fails outright" — per the workflow's own comment, the broken
  proot executes openat2 but returns a bogus ENOENT because it doesn't rewrite the guest path to the rootfs path (a path-translation bug). The probe opens /etc/hostname under proot -0 -r
  /, where guest /etc/hostname maps to the host's real /etc/hostname regardless of whether translation happens — so a proot with the #248 path-translation bug could still false-pass this
  test. The bug bites in production because the sandbox rootfs is a BlackArch tree at a non-/ path.

  To make the smoke test actually prove the #248 fix, open a file under a non-identity rootfs: e.g. mkdir -p /tmp/rootfs/etc && echo hi > /tmp/rootfs/etc/hostname, then ./proot-x86_64 -0
  -r /tmp/rootfs /tmp/openat2_probe (probe still opens /etc/hostname). Under a broken proot the path wouldn't rewrite to /tmp/rootfs/etc/hostname → ENOENT → build fails, which is the real
  #248 signature. As written, the test guarantees "openat2 is translated at all," not "openat2 path-translation works under a real rootfs." Given #252's follow-up will do an actual
  end-to-end tool install (the ultimate test), this is a reasonable interim — but the one-line rootfs change would make the committed guarantee match its comment ("a regression here must
  fail the build").

  Minor
  
  - aarch64 gets no smoke test (QEMU syscall-probing unreliable) — reasonably documented; the x86_64 gate + end-to-end follow-up cover it.
  - permissions: contents: write is correctly scoped (needed for the release publish) and duplicated at the publish job level — fine.

  init-dev relevance
  
  Indirect but real: #248's sandbox-install bug is why init-dev deploys Pick with pick.disableSandbox=true (nmap file-caps need root; proot OOMs/fails). If #252 lands a working openat2
  proot end-to-end, that init-dev workaround could eventually be revisited — but nothing changes for init-dev from this PR (it only builds/hosts the binary; the downloader isn't repointed
  yet).

jtomek-strike48:
Good catch on the smoke test — you're right, and it's fixed in `9a4b5c7`.

The identity-rootfs (`-r /`) probe would false-pass the real #248 bug: since #248 is a path-*translation* failure (openat2 runs but the guest path isn't rewritten to the rootfs path), and under `-r /` guest `/etc/hostname` maps to the host's real file regardless of translation, a broken proot would still open it and pass.

Now the probe runs under a **non-identity rootfs**: `mkdir /tmp/rootfs/etc`, write a marker `hostname` there, then `proot -0 -r /tmp/rootfs`. The probe opens `/etc/hostname` — a correct proot rewrites that to `/tmp/rootfs/etc/hostname` (fd ok); a #248-broken proot leaves it untranslated → ENOENT → non-zero exit → build fails. That's the real signature, so the guard now matches its comment.

(The probe binary is also copied into `/tmp/rootfs/tmp/` so it's reachable inside the confined rootfs.)

The other notes I'm treating as acknowledged/non-blocking: aarch64 has no smoke test (QEMU probing unreliable — x86_64 gate + the #286 end-to-end install cover it), and `contents: write` scoping is intentional for the publish job.

PR #286, opened 2026-07-20
feat(#252): verify + repoint proot download; curl XferCommand

### LINKED ISSUE #248: proot sandbox: pacman -Sy succeeds but databases unreadable -> all tool installs fail (target not found)

jtomek-strike48:
Part of #252. Completes the code for the #248 fix. **Now live** - the release
exists and the checksums are pinned, so this closes #248.

Stacked on #285 (the build workflow), which is now merged; rebased onto main.

### Downloader (the original scope of this PR)
- **`proot.rs`**: replace the hardcoded proot-me v5.3.0 URLs with a
  `PROOT_DOWNLOADS` table `(arch, url, sha256)` pointing at the
  `proot-openat2-v1` release. `download_proot` verifies the SHA-256 before
  writing/chmod (`verify_sha256`), closing the supply-chain gap in the old
  unchecked `get -> write -> chmod` path. Fails closed on an unpinned or
  mismatched checksum.
- **`Cargo.toml`**: add `dep:sha2` to the `desktop` feature.
- **`rootfs.rs`** (`fix_pacman_config`): force
  `XferCommand = /usr/bin/curl -L -f -o %o %u` - pacman 7's internal parallel
  downloader is a silent no-op under proot (#248, step 4).

### Build-script fixes (found by actually running #285's workflow)
#285 merged the build workflow but it had never been run. Running it surfaced
four real bugs in `build.sh`, each fixed here (see individual commits):
1. **`--rosegment`**: the sed to strip the lld-only flag corrupted termux/proot's
   own feature-detection probe, forcing the flag on -> loader link failed on GNU ld.
   Removed it; the makefile already auto-detects.
2. **`LDFLAGS`**: `make LDFLAGS=-static` overrode the makefile's `-ltalloc`,
   failing the main link with `undefined reference to _talloc_*`. Pass it via the
   environment so the makefile's `+=` appends.
3. **`-m32`**: the 32-bit compat loader can't build on aarch64 (`cc -m32`
   unrecognized). Disabled with `make HAS_LOADER_32BIT=`; the sandbox only runs
   native-arch tools.
4. **`gawk`**: `loader-info.awk` uses `strtonum()`, a gawk extension; Debian
   defaults to mawk. Install gawk so `awk` resolves it.

### Checksum pin (makes it live)
- Ran build-proot.yml (run 29788507360): both arches built, openat2 smoke test
  passed, `proot-openat2-v1` published with proot-{x86_64,aarch64} + `.sha256`.
- Pinned both verified digests into `PROOT_DOWNLOADS`. Verified three ways:
  against the release `.sha256` files, by recomputing `sha256sum` over the
  downloaded binaries, and by running the x86_64 binary locally.

## Test plan
- [x] `verify_sha256` unit tests: match (case-insensitive) / mismatch / empty-pin-refused.
- [x] New guard test: every `PROOT_DOWNLOADS` entry has a well-formed 64-hex
      lowercase SHA-256 (regression to the empty placeholder fails at test time).
      Mutation-checked - emptying a pin turns it red.
- [x] `cargo fmt --all --check`, `clippy --workspace -D warnings`, full
      `cargo test --workspace` - all green.
- [x] build-proot.yml ran green end-to-end on both arches incl. the openat2
      smoke test (the exact #248 signature); release published.
- [x] Published binaries verified static ELF + correct arch; x86_64 proot runs
      locally.
- [ ] Live end-to-end in a fresh connector sandbox: proot downloaded + verified,
      `pacman -Sy` yields readable DBs, a tool installs (the on-device proof of
      #248). Requires a Linux sandbox host with a real BlackArch rootfs; not run
      from here.

Closes #248 (pending the on-device confirmation above).

### DIFF
--- a/.github/proot-build/build.sh
+++ b/.github/proot-build/build.sh
@@ -35,8 +35,14 @@ git checkout "$TERMUX_PROOT_REF"
 mkdir -p /usr/include/linux
 cp /work/.github/proot-build/ashmem.h /usr/include/linux/ashmem.h
 
-# Static glibc build.
-make -C src LDFLAGS="-static" V=1
+# Static glibc build. Pass LDFLAGS through the ENVIRONMENT, not as a make
+# argument: `make LDFLAGS=-static` is a command-line assignment that OVERRIDES
+# the makefile's `LDFLAGS += -ltalloc -Wl,-z,noexecstack` (src/GNUmakefile:19),
+# dropping -ltalloc and failing the final link with `undefined reference to
+# _talloc_*`. Setting it in the environment lets the makefile's `+=` append, so
+# both -static and -ltalloc reach the link. (Verified: GNU make gives env vars
+# lower precedence than `+=`, command-line vars higher.)
+LDFLAGS="-static" make -C src V=1
 /src/src/proot --version
 
 cp /src/src/proot "/work/proot-${ARCH}"
--- a/.github/proot-build/build.sh
+++ b/.github/proot-build/build.sh
@@ -35,14 +35,23 @@ git checkout "$TERMUX_PROOT_REF"
 mkdir -p /usr/include/linux
 cp /work/.github/proot-build/ashmem.h /usr/include/linux/ashmem.h
 
-# Static glibc build. Pass LDFLAGS through the ENVIRONMENT, not as a make
-# argument: `make LDFLAGS=-static` is a command-line assignment that OVERRIDES
-# the makefile's `LDFLAGS += -ltalloc -Wl,-z,noexecstack` (src/GNUmakefile:19),
-# dropping -ltalloc and failing the final link with `undefined reference to
-# _talloc_*`. Setting it in the environment lets the makefile's `+=` append, so
-# both -static and -ltalloc reach the link. (Verified: GNU make gives env vars
-# lower precedence than `+=`, command-line vars higher.)
-LDFLAGS="-static" make -C src V=1
+# Static glibc build.
+#
+# LDFLAGS via the ENVIRONMENT, not as a make argument: `make LDFLAGS=-static` is
+# a command-line assignment that OVERRIDES the makefile's
+# `LDFLAGS += -ltalloc -Wl,-z,noexecstack` (src/GNUmakefile:19), dropping
+# -ltalloc and failing the final link with `undefined reference to _talloc_*`.
+# In the environment, the makefile's `+=` appends, so both -static and -ltalloc
+# reach the link. (GNU make: env vars have lower precedence than `+=`.)
+#
+# HAS_LOADER_32BIT= (command-line, empty) disables the 32-bit compat loader. The
+# makefile derives it from arch.h and, on aarch64, tries to build loader-m32 with
+# `cc -m32` — which the aarch64 toolchain has no notion of (`unrecognized
+# command-line option '-m32'`). The sandbox only runs native-arch tools, so the
+# 32-bit guest loader is dead weight; disable it on every arch for a uniform,
+# reproducible build. (GNU make: a command-line `VAR=` overrides the makefile's
+# `:=`, and `ifdef` on the resulting empty value is false.)
+LDFLAGS="-static" make -C src HAS_LOADER_32BIT= V=1
 /src/src/proot --version
 
 cp /src/src/proot "/work/proot-${ARCH}"
--- a/.github/proot-build/build.sh
+++ b/.github/proot-build/build.sh
@@ -5,7 +5,12 @@
 # bind-mounted at /work. Reads $TERMUX_PROOT_REF and $ARCH from the env.
 # Recipe verified A/B during the #248 spike:
 #   - glibc-static (musl fails on `struct rlimit64`)
-#   - strip the lld-only `--rosegment` linker flag (GNU ld rejects it)
+#   - let the makefile auto-detect the lld-only `--rosegment` flag: its src/GNUmakefile
+#     probe test-links with `-Wl,--rosegment` and only enables it if the linker
+#     accepts it, so on GNU ld (which rejects it) it is skipped automatically.
+#     Do NOT sed it out — the flag's only literal occurrence is inside that probe,
+#     so editing it breaks detection and forces `--rosegment` onto the real link
+#     line, which GNU ld then rejects (the loader/loader link fails).
 #   - stub <linux/ashmem.h> so the Android-only ashmem_memfd extension compiles
 #     (do NOT delete the object — cli/proot.c references ashmem_memfd_callback)
 set -euo pipefail
@@ -22,12 +27,11 @@ git clone https://github.com/termux/proot /src
 cd /src
 git checkout "$TERMUX_PROOT_REF"
 
-# Patch 1: strip the lld-only --rosegment flag (GNU ld rejects it).
-if grep -rl -- "--rosegment" . >/dev/null 2>&1; then
-  grep -rl -- "--rosegment" . | xargs sed -i "s/-Wl,--rosegment//g"
-fi
+# NOTE: no --rosegment patch. The makefile's own probe (src/GNUmakefile) disables
+# the flag on GNU ld automatically; sed-ing it out corrupts that probe and forces
+# the flag on, breaking the loader link (#252, first live run of this workflow).
 
-# Patch 2: install the committed linux/ashmem.h stub.
+# Patch: install the committed linux/ashmem.h stub.
 mkdir -p /usr/include/linux
 cp /work/.github/proot-build/ashmem.h /usr/include/linux/ashmem.h
 
--- a/.github/proot-build/build.sh
+++ b/.github/proot-build/build.sh
@@ -19,9 +19,13 @@ set -euo pipefail
 : "${ARCH:?ARCH must be set}"
 
 apt-get update
+# gawk (not the default mawk): the makefile's loader-info.awk uses strtonum(),
+# a gawk extension. Under mawk the loader-info.c generation step dies with
+# "function strtonum never defined" (GNUmakefile:246). Installing gawk points
+# Debian's `awk` alternative at gawk, so `awk` resolves strtonum on every arch.
 apt-get install -y --no-install-recommends \
   git ca-certificates build-essential libc6-dev \
-  pkg-config uthash-dev libtalloc-dev python3 gzip file
+  pkg-config uthash-dev libtalloc-dev python3 gzip file gawk
 
 git clone https://github.com/termux/proot /src
 cd /src
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -55,6 +55,14 @@ jobs:
   test:
     name: Test
     runs-on: ubuntu-latest
+    # CI runners have no BlackArch sandbox backend (no bwrap/proot installed).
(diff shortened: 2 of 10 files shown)


jtomek-strike48:
## Security review — APPROVED (no blocking findings)

Ran a security-reviewer pass focused on the download-verify path (this is a supply-chain-sensitive change). Verdict: **safe to merge, including before the proot-openat2-v1 release exists.**

### The crux — fail-closed CONFIRMED
With the placeholder empty `sha256` values, `verify_sha256` hits the `expected.is_empty()` guard and returns "refusing to run an unverified binary." So merging this before the release exists **cannot** regress to fetching/running the old unchecked binary — it errors loudly instead. That's the property that makes this mergeable now.

### Verified sound
- **`verify_sha256`**: correct digest (`Sha256::digest` + lowercase-hex `{:x}`), case-insensitive compare, length-safe (a wrong-length/non-hex `expected` fails the compare — no false ACCEPT), empty-pin refused first.
- **Verify-before-write ordering**: verification (line ~185) runs *before* `fs::write` and `chmod`, so a failed check never leaves an executable bad binary on disk.
- **XferCommand**: hardcoded string, no user interpolation; `%o`/`%u` are pacman placeholders (not shell-eval'd); insertion is idempotent (`!content.contains("XferCommand")` guard).
- **Cargo.toml**: `dep:sha2` on the `desktop` feature is minimal and matches the existing android pattern.
- Unit tests cover match / mismatch / empty-refuse.

### Non-blocking note (pre-existing, out of scope)
- **LOW / TOCTOU**: there's a write→exec window where a local attacker with write access to the proot bin dir could swap the binary after verification. Pre-existing, not introduced here; future hardening would verify-on-exec (fd pinning). Filed as a note, not a blocker.

### Test-plan honesty
The test plan is split into code-level checks (done: verify_sha256 tests, fmt/clippy/check, fail-closed by design) and **deployment-gated** items (run #285's build, pin checksums, live-sandbox end-to-end) that are inherently only verifiable after this ships — since this PR *is* the #248 fix. Those are tracked, not checked.

cchao-devo (approved):
Review: Pick PR #286 — feat(#252): verify + repoint proot download; curl XferCommand

  Verdict: ✅ Approve. The payload PR that closes #248. Supply-chain verification is correct and I independently confirmed the pinned checksums against the real release assets. Build-script
  fixes are well-reasoned (and correct the --rosegment issue from #285). Gates green. No blockers.

  Lenses run: mechanical gates (shellcheck/shfmt + fmt + workspace clippy + tests) + manual security trace of the download-verify path + independent checksum verification against the
  published release + build.sh + test-flake review. Verified in pick_08/pick; pick_07 untouched, no cluster.

  Scope note: PR body says stacked on #285 — #285 has since merged, so pr-286 is clean on current main; the diff is just #252's payload.

  Gates (on pr-286)

  - shellcheck -s bash build.sh clean · shfmt -i 2 -ci -bn clean · cargo fmt --all --check clean · cargo clippy --workspace -- -D warnings clean
  - tenant_uuid_from_env tests: 2/2 × 5 runs, no flake
  - ⚠️  Could not execute verify_sha256/pin-guard tests locally — they (and the whole proot.rs download path) are #[cfg(target_os = "linux")], filtered out on macOS. Verified by reading +
  the independent checksum check below. They run in CI.

  Supply-chain verification — independently confirmed ✅

  - I downloaded both actual release assets and hashed them: 2b5e31b7… (x86_64) and fddb9292… (aarch64) — both match the pins in PROOT_DOWNLOADS exactly. The strongest possible check on
  this PR, and it holds.
  - verify_sha256 is correct & fail-closed: lowercase-hex digest via {:x}, case-insensitive compare, empty-pin refused first (can't run unverified), wrong-length/non-hex expected simply
  fails the compare (no false ACCEPT).
  - Verify-before-write ordering confirmed: verify_sha256(&bytes,…)? runs before fs::write + chmod — a bad binary never lands executable on disk.
  - URL moved to the org's own release (Strike48-public/pick/…/proot-openat2-v1), pinned by checksum — no external-party trust.
  - The pin-guard test is a real safety net: catches a regression to empty/truncated pins at test time, and round-trips through verify_sha256 to prove the pin is well-formed.

  Build-script fixes — sound, and they close my #285 note

  All four are the real bugs only findable by running the workflow (which #285 couldn't), each documented with the underlying GNU-make/toolchain reason:
  - --rosegment sed removed — it corrupted the makefile's own detection probe and forced the flag on (breaking the GNU-ld link). This directly corrects the patch I flagged in my #285
  review; the makefile auto-detects correctly on its own.
  - LDFLAGS via env not make-arg — so the makefile's += appends -ltalloc instead of a command-line assignment clobbering it.
  - HAS_LOADER_32BIT= — disables the 32-bit compat loader that can't build on aarch64 (cc -m32).
  - gawk — loader-info.awk needs strtonum(), absent in Debian's default mawk.

  Other changes

  - rootfs.rs XferCommand: hardcoded curl string (no interpolation; %o/%u are pacman placeholders), idempotent (!contains("XferCommand")), inserted after [options] with a fallback. Correct
  — addresses #248 step 4 (pacman 7's internal parallel downloader is a silent no-op under proot).
  - config.rs test-flake fix (slightly out of stated scope, but a real fix): consolidates two separate per-test static LOCKs into one shared TENANT_ENV_LOCK. This is the same bug class as
  #272 — two private mutexes provide zero mutual exclusion, so the tenant-env tests raced and one flipped the other's is_none() assertion. Correctly fixed. Reasonable to fold in (it's the
  macOS flake named in the commit).
  - Cargo.toml: dep:sha2 on the desktop feature — minimal, matches the android pattern.

  Honest test-plan gap (acknowledged)

  The final live end-to-end (fresh sandbox: download+verify proot → pacman -Sy readable DBs → tool installs) is unchecked — it needs a Linux sandbox host with a real BlackArch rootfs.
  That's the on-device proof of #248 and can only happen post-merge. Appropriately tracked, not claimed. Everything code-level is verified; the checksum pins are confirmed against the live
  release.

  init-dev relevance

  Indirect: once this lands and the on-device #248 proof passes, the sandbox pacman-install path works — which is the underlying reason init-dev currently runs Pick with
  pick.disableSandbox=true. That workaround could eventually be revisited, but nothing changes for init-dev from this PR (init-dev's headless connector runs sandbox-disabled; this only
  affects the sandbox proot path).

jtomek-strike48:
Heads-up before merge: the head advanced past the approval (approved at a6a6eae). The five commits since are all CI/test hardening, no production-behavior change beyond the proot download timeout:

- test(#252): disable sandbox in execute_command provenance tests
- test(#252): disable sandbox in catalog tests that probe install-state
- fix(#252): bound proot download timeout + ignore network-hitting detection test
- ci(#252): disable sandbox for the test jobs (DISABLE_SANDBOX=true) — fixes a 6-hour CI Test hang systemically
- test(#252): serialize HOME-mutating config tests under the shared env lock (pre-existing flake)

Root cause of the hang: with the SHA pins populated, tests that ran a real command fell through to download_proot + BlackArch rootfs extraction, which fails/hangs as a non-root CI runner. CI now runs with the sandbox disabled (no backend to sandbox into anyway). All checks green.

The one production change (download_proot HTTP timeout) also addresses the latent fail-hang tracked in #294.

Issue #288, opened 2026-07-20
Contain liveview render panics with ErrorBoundary (one bad value should not blank the whole connector pane)


jtomek-strike48:
## Summary

A single render-time panic anywhere in the Pick liveview UI **blanks the entire connector
pane** and replaces it with `"Encountered panic: Any { .. }"`. There is no graceful
degradation: one bad conversation title, one unexpected value, one edge case in any
component takes down the whole `WorkspaceApp`. This issue asks for `ErrorBoundary`
containment so a render panic degrades to a localized, readable fallback (and ideally a
retry) instead of destroying the connector UI.

This is **defense-in-depth**, distinct from fixing any individual panic (e.g. #287). It
does not replace fixing the underlying bugs - it stops the *next* unanticipated render
panic from being pane-fatal.

## Current behavior (confirmed in code)

- Pick's liveview root is `LiveViewApp -> WorkspaceApp {}` (`crates/ui/src/liveview_server.rs:194-197`).
- **Pick declares no `ErrorBoundary` anywhere** (`grep -rIn "ErrorBoundary" crates/ui/src`
  returns nothing). So the only boundary is Dioxus's implicit **root** boundary.
- dioxus-core 0.7.3 catches a component render panic via `catch_unwind`
  (`any_props.rs:88-95`) and, with no nearer boundary, the **root** boundary renders
  `"Encountered panic: {:?}"` in place of the whole tree -> the entire pane goes blank.

Net: any render panic in any component = total pane loss, with an opaque message the user
can't act on. #287 (UTF-8 byte-slice on a conversation title) is the live example - a data
value from one tenant blanked the connector.

## Proposed change

Wrap the major `WorkspaceApp` subtrees in `ErrorBoundary` with a `handle_error` fallback so
a panic is **contained to that region**:

- At minimum, wrap the always-mounted chrome pieces independently (`Sidebar`, each
  `WorkspacePages` page, `ChatPanel`) so a panic in one does not take the others down. E.g.
  a `Sidebar` render panic should leave the main content usable and show
  "Couldn't load the sidebar" rather than blanking the connector.
- The fallback should render a short, human-readable message (not the raw `Any { .. }`),
  and where practical a **Retry** affordance. Dioxus 0.7.3 supports this: `ErrorBoundary`
  with `handle_error: |errors: ErrorContext| { ... }`, and `ErrorContext::clear_errors()`
[...]

PR #290, opened 2026-07-20
fix(#287, #288): stop UTF-8 render panics from blanking the connector pane

### LINKED ISSUE #287: WorkspaceApp liveview panics on render -> "Encountered panic: Any { .. }" in StrikeHub dashboard pane (UTF-8 byte-slice on conversation title in Sidebar)
# WorkspaceApp liveview panics on render -> "Encountered panic: Any { .. }" in the StrikeHub dashboard pane (UTF-8 byte-slice on conversation title in Sidebar)

## Summary

A customer on Windows running Pick **v0.1.8** (current latest release) cannot open the
Pick connector in StrikeHub. Sign-in succeeds, both connectors report Online, and the
dashboard "flashes" - then the Pick pane is replaced by the text:

```
Encountered panic: Any { .. }
```

This is Dioxus's default error-boundary fallback. It means a Rust panic occurred
**during a render** of Pick's liveview UI (`WorkspaceApp`), was caught by the
per-VirtualDom error boundary, and the pane was swapped for the fallback text. The
`pentest-agent` process does not exit. (The panic is on a post-mount re-render, not the
first paint - see Root cause below; that is why the dashboard "flashes" before the error.)

## Why this is a Pick issue (not StrikeHub)

- The Pick connector hosts an **in-process `dioxus-liveview` server** whose root is
  `LiveViewApp -> WorkspaceApp {}` (`crates/ui/src/liveview_server.rs:194-197`). The
  VirtualDom runs inside `pentest-agent` and streams DOM into StrikeHub's content-pane
  iframe.
- StrikeHub's shell and Pick's liveview UI are **separate VirtualDoms**, each with its
  own Dioxus default error boundary (neither app declares an explicit `ErrorBoundary`).
- In the customer's StrikeHub log, the shell's `[connector-start]` `use_effect` keeps
  firing every ~3s to the end of the capture, long after the error appears. If a
  **StrikeHub** component had panicked in render, Dioxus's root boundary would tear down
  the whole tree and those effects would stop. They don't - so the StrikeHub shell is
  healthy and the caught panic lives in **Pick's** VirtualDom.
- "Flashes then error, confined to the pane" is the exact signature of a render panic in
  `WorkspaceApp` caught by Pick's own boundary.

## Environment

| | |
|---|---|
| Pick version | **v0.1.8** (latest release tag as of 2026-07-13; HEAD is only 21 commits ahead) |
| Platform | **Windows** (`pentest-agent.exe`, spawned by StrikeHub via IPC / named pipe) |
| Dioxus | 0.7.3 (`dioxus`, `dioxus-liveview`) |
| Host | StrikeHub desktop, Studio `<customer-studio-host>`, transport wss |
| Mode | liveview / IPC (`STRIKEHUB_SOCKET` set) |

## The error string is opaque by design

`"Encountered panic: Any { .. }"` comes from `dioxus-core`'s `CapturedPanic`, which
formats the panic payload with `{:?}` on a `Box<dyn Any>` - that is **always**
`Any { .. }` regardless of the real message. The panic reason is NOT recoverable from the
string itself; it was identified by auditing the render tree (below) and reproduced
standalone. A `RUST_BACKTRACE=full` capture (see Reproduction) would additionally show
`sidebar.rs:204` in a frame for belt-and-suspenders confirmation.

## Root cause: UTF-8 byte-slice panic during render in `Sidebar` (`sidebar.rs:204`)

A caught **render-time** panic - dioxus-core 0.7.3 wraps component render in `catch_unwind`
(`any_props.rs:88-95`) and renders the caught payload as `"Encountered panic: {:?}"` =
`Any { .. }`. (Event-converter/handler panics do NOT surface this way - they abort the
process; see the retraction note below.)

**`crates/ui/src/components/sidebar.rs:204`:**

```rust
} else if conv.title.len() > 28 {
    format!("{}...", &conv.title[..25])   // byte slice; panics if byte 25 is not a char boundary
}
```

`&conv.title[..25]` is a **byte** slice on a `String`. The `.len() > 28` guard checks byte
length only - it does not guarantee byte 25 is a UTF-8 char boundary. If a multi-byte char
(accent, emoji, CJK, smart-quote - all common in AI-generated titles) straddles byte 25,
Rust panics with `byte index 25 is not a char boundary`. This runs inside the `Sidebar`
render body (`rsx!`, not an event handler), so dioxus-core catches it and renders the
`Any { .. }` fallback in the pane.

**Empirically reproduced** (standalone Rust): a title where a 2-byte `é` straddles byte 25
panics with exactly `byte index 25 is not a char boundary`; a boundary-aligned title of the
same length is fine. Content-dependent - which is why it hits this customer's tenant and not
every user.

**Why it matches "dashboard flashes, THEN errors" (post-mount re-render, not first paint):**

1. `Sidebar` is always-mounted chrome (`AppLayout` mounts it on every page, `app_layout.rs:67`).
   The recent-convos block is inside `for page in ALL_PAGES` and gated on the **loop
   variable** `matches!(page, NavPage::Chat)` - NOT on the active page - so it renders
   whenever the Chat nav item is drawn, i.e. on the Dashboard too (`sidebar.rs:196-198`).
2. On first paint `recent_convos` is empty (`use_signal(Vec::new)`, `sidebar.rs:109`), so
   the `!is_empty()` guard is false and no slice runs. Dashboard paints -> the "flash."
3. Post-mount, StrikeHub credentials arrive (`api_url`/`auth_token` threaded into `Sidebar`
   props); a `use_effect` fetches conversations and calls `recent_convos.set(list)`
   (`sidebar.rs:111-134`).
4. That `.set()` re-renders `Sidebar`; now the list is non-empty, the byte-slice runs on
   real backend titles (`title: edge.node.title.unwrap_or("Untitled")`,
   `crates/core/src/matrix/client.rs:754`), and a boundary-straddling title panics -> caught
   -> `Any { .. }`.

**Byte-identical at v0.1.8** (`git diff v0.1.8 -- sidebar.rs` empty; 0 commits since) -
live bug on the customer's build.

### Sibling instances of the same antipattern (same fix, different mount conditions)

Identical byte-slice bug, but in `ChatPanel` children (only mounted when
`active_page == NavPage::Chat`), so not the Dashboard-pane panic - but they will panic the
same way on chat titles/tool strings and should be fixed together:

- `crates/ui/src/components/chat_panel/history.rs:45` - `&conv.title[..37]` (guard `> 40`)
- `crates/ui/src/components/chat_panel/agent_selector.rs:382` - `&conv.title[..47]` (guard `> 50`)
- `crates/ui/src/components/chat_panel/render.rs:108` - `&t[..57]` (guard `> 60`) on a tool `task` string

### Suggested fix

Slice on **char boundaries**, not bytes, at all four sites. E.g.
`conv.title.chars().take(25).collect::<String>()` (then append `...`), or a small
`truncate_chars(&str, n)` helper shared across the four call sites. No platform-specific
handling and no `document::eval` bridge is involved - this is pure Rust string logic, safe
under the Windows WebView2 + LiveView + iframe path (unlike the retracted approach).

### Ruled out (confirmed safe on the render path)

- `load_settings()` / `save_settings()` (`crates/core/src/settings.rs`) - `unwrap_or_default`
  / `?` throughout.
- `jwt_validator::validate_token()` - `parts[1]` guarded by `parts.len() != 3`.
- `rendering.rs:22` `ts.themes["base16-ocean.dark"]` - compiled-in syntect default; effectively safe.
- `use_context` consumers in the tree - `ChatHeaderCtx` is provided by `AppLayout`
  (`app_layout.rs:37`); `AppState`/`ViewRegistry` consumers are not mounted in this tree.
- StrikeHub shell (separate VirtualDom) - healthy; its root-scope effects fire to
  end-of-log, which a shell render panic would have stopped.

---

## Retraction note (earlier attribution was wrong)

An earlier version of this issue attributed the panic to `dioxus-liveview`'s
`convert_keyboard_data` via the `onkeydown` in `keyboard_shortcuts.rs:101`, with a
`document::eval` JS-bridge fix. **That was wrong** and is retracted:

- dioxus-core catches **render** panics (`any_props.rs:88-95`); event-**converter** panics
  run outside that `catch_unwind` and **abort** the process (#130: `fatal runtime error ...
  aborting`; #224: `Abort trap: 6`). The customer's panic was *caught and rendered*, so it
  is a render panic, not a converter panic.
- Merged **#191** already established the real converter crash site is `convert_mounted_data`
  (not keyboard) and **empirically confirmed native keyboard events work on Windows MSI**.
- The `document::eval` JS-bridge is exactly what **#191 reverted** because the reverse
  channel does not deliver in Windows WebView2 + LiveView + iframe (#189).

## Reproduction

The **mechanism** is reproduced (standalone Rust): a title where a 2-byte `é` straddles
byte 25 panics with `byte index 25 is not a char boundary` at `&title[..25]`; a
boundary-aligned title of the same length does not. A minimal case:

```rust
let t = format!("{}{}review of scope", "a".repeat(24), "\u{e9}"); // 'é' spans bytes 24-25
assert!(t.len() > 28);
let _ = &t[..25]; // panics: byte index 25 is not a char boundary
```

To confirm the **full end-to-end path on Windows** (optional, for belt-and-suspenders):
run `pentest-agent.exe` directly (not via StrikeHub) with `RUST_BACKTRACE=full` and
`STRIKEHUB_SOCKET` set so it takes the liveview path, ensure the tenant has a conversation
whose title has a multi-byte char near byte 25, open `/liveview`, and capture stderr - the
backtrace should name `sidebar.rs:204`.

> Note: today a connector panic on Windows is **invisible** in the host log - StrikeHub
> nulls the child's stdout/stderr on Windows, and Pick has no file logging or Sentry. That
> observability gap is tracked in strikehub#57 (and Pick #278 for connector-side capture);
> it is why this had to be diagnosed statically rather than from a backtrace.

## Acceptance criteria

- [ ] `sidebar.rs:204` slices `conv.title` on a **char boundary**, not a byte index (e.g.
      `chars().take(25).collect()` or a shared `truncate_chars` helper).
- [ ] The three sibling sites are fixed the same way: `chat_panel/history.rs:45`,
      `chat_panel/agent_selector.rs:382`, `chat_panel/render.rs:108`.
- [ ] A regression test covers the truncation with a multi-byte title whose char straddles
      the byte cutoff (e.g. an accented or emoji title just over the length guard) and
      asserts no panic + correct char-count truncation. A unit test on the helper is enough;
      a `VirtualDom` render test of `Sidebar` with such a title in `recent_convos` is
      stronger.
- [ ] (Optional but recommended) grep the codebase for other `&<str>[..N]` byte-slice
      truncations on backend/user data and convert them, so the class doesn't recur.
- [ ] Verify in the StrikeHub Windows pane (or standalone connector run) that a tenant with
      a multi-byte conversation title no longer panics the Pick pane.

## Labels

`product/pick`, `type/bug`, `area/frontend`, `platform/windows`, `priority/P1`

## Related

- Pick #278 - Add Sentry telemetry to Pick (would have surfaced this panic's backtrace).
- strikehub#57 - Capture connector stdout/stderr into the host log + pass `RUST_BACKTRACE`
  (Windows connector panics are currently invisible - the reason this needed static
  diagnosis).
- Pick #130 / #191 / #224 - the *event-converter* panic class (abort, not caught); see the
  retraction note for why this issue is a distinct render-time panic.

jtomek-strike48:
Fixes #287 and #288.

## Summary

A customer on Windows (Pick v0.1.8) saw the connector pane replaced by
`Encountered panic: Any { .. }`. Root cause (#287): a byte-slice truncation
of a conversation title (`&conv.title[..25]`) panics when byte 25 is not a
UTF-8 char boundary. AI-generated titles routinely contain multi-byte
characters (accents, emoji, CJK, smart quotes), so the slice panics; dioxus
catches it in the per-VirtualDom render boundary and blanks the whole pane.

This PR fixes the concrete bug (#287) and adds defense-in-depth so the next
unanticipated render panic degrades gracefully instead of blanking the pane
(#288).

### #287 - char-boundary-safe truncation (the cause)
- New `pentest_ui::text::truncate_chars` counts characters, not bytes, and
  appends an ellipsis. Routed through all four UI truncation sites:
  `sidebar.rs`, `chat_panel/{history,agent_selector,render}.rs`.
- Grepped the tree for the whole class and hardened a fifth site,
  `matrix/client.rs::truncate_body` (`&body[..200]` on a hostile GraphQL
  error body), to back off to the nearest char boundary while keeping its
  byte-count report. The grep (`&.*\[\.\..*\]` over UI + core) is exhaustive
  for user/backend strings.

### #288 - `PaneBoundary` containment (the class)
- New `PaneBoundary` (a thin `ErrorBoundary` wrapper) with a readable,
  inline-styled fallback and a Retry button (`clear_errors()`).
- Wraps `Sidebar` and page content independently in `AppLayout`, and the
  always-mounted Files/Shell panes in `WorkspaceApp`, so a render panic in
  one region no longer blanks the others.
- Scope caveat (documented in the module and fallback): this catches
  render-time panics only; dioxus-liveview event-converter/handler panics run
  outside the render `catch_unwind` and abort the process (#130/#191/#224).

## Test plan
- [x] `truncate_chars` unit tests: accent/emoji/CJK straddling the byte
      cutoff, exact-limit, empty input (+ doctest).
- [x] `truncate_body` unit test: a 2-byte char straddling byte 200.
- [x] End-to-end guard on the `render.rs` call site (`webwright_display_name`,
      a pure function): panics today if byte-sliced, asserts char truncation
      now. Mutation-checked - reverting the site to `&t[..57]` turns it red.
- [x] `PaneBoundary` render test: a component that panics on render is
      contained, an unwrapped sibling still renders, and the raw `Any { .. }`
      payload never surfaces. Mutation-checked - neutering the boundary turns
      it red.
- [x] All new tests run in the required CI job (`cargo test --workspace`,
      no `liveview` feature); `dioxus-ssr` added as a dev-dependency for the
      render tests.
- [x] `cargo fmt --all`, `cargo clippy --workspace --all-targets -D warnings`,
      full `cargo test --workspace` - all green.
- [ ] Live verification in the StrikeHub Windows pane with a multi-byte
      conversation title (AC #287.5 - manual/release testing; not run here).

## Notes
- Blast radius traced: inserting `PaneBoundary` between `AppLayout` and its
  children does not break the `ChatHeaderCtx` context path (context resolves
  through the scope chain), the flex layout (`ErrorBoundary` is DOM-
  transparent), or the mobile sidebar drawer.

### DIFF
--- a/crates/core/src/matrix/client.rs
+++ b/crates/core/src/matrix/client.rs
@@ -191,11 +191,18 @@ fn check_errors(errors: Option<Vec<GqlError>>) -> crate::error::Result<()> {
 }
 
 fn truncate_body(body: &str) -> String {
-    if body.len() <= 200 {
-        body.to_string()
-    } else {
-        format!("{}…({} bytes total)", &body[..200], body.len())
+    const MAX_PREVIEW_BYTES: usize = 200;
+    if body.len() <= MAX_PREVIEW_BYTES {
+        return body.to_string();
     }
+    // Walk down to the nearest UTF-8 char boundary at or below the byte budget.
+    // A bare `&body[..200]` panics if byte 200 falls inside a multi-byte char,
+    // which a hostile/compromised server could trigger via the error path (#287).
+    let mut end = MAX_PREVIEW_BYTES;
+    while end > 0 && !body.is_char_boundary(end) {
+        end -= 1;
+    }
+    format!("{}…({} bytes total)", &body[..end], body.len())
 }
 
 // ---------------------------------------------------------------------------
@@ -841,5 +848,47 @@ impl MatrixChatClient {
         }
 
         Ok(response)
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::truncate_body;
+
+    #[test]
+    fn returns_short_body_unchanged() {
+        let body = "small error body";
+        assert_eq!(truncate_body(body), body);
+    }
+
+    #[test]
+    fn truncates_long_ascii_body_with_byte_count() {
+        let body = "a".repeat(500);
+        let out = truncate_body(&body);
+        assert!(out.starts_with(&"a".repeat(200)));
+        assert!(out.ends_with("(500 bytes total)"));
+    }
+
+    #[test]
+    fn does_not_panic_when_multibyte_char_straddles_byte_cutoff() {
+        // Place a 2-byte 'é' straddling byte 200: bytes 0..199 are ASCII, then
+        // 'é' occupies bytes 199-200, so a bare `&body[..200]` would panic with
+        // "byte index 200 is not a char boundary" (#287). truncate_body must not.
+        let body = format!("{}\u{e9}{}", "a".repeat(199), "b".repeat(400));
+        assert!(
+            !body.is_char_boundary(200),
+            "precondition: 200 splits a char"
+        );
+
+        let out = truncate_body(&body);
+
+        // It backs off to byte 199 (the last boundary <= 200) and still reports
+        // the true byte total.
+        assert!(out.starts_with(&"a".repeat(199)));
+        assert!(out.contains("bytes total)"));
+        assert_eq!(
+            out,
+            format!("{}…({} bytes total)", "a".repeat(199), body.len())
+        );
     }
 }
--- a/crates/ui/Cargo.toml
+++ b/crates/ui/Cargo.toml
@@ -89,4 +89,8 @@ strike48-proto = { workspace = true, optional = true }
 
 [dev-dependencies]
 tempfile = { workspace = true }
+# Server-side render used by ErrorBoundary containment tests (#288). Declared
+# here (not gated behind `liveview`) so the render tests run in the required CI
+# job, which does NOT enable the `liveview` feature.
+dioxus-ssr = { workspace = true }
 
--- a/crates/ui/src/components/app_layout.rs
+++ b/crates/ui/src/components/app_layout.rs
@@ -4,6 +4,7 @@ use dioxus::prelude::*;
 
 use super::chat_panel::{ChatHeaderActions, ChatHeaderCtx};
 use super::icons::{Menu, STRIKE48_SIDEBAR_LOGO_SVG};
+use super::pane_boundary::PaneBoundary;
 use super::sidebar::{NavPage, Sidebar};
 use super::status_bar::StatusBar;
 
@@ -63,20 +64,24 @@ pub fn AppLayout(
                 }
             }
 
-            // Sidebar / Drawer
-            Sidebar {
-                active_page,
-                on_navigate,
-                sidebar_open: *sidebar_open.read(),
-                sidebar_collapsed: *sidebar_collapsed.read(),
-                on_close,
-                on_toggle_collapse,
-                unread_logs,
-                connected,
-                host: host.clone(),
-                api_url,
-                auth_token,
-                on_open_conversation,
+            // Sidebar / Drawer — wrapped so a render panic in the sidebar
+            // (e.g. a bad conversation title, #287) is contained and does not
+            // blank the main content (#288).
+            PaneBoundary { name: "sidebar".to_string(),
+                Sidebar {
+                    active_page,
+                    on_navigate,
+                    sidebar_open: *sidebar_open.read(),
+                    sidebar_collapsed: *sidebar_collapsed.read(),
+                    on_close,
+                    on_toggle_collapse,
+                    unread_logs,
+                    connected,
+                    host: host.clone(),
+                    api_url,
+                    auth_token,
+                    on_open_conversation,
+                }
             }
 
             // Main column (header + content + status bar)
@@ -116,9 +121,13 @@ pub fn AppLayout(
                     }
                 }
 
-                // Content area
+                // Content area — wrapped so a render panic in the active page
+                // or ChatPanel is contained to the content region, leaving the
+                // sidebar and header usable (#288).
                 div { class: "content-area",
-                    {children}
+                    PaneBoundary { name: "page".to_string(),
+                        {children}
+                    }
                 }
 
                 // Status bar
--- a/crates/ui/src/components/chat_panel/agent_selector.rs
+++ b/crates/ui/src/components/chat_panel/agent_selector.rs
@@ -10,6 +10,7 @@ use pentest_core::matrix::{AgentInfo, ConversationInfo};
 use super::constants::SUGGESTED_ACTIONS;
 use super::render::format_relative_time;
 use crate::components::icons::{ChevronDown, FileText, History, Plus, Shield};
+use crate::text::truncate_chars;
 
 /// Props for [`ChatHeader`].
 #[derive(Props, Clone, PartialEq)]
@@ -378,10 +379,9 @@ pub fn SuggestedActions(props: SuggestedActionsProps) -> Element {
                                 let cid = conv.id.clone();
                                 let title = if conv.title.is_empty() {
                                     "Untitled".to_string()
-                                } else if conv.title.len() > 50 {
-                                    format!("{}...", &conv.title[..47])
                                 } else {
-                                    conv.title.clone()
+                                    // Char-boundary-safe truncation (#287).
+                                    truncate_chars(&conv.title, 47)
                                 };
                                 let time_str = format_relative_time(&conv.updated_at);
                                 let on_select = props.on_select_conversation;
--- a/crates/ui/src/components/chat_panel/history.rs
+++ b/crates/ui/src/components/chat_panel/history.rs
@@ -8,6 +8,7 @@ use dioxus::prelude::*;
 use pentest_core::matrix::ConversationInfo;
 
 use super::render::format_relative_time;
+use crate::text::truncate_chars;
 
 /// Props for [`HistoryDropdown`].
 #[derive(Props, Clone, PartialEq)]
@@ -41,10 +42,9 @@ pub fn HistoryDropdown(props: HistoryDropdownProps) -> Element {
                         let conv_id_val = conv.id.clone();
                         let conv_title = if conv.title.is_empty() {
                             "Untitled".to_string()
-                        } else if conv.title.len() > 40 {
-                            format!("{}...", &conv.title[..37])
                         } else {
-                            conv.title.clone()
+                            // Char-boundary-safe truncation (#287).
+                            truncate_chars(&conv.title, 37)
                         };
                         let is_active = conversation_id
                             .read()
--- a/crates/ui/src/components/chat_panel/render.rs
+++ b/crates/ui/src/components/chat_panel/render.rs
@@ -761,3 +761,46 @@ fn load_screenshots_from_result(result_json: &str) -> Vec<(String, String)> {
         })
         .collect()
 }
+
+#[cfg(test)]
+mod tests {
+    use super::webwright_display_name;
+
+    // End-to-end guard for the render.rs truncation call site (#287). Unlike the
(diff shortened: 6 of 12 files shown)


cchao-devo (approved):
Review: Pick PR #290 — fix(#287, #288): stop UTF-8 render panics from blanking the connector pane
  
  Verdict: ✅ Approve. Textbook root-cause fix + defense-in-depth, with genuinely thorough coverage. I independently confirmed the grep was complete (the exact gap my own rules warn about)
  and all tests pass. No blockers, no findings.

  Lenses run: mechanical gates (fmt + workspace clippy --all-targets + tests) + manual correctness trace of truncate_chars/truncate_body + independent whole-tree hunt for the byte-slice 
  panic class + PaneBoundary containment review + call-site behavior-change check. Verified in pick_08/pick; pick_07 untouched, no cluster.

  Gates (on pr-290)

  - cargo fmt --all --check clean · cargo clippy --workspace --all-targets -- -D warnings clean
  - All new tests pass: truncate_chars (8, incl. é/emoji/CJK straddling byte 25, exact-limit, empty), truncate_body (2), webwright_display_name call-site guard (2), PaneBoundary render
  containment (2). Run with --features liveview to compile the UI crate (the known feature-gate quirk); the PR notes they run in CI's cargo test --workspace.

  Root cause fix (#287) — correct

  - truncate_chars uses chars().count()/chars().take() — panic-free on any UTF-8, appends ... only when actually over the limit. Matches (and slightly improves on) the old sites.
  - Behavior-preserving at the 4 UI call sites: every old site was format!("{}...", &title[..N]) — always appended .... Two even had a subtle secondary bug (guard threshold in bytes > 40
  differing from the slice point [..37]); the new truncate_chars(&title, 37) cleanly unifies both at 37 chars. No regression.

  Grep completeness — independently verified ✅ (the key check)

  I ran my own whole-tree hunt for string byte-slicing, not just trusting the PR's grep. Every string-slice site is accounted for:
  - The 4 UI sites + matrix/client.rs::truncate_body — fixed here (the 5th, the "different shape on hostile GraphQL error body," is exactly the kind of sibling site my rules say to check;
  the author found it).
  - provenance.rs:124 — already had is_char_boundary back-off (pre-existing, correct).
  - orchestrator.rs:448, theme.rs:681, whatweb.rs:118/121 — all slice at a str::find()-derived byte index, which is always a char boundary → safe, correctly left alone.
  - All &buf[..n] sites are on &[u8] byte buffers via from_utf8_lossy — not &str, panic-free.

  No missed panic site. The PR's "the grep is exhaustive" claim holds.

  Defense-in-depth (#288) — sound
  
  PaneBoundary is a thin ErrorBoundary wrapper with an inline-styled fallback (correctly not depending on possibly-unmounted CSS) + Retry via clear_errors(). The test is rigorous: renders a
  panicking component next to a healthy sibling through a real VirtualDom+dioxus-ssr, asserts the panic is contained, the sibling survives, and the raw Any { .. } payload never leaks. The
  two-tick render (rebuild then render_immediate) shows correct understanding of dioxus error routing. Scope caveat is honestly documented: this catches render-time panics only; liveview
  event-handler panics still abort the process (#130/#191/#224) — a boundary genuinely can't help those.

  Notes (non-blocking)

  - Blast-radius claims (context path through ChatHeaderCtx, flex layout with a DOM-transparent ErrorBoundary, mobile drawer) are asserted in the PR body but not independently exercised
  here — they're plausible and the render tests + clean compile support them; the live Windows-pane verification (AC #287.5) is appropriately deferred to release testing.
  - dioxus-ssr added as a dev-dependency (not a runtime dep) — correct, keeps it out of the shipped binary.

  init-dev relevance

  Low but positive: this hardens the connector UI (pentest-web/liveview) against render panics. init-dev deploys the headless connector which serves this same UI — so a Windows-title-style
  multi-byte panic that blanked the pane would affect init-dev's Studio-embedded connector view too. This fix helps there; no config change needed.

PR #291, opened 2026-07-21
fix(#256): fail closed when an enabled sandbox errors, don't run on host

### LINKED ISSUE #256: Non-interactive command path fail-opens to host execution when the sandbox errors
## Summary

Filed as a follow-up from the #239 review (cchao-devo). Pre-existing, not introduced by #239.

`execute_command` in `crates/platform/src/desktop/command.rs` runs tools inside the sandbox when it is enabled, but if the sandbox **errors** (either execution failure or sandbox-manager init failure) it silently falls back to running the command **directly on the host** — outside the sandbox the operator expects.

## Where

`crates/platform/src/desktop/command.rs`:
- Sandbox execution error → `warn!` + fall through (`~:108-111`)
- Sandbox manager init failure → `warn!` + fall through (`~:114-120`)
- Terminal fallback: `execute_command_direct(...)` on the host (`~:123-128`)

This is distinct from the intended `DISABLE_SANDBOX` / sandbox-disabled path (that host execution is deliberate). The concern is the *enabled-but-errored* path: a sandbox failure downgrades to host execution rather than failing closed.

## Why it matters

For a penetration-testing connector, "run this tool" implicitly means "run it in the sandbox." Falling back to the host on error means a command the operator believed was contained can execute against the host environment, with only a `warn!` log as signal. This is a fail-open; the security-conservative behavior is fail-closed (surface the sandbox error to the caller).

Note: the interactive PTY-shell path (#239) already fails closed — a backend that can't spawn/init surfaces an error rather than dropping to the host. This issue is specifically the **non-interactive tool-execution** path.

## Proposed direction

- When the sandbox is *enabled* and a command was expected to run sandboxed, treat a sandbox error as a hard error (return it) rather than silently retrying on the host.
- If a host fallback is ever intentional, gate it behind an explicit, logged-and-surfaced opt-in — not the default.
- Add a test asserting that an enabled-sandbox execution error does not silently run on the host.

## Acceptance criteria

- [ ] With the sandbox enabled, a sandbox execution/init error does not result in silent host execution.
- [ ] The failure is surfaced to the caller (and thus the operator/UI), not just `warn!`-logged.
- [ ] Test coverage for the enabled-sandbox-errors path.

**Recommended agent:** security-reviewer (fail-open / containment) with rust-reviewer for the implementation.

jtomek-strike48:
## Summary

Closes #256.

The non-interactive tool path (`execute_command_in_dir` in
`crates/platform/src/desktop/command.rs`) fell back to running commands
**directly on the host** when the sandbox failed to initialize or execute,
emitting only a `warn!`. For a pentest connector, "run this tool" implicitly
means "run it in the sandbox", so a sandbox failure silently downgrading to
host execution is a **fail-open**: a command the operator believed was
contained could execute against the host with no surfaced signal.

This change makes the enabled-sandbox path **fail closed**.

## What changed

- When the sandbox is *enabled*, a sandbox init or execution failure is now
  returned as `Error::ToolExecution(...)` instead of retried on the host. The
  tool-run wrapper (`crates/core/src/tools.rs`) maps that to
  `ToolOutcome::Failed`, so the operator and the model see an honest failure
  rather than host output posing as a sandboxed result.
- The fail-open/fail-closed decision is extracted into a pure
  `resolve_enabled_attempt(cmd, SandboxAttempt)` function so it is unit-testable
  without global sandbox state.
- Refreshed the stale doc comment on `execute_command_direct`, which claimed a
  sandbox failure falls back to it (no longer true).

## What is deliberately unchanged

- The `!is_sandbox_enabled()` branch (Native shell mode / `DISABLE_SANDBOX=true`)
  still runs on the host. That is the **explicit, logged opt-in** the issue asks
  for. The `pentest-core` integration tests call `set_use_sandbox(false)` and take
  this branch, so they are unaffected (their green CI signal is meaningful, not
  incidental).
- A command that *ran* in the sandbox but exited nonzero or timed out is a
  **successful sandbox run** (the executor returns `Ok(CommandResult)`);
  `Err(SandboxError)` is reserved for infrastructure failures. Failing closed on
  `Err` therefore rejects only genuine sandbox breakage, never a legitimate tool
  run that exited nonzero. Guarded by a regression test.

## Blast radius (consumer trace)

- `crates/tools/src/execute_command.rs` and the network/installer tool wrappers
  all propagate via `.await?`, so a broken sandbox now surfaces as a `Failed`
  tool result instead of host output. Verified `tools.rs` maps closure `Err` ->
  `ToolResult::error` -> `ToolOutcome::Failed`.
- The `which`-probe availability checks (`external/zap.rs`, `external/burp.rs`)
  use `.map(|r| r.exit_code == 0).unwrap_or(false)`. Under this change a broken
  sandbox makes them report "tool not available" -> **fail-safe** (nothing runs
  on the host), and they are pre-existing, not introduced here.
- **Headless connector IS affected, and this is intended.** Headless never calls
  `set_use_sandbox` (only desktop/web/UI wire the toggle) and sets no
  `DISABLE_SANDBOX` in `run-pentest.sh` or the `Dockerfile`, so `USE_SANDBOX`
  keeps its `true` default: headless runs **sandboxed**. The container image is
  `debian:bookworm-slim` as a non-root user with no bwrap/proot bundled, so it
  relies on the proot download (see #252/#286). Previously, if that download
  failed, headless silently ran tools **directly in the container** (the
  fail-open this PR closes); now it fails closed with a surfaced error. A headless
  carve-out would reintroduce the exact fail-open #256 is about, so headless is
  deliberately not exempted.

## Operator-facing behavior change (disclosed)

In a headless container where no sandbox backend is available and the proot
download fails, every tool run now returns a surfaced error instead of silently
executing in-container. This is the intended fail-closed direction, and it lands
alongside the proot-download hardening in #252/#286 that makes that backend
reliable. Operators who deliberately want unsandboxed execution set
`DISABLE_SANDBOX=true` (or Native shell mode), which is untouched.

## Acceptance criteria

- [x] With the sandbox enabled, a sandbox execution/init error does not result in
      silent host execution.
- [x] The failure is surfaced to the caller (returned `Err`), not just
      `warn!`-logged.
- [x] Test coverage for the enabled-sandbox-errors path.

## Test plan

- [x] `cargo test -p pentest-platform --lib desktop::command` (5 pass)
- [x] Mutation guard: neutering an error arm to a host result turns
      `sandbox_exec_failure_fails_closed` red; restored -> green.
- [x] `cargo test -p pentest-platform --lib` (42 pass)
- [x] `cargo test -p pentest-tools -p pentest-core --lib` (308 pass)
- [x] `cargo clippy --package pentest-platform --package pentest-core --lib --features pentest-platform/desktop-pcap --all-targets -- -D warnings` (clean)
- [x] `cargo fmt --all`

### DIFF
--- a/crates/core/src/config.rs
+++ b/crates/core/src/config.rs
@@ -873,6 +873,16 @@ impl AppSettings {
 mod tests {
     use super::*;
 
+    /// Serialises the tenant-env-var tests. These mutate process-global env
+    /// vars (`TENANT_ENV_VARS`), so they must not run concurrently or one
+    /// test's `set_var` leaks into another's view. A SINGLE shared lock is
+    /// required: two separate per-test locks (the previous state) provide no
+    /// mutual exclusion between the tests and let them race — on a busy runner
+    /// `..._prefers_matrix_tenant_id` (which sets MATRIX_TENANT_ID to a UUID)
+    /// could interleave with `..._returns_none_when_only_slugs_present` and
+    /// flip its `got.is_none()` assertion (seen flaking on macOS CI).
+    static TENANT_ENV_LOCK: std::sync::Mutex<()> = std::sync::Mutex::new(());
+
     #[test]
     fn preserves_explicit_wss_with_port() {
         let n = ConnectorConfig::normalize_host("wss://studio.example.com:443").unwrap();
@@ -1018,10 +1028,9 @@ mod tests {
 
     #[test]
     fn tenant_uuid_from_env_prefers_matrix_tenant_id() {
-        // Serialise all tenant-env manipulation on a single lock so parallel
+        // Serialise all tenant-env manipulation on the shared lock so parallel
         // tests can't leak into each other's env-var view.
-        static LOCK: std::sync::Mutex<()> = std::sync::Mutex::new(());
-        let _guard = LOCK.lock().unwrap_or_else(|e| e.into_inner());
+        let _guard = TENANT_ENV_LOCK.lock().unwrap_or_else(|e| e.into_inner());
 
         let previous: Vec<(&str, Option<String>)> = ConnectorConfig::TENANT_ENV_VARS
             .iter()
@@ -1053,8 +1062,7 @@ mod tests {
 
     #[test]
     fn tenant_uuid_from_env_returns_none_when_only_slugs_present() {
-        static LOCK: std::sync::Mutex<()> = std::sync::Mutex::new(());
-        let _guard = LOCK.lock().unwrap_or_else(|e| e.into_inner());
+        let _guard = TENANT_ENV_LOCK.lock().unwrap_or_else(|e| e.into_inner());
 
         let previous: Vec<(&str, Option<String>)> = ConnectorConfig::TENANT_ENV_VARS
             .iter()
--- a/crates/platform/src/desktop/command.rs
+++ b/crates/platform/src/desktop/command.rs
@@ -88,12 +88,24 @@ pub async fn execute_command_in_dir(
         full_cmd
     );
 
-    // Try sandboxed execution first
+    // Try sandboxed execution. The sandbox is enabled, so the operator expects
+    // the command to run *inside* it. We fail closed: a sandbox init or execution
+    // error is surfaced to the caller rather than silently retried on the host.
+    // Dropping to the host here would be a fail-open — a command the operator
+    // believed was contained could run against the host with only a `warn!` as
+    // signal (GitHub issue #256). The deliberate host path lives above, behind
+    // the `!is_sandbox_enabled()` check (Native shell mode / DISABLE_SANDBOX).
+    //
+    // Note the executors return `Ok(CommandResult)` even for a nonzero exit or a
+    // timeout (a command that *ran* but failed); `Err(SandboxError)` is reserved
+    // for infrastructure failures (rootfs missing, spawn failure, no backend). So
+    // failing closed here only rejects genuine sandbox breakage, never a
+    // legitimate tool run that exited nonzero.
     tracing::info!(
         "[execute_command] Sandbox enabled, attempting to get sandbox manager for: {}",
         cmd
     );
-    match sandbox::get_sandbox_manager().await {
+    let attempt = match sandbox::get_sandbox_manager().await {
         Ok(manager) => {
             tracing::info!("[execute_command] Sandbox manager obtained, backend={}, is_ready={}, executing: {}",
                 manager.backend(), manager.is_ready(), cmd);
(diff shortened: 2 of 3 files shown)


cchao-devo (approved):
Review: Pick PR #291 — fix(#256): fail closed when an enabled sandbox errors, don't run on host
  
  Verdict: ✅ Approve. A correct, well-tested, security-improving fail-open→fail-closed change with an honest and accurate consumer trace. Gates green, all guarantees verified end-to-end.
  No blockers — one init-dev-side item to confirm (below), which is not a defect in this PR.

  Lenses run: mechanical gates (fmt + workspace clippy --all-targets + tests) + manual trace of the fail-closed decision + the Err-vs-Ok(nonzero) distinction + full consumer trace (headless
  env, .await? → ToolOutcome::Failed) + cross-repo init-dev trace. Verified in pick_08/pick; pick_07 untouched, no cluster.

  Gates (on pr-291)
  
  - cargo fmt --all --check clean · cargo clippy --workspace --all-targets -- -D warnings clean
  - All 4 new tests pass: sandbox_exec_failure_fails_closed, sandbox_init_failure_fails_closed, successful_sandbox_run_passes_through, nonzero_exit_is_a_successful_run_not_a_sandbox_failure

  The security change is correct

  - resolve_enabled_attempt fails closed properly: an init or exec failure → Err(Error::ToolExecution(...)) with an actionable message (names the command, states host refusal, points to
  DISABLE_SANDBOX/Native mode). No host fallback. Extracted as a pure function → unit-testable without global state.
  - The crux distinction is right: Ok(CommandResult) = a command that ran (even nonzero exit / timeout) passes through unchanged; only Err(SandboxError) = infrastructure breakage fails
  closed. The nonzero_exit_is_a_successful_run_... test guards this explicitly — essential, because keying off exit code instead of Err would break every tool that signals findings via
  nonzero exit. Verified correct.
  - Host opt-in preserved: the !is_sandbox_enabled() branch (Native / DISABLE_SANDBOX=true) is untouched and precedes the fail-closed logic — the deliberate, logged host path the issue asks
  for still works.

  Consumer trace — verified end-to-end
  
  - Honest-failure propagation: execute_command.rs and tool wrappers propagate via .await?; tools.rs:639/658 maps a closure Err → ToolResult::error → outcome: ToolOutcome::Failed (asserted
  by the test at :987). So a broken sandbox surfaces as a Failed tool result the operator/model see — not host output posing as sandboxed. The full fail-closed chain holds.
  - execute_command.rs test change is legitimate, not masking: the two unit tests now call set_use_sandbox(false) because CI has no BlackArch backend and (post-#256) an
  enabled-but-unavailable sandbox fails closed with no provenance. Disabling the sandbox is the explicit host path these tests want — same precedent as the pentest-core integration tests.
  Correct adaptation to the new contract.
  - Headless is affected, and intentionally: verified at source — run-pentest.sh and both Dockerfiles set no DISABLE_SANDBOX, so headless keeps USE_SANDBOX=true and runs sandboxed. A
  headless carve-out would reintroduce the exact #256 fail-open, so not exempting it is the right call.

  ⚠️  init-dev-side item to confirm (NOT a defect in this PR — my cross-repo trace rule)
  
  This is the one thing worth verifying on the init-dev side before/around merge, per my consumer-trace practice: init-dev deploys Pick's chart with pick.disableSandbox=true. That behavior 
  only protects init-dev if the chart maps pick.disableSandbox=true → the DISABLE_SANDBOX=true env var on the container. If it does (very likely), init-dev takes the !is_sandbox_enabled()
  host path and is unaffected by this PR — tools keep running in-container as today. If it does not (e.g. the flag only sets a chart-level securityProfile and never exports the env var),
  then after this PR init-dev's headless connector would fail closed on every tool run (its container has no bwrap/proot and relies on the proot download). That mapping lives in the
  init-dev repo, not this Pick checkout, so I can't confirm it from here. Worth a 30-second check of the init-dev pick-connector chart's env: block. (My memory notes init-dev runs
  pick.disableSandbox=true specifically because nmap file-caps need root + proot OOMs there, so the intent is clearly host execution — this is just confirming the wiring delivers it.)

  Other

  - config.rs carries the same TENANT_ENV_LOCK test-flake fix as #286 (not yet merged to main) — identical correct change; the two will converge on merge.

  Bottom line

  The fail-closed logic, the Err-vs-nonzero distinction, and the end-to-end honest-failure propagation are all correct and well-tested. The only follow-through is the init-dev chart env-var
  check — which is a deployment wiring verification on the sibling repo, not a change needed in this PR.

jtomek-strike48:
Rebased onto main after #286 merged. The two touched files (config.rs, execute_command.rs) conflicted only because #286 independently landed the same test hardening (the set_use_sandbox(false) provenance-test fix and the env-lock generalization); resolved by taking main's versions, which are identical in code. #291's own contribution — the fail-closed resolve_enabled_attempt in command.rs and its guard tests — is unchanged. All checks green (Test now inherits DISABLE_SANDBOX=true from main, 7.3 min, no hang).

Issue #289, opened 2026-07-20
Tool target validation hardcodes ValidationMode::Production — ignores PENTEST_ALLOW_PRIVATE_IPS (nikto/ffuf reject private targets in non-prod)


cchao-devo:
> **Related to matrix#3216 (Pick → newdev) / follow-on to Pick #231.** Surfaced during #3216 e2e testing on the newdev k3s cluster (private `10.42.0.0/24`).

## Summary

Per-tool **scan-target** URL validation is hardcoded to `ValidationMode::Production`, so it rejects private-IP targets **even when `PENTEST_ALLOW_PRIVATE_IPS=true`**. Pick #231 relaxed the SSRF guard for the **connector host** (config.rs reads the env → `ValidationMode::PrivateNetwork`), but that opt-in was **not propagated to the individual tools' target validation** — so tools like `nikto`, `ffuf`, etc. still refuse to scan a private-IP target on a non-prod, private-network engagement.

## Evidence

Running a StrikeKit Vulnerability Assessment phase against the k3s cluster, `nikto` on a discovered host failed:

```
$ nikto --port=8180 --timeout=1800 10.42.0.144
Failed: Permission denied: Private IP addresses are not allowed in production mode
```

The connector itself registered fine (proof `PENTEST_ALLOW_PRIVATE_IPS=true` is set and honored for the host), but the tool's target check is a separate path:

- `crates/tools/src/external/nikto.rs:94` — `let target = validate_url(&target, ValidationMode::Production, None)?;`
- `crates/tools/src/external/ffuf.rs:120` — `let _ = validate_url(&base_url, ValidationMode::Production, None)?;`

Both **hardcode `ValidationMode::Production`** rather than deriving it from the same `PENTEST_ALLOW_PRIVATE_IPS` opt-in that `crates/core/src/config.rs` uses for the connector host. The error text is `crates/core/src/url_validation.rs:132`.

## Impact

On any private-network engagement (the newdev k3s cluster is entirely `10.42.0.0/24`), every web-scan target is a private IP, so `nikto`/`ffuf`/other tools that call `validate_url(..., ValidationMode::Production, ...)` **always** fail — the Vulnerability Assessment / web-enumeration phases can't run, despite the environment being explicitly non-prod with private IPs allowed.

## Ask

[...]

Issue #233, opened 2026-07-15
test: cover hostname->private-IP blocking in PrivateNetwork SSRF mode


jtomek-strike48:
## Summary

`ValidationMode::PrivateNetwork` (added in #231) claims a specific security property in its code comments: it intentionally does NOT let a hostname resolve into the LAN-private allow-set, so that a name cannot be used to DNS-rebind into a private range (e.g. the metadata service) even when the private-IP opt-in is on.

There is currently **no test** exercising this. All 7 new tests in #231 use literal IPs, so the DNS-resolution branch of `classify_lan_private` / `validate_url` in `PrivateNetwork` mode is entirely uncovered.

## Why it matters

This gap is what allowed the HIGH finding in #231 to go unnoticed by CI: because only literal ClusterIPs are tested, the fact that a ClusterIP **DNS name** (`connectors-studio-grpc.default.svc`) is rejected in `PrivateNetwork` mode never shows up. A test here protects the security property AND documents the actual (literal-IP-only) behavior.

## Suggested tests (`crates/core/src/url_validation.rs`)

- A hostname resolving to a private RFC-1918 IP is BLOCKED in `PrivateNetwork` mode (`is_err()`), asserting the DNS-rebinding defense. Use `localhost` (resolves to loopback) or gate on DNS availability like the existing `test_production_mode_allows_public_hostnames` does, to stay CI-safe offline.
- A hostname resolving to public IPs is ALLOWED in `PrivateNetwork` mode (falls through to Production path), DNS-gated.
- Optional: assert the `BlockedInternal` error message content for a literal link-local/metadata address, so message regressions are caught.

## Related follow-ups worth folding in

- 172.16/12 exclusive boundary edges (`172.15.255.255`, `172.32.0.0`) blocked in `PrivateNetwork` mode
- CGN `100.64.0.0/10` blocked in `PrivateNetwork` mode (private per full guard, but NOT in the LAN allow-set)
- IPv6 multicast (`ff02::1`) and unspecified (`::`) blocked in `PrivateNetwork` mode
- IPv4-mapped/NAT64 cases once #232 lands

## References

- Discovered during review of #231 (comment: https://github.com/Strike48-public/pick/pull/231#issuecomment-4983566653)
- Related: #232 (IPv6 canonicalization)
- Recommended agent: test-engineer

PR #296, opened 2026-07-21
fix: canonicalize IPv6-encoded IPv4 in SSRF guard; honor private-IP opt-in in tool targets

### LINKED ISSUE #232: SSRF: canonicalize IPv4-mapped and NAT64 IPv6 addresses in url_validation
## Summary

The IPv6 SSRF classifier in `crates/core/src/url_validation.rs` does not canonicalize IPv4 addresses embedded in IPv6 representations, so IPv6-encoded forms of blocked IPv4 targets are classified as public and allowed.

Verified with a standalone probe replicating the classifier logic:

- `::ffff:169.254.169.254` (IPv4-mapped form of the AWS/GCP metadata endpoint) -> classified as public -> allowed
- `::ffff:10.0.0.1` (IPv4-mapped RFC-1918) -> classified as public
- `64:ff9b::a9fe:a9fe` (NAT64 well-known prefix `64:ff9b::/96` embedding `169.254.169.254`) -> classified as public -> allowed

`is_private_ipv6` only checks `fe80::/10`, `fc00::/7`, `ff00::/8`, and `::`; it never extracts the embedded IPv4 to run it through `is_private_ipv4`.

## Scope / severity

This is **pre-existing** behavior, not introduced by PR #231. `is_private_ipv6` is byte-identical on `main`, and the gap affects the default `Production` guard identically. It surfaced during review of #231 because that PR added a `PrivateNetwork` mode whose docstring asserts it "STILL blocks the cloud-metadata / link-local range" - a claim that is incomplete for these IPv6-encoded forms.

Fixing this improves both `Production` and the new `PrivateNetwork` mode.

## Suggested fix

Canonicalize IPv4-mapped addresses (via `Ipv6Addr::to_ipv4_mapped()`) at the top of the IPv6 predicates and route them through the IPv4 classifier. Handle the NAT64 well-known prefix (`64:ff9b::/96`) by extracting the embedded IPv4 and doing the same. Apply consistently to `is_localhost_ipv6`, `is_private_ipv6`, and `is_lan_private_ipv6`.

## Acceptance criteria

- `validate_url("wss://[::ffff:169.254.169.254]:443", Production, None)` returns `Err`
- `validate_url("wss://[::ffff:169.254.169.254]:443", PrivateNetwork, None)` returns `Err`
- `validate_url("wss://[::ffff:10.0.0.1]:443", PrivateNetwork, None)` returns `Ok` (RFC-1918 via mapped form is a LAN-private target)
- NAT64-embedded metadata (`64:ff9b::a9fe:a9fe`) is blocked in both modes
- Tests cover each of the above

## References

- Discovered during review of #231 (comment: https://github.com/Strike48-public/pick/pull/231#issuecomment-4983566653)
- Recommended agent: security-reviewer

### LINKED ISSUE #236: SSRF guard: canonicalize IPv6-encoded IPv4 in is_private_ipv6 (IPv4-mapped / NAT64 metadata bypass)
## Summary

`is_private_ipv6` (crates/core/src/url_validation.rs) does not canonicalize IPv6-encoded IPv4 addresses, so IPv4-mapped and NAT64 forms of the cloud-metadata / link-local range classify as **public** and are not blocked by the SSRF guard.

- IPv4-mapped: `::ffff:169.254.169.254`
- NAT64: `64:ff9b::a9fe:a9fe`

Both resolve to the AWS/GCP metadata service (`169.254.169.254`) but bypass the IPv6 private-range checks.

## Scope / severity

- **Pre-existing** on `main` — affects the default **Production** guard identically. Confirmed `is_private_ipv6` is byte-identical to `main`; this is **not** a regression introduced by PR #231.
- Surfaced during review of #231 (which added `ValidationMode::PrivateNetwork`). The PrivateNetwork docstring was softened there to note this caveat rather than overclaim "still blocks metadata"; this issue tracks the actual fix.
- Practical exploitability is limited (requires an attacker-influenced host reaching the connector-host validation, which the operator sets), but it's a real gap in a defense-in-depth SSRF control.

## Suggested fix

In `is_private_ipv6`, canonicalize embedded IPv4 before classifying:
- IPv4-mapped `::ffff:a.b.c.d` — extract the trailing IPv4 and run it through `is_private_ipv4` (Rust: `Ipv6Addr::to_ipv4_mapped()`).
- NAT64 `64:ff9b::/96` — extract the embedded IPv4 and run `is_private_ipv4`.

This benefits **Production** mode too, not just `PrivateNetwork`. Add tests for both encoded forms of `169.254.169.254` asserting they're blocked in Production and not admitted in PrivateNetwork.

## Context

- PR: #231 (fix/allow-private-ips-env)
- File: `crates/core/src/url_validation.rs` — `is_private_ipv6`

jtomek-strike48:
## Summary

Closes #232, closes #236, addresses #289, and adds the #233 test coverage.

Two related SSRF-guard gaps in `crates/core/src/url_validation.rs` plus the per-tool target-validation propagation:

### 1. IPv6-encoded IPv4 bypass (#232 / #236)

`is_private_ipv6` never canonicalized IPv4 addresses embedded in IPv6 forms, so IPv6-encoded forms of blocked IPv4 targets classified as **public** and were allowed:

- IPv4-mapped `::ffff:169.254.169.254` - dual-stack sockets route mapped addresses straight to the v4 metadata service
- NAT64 `64:ff9b::a9fe:a9fe` (well-known prefix `64:ff9b::/96`, RFC 6052) - a NAT64 gateway translates to the same `169.254.169.254`

New `embedded_ipv4()` extracts the IPv4 (via `Ipv6Addr::to_ipv4_mapped()` for mapped, explicit prefix match for NAT64) and routes it through the existing IPv4 classifier. Applied consistently to `is_localhost_ipv6`, `is_private_ipv6`, and `is_lan_private_ipv6`, so it hardens both the default **Production** guard and the **PrivateNetwork** mode.

### 2. Per-tool target validation ignored the private-IP opt-in (#289)

`nikto`/`ffuf`/`dirb` hardcoded `ValidationMode::Production`, so they rejected private-IP scan targets even with `PENTEST_ALLOW_PRIVATE_IPS=true` - blocking vulnerability-assessment phases on a legitimately non-prod private-network engagement. `gobuster` used a separate `PENTEST_ENV` var that defaulted to `Development` (allow-all) - a weaker, inconsistent posture.

Promoted `config.rs`'s `resolve_validation_mode` into `url_validation` as the single shared, secure-by-default source of truth (`target_validation_mode()`), used at every tool call site; `config.rs` now delegates to it.

## Behavior change

`gobuster` dir/vhost target validation now defaults to **Production** (private IPs blocked) instead of Development (allow-all), honoring `PENTEST_ALLOW_PRIVATE_IPS` like the other tools. `PENTEST_ENV` was read nowhere else.

## Test plan

- [x] 12 new `url_validation` unit tests (encoded metadata blocked in Production + PrivateNetwork, mapped RFC-1918 admitted in PrivateNetwork, DNS-rebinding property, CGN/benchmark/doc/multicast/unspecified reserved ranges, 172.16/12 exclusive edges, shared resolver)
- [x] Mutation-verified per Lens 1: disabling `embedded_ipv4()` turns the 5 metadata-blocking tests RED; restoring makes them GREEN
- [x] `cargo test -p pentest-core --lib url_validation` - 45 passed
- [x] `cargo test -p pentest-tools --test security_tests` - 55 passed
- [x] config's `resolve_validation_mode` tests still pass (delegation preserved)
- [x] `cargo clippy -p pentest-core -p pentest-tools --all-targets -- -D warnings` - clean
- [x] `cargo fmt` - clean

## Notes for review

- #233 was filed against an earlier revision of #231 that *blocked all hostnames* in PrivateNetwork. The shipped behavior deliberately *allows* a hostname resolving entirely to LAN-private IPs (the documented in-cluster `.svc` case, guarded by `test_private_network_mode_allows_hostname_resolving_to_private`). The new tests lock in the *actual* rebinding defense: admission requires every resolved IP to be LAN-private, so a name resolving to metadata hits the `BlockedInternal` arm.
- Exotic IPv4 encodings (decimal/octal/hex) are not a bypass - Rust's `IpAddr` parser rejects them.

### DIFF
--- a/crates/core/src/config.rs
+++ b/crates/core/src/config.rs
@@ -215,36 +215,13 @@ impl ConnectorConfig {
     /// Secure by default: any other value (including unset) falls back to
     /// [`ValidationMode::default()`], which is Production in release builds.
     /// Pure (no I/O) so it can be unit-tested independent of the build profile.
+    ///
+    /// Delegates to [`crate::url_validation::resolve_validation_mode`], the
+    /// single source of truth now shared with per-tool target validation
+    /// (`nikto`/`ffuf`/`dirb`/`gobuster`). Kept as an associated function so the
+    /// connector-host call site and its tests read naturally.
     fn resolve_validation_mode(env_val: Option<&str>) -> crate::url_validation::ValidationMode {
-        use crate::url_validation::ValidationMode;
-
-        match env_val.map(|v| v.trim().to_ascii_lowercase()) {
-            Some(v) if v == "true" || v == "1" => {
-                // Loud, auditable: an operator (or a leaked env var) has relaxed
-                // the SSRF guard. Emitted once per validate() call, not per request.
-                tracing::warn!(
-                    "PENTEST_ALLOW_PRIVATE_IPS is set: connector host SSRF validation \
-                     relaxed to PrivateNetwork mode (RFC-1918/loopback allowed; \
-                     link-local/metadata still blocked). Do NOT use in production."
-                );
-                ValidationMode::PrivateNetwork
-            }
-            // Set, but NOT a recognized truthy value (e.g. a typo like "ture",
-            // or "yes"/"on"/"0"/"false"). Fall back to the secure default, but
-            // WARN so the operator isn't left debugging a "private IP blocked"
-            // failure with no hint that their opt-in was silently ignored.
-            Some(v) if !v.is_empty() => {
-                tracing::warn!(
-                    "PENTEST_ALLOW_PRIVATE_IPS is set to an unrecognized value ({:?}); \
-                     expected \"true\" or \"1\". Ignoring and using the secure default \
-                     (private IPs blocked in release builds).",
-                    v
-                );
-                ValidationMode::default()
-            }
-            // Unset or empty: secure default, silently.
-            _ => ValidationMode::default(),
-        }
+        crate::url_validation::resolve_validation_mode(env_val)
     }
 
     /// Check if this config has authentication
--- a/crates/core/src/url_validation.rs
+++ b/crates/core/src/url_validation.rs
@@ -290,14 +290,19 @@ fn is_localhost_ipv6(ip: Ipv6Addr) -> bool {
 /// Extract an IPv4 address embedded in an IPv6 representation, if any.
 ///
 /// SSRF defenses must not be bypassable by re-encoding a blocked IPv4 target as
-/// IPv6. Four encodings smuggle an IPv4 address through the IPv6 classifier:
+/// IPv6. Six encodings smuggle an IPv4 address through the IPv6 classifier:
 ///
 /// * **IPv4-mapped** `::ffff:a.b.c.d` (`::ffff:0:0/96`) — dual-stack sockets
 ///   route these straight to the v4 address, so `::ffff:169.254.169.254`
 ///   reaches the real cloud-metadata service.
+/// * **IPv4-translated** `::ffff:0:a.b.c.d` (`::ffff:0:0:0/96`, RFC 2765 SIIT) —
+///   the mapped form's sibling, one segment further left. Trivially confused
+///   with the mapped prefix, and a SIIT translator resolves it to the v4.
 /// * **NAT64** `64:ff9b::a.b.c.d` (well-known prefix `64:ff9b::/96`, RFC 6052) —
 ///   a NAT64 gateway translates these to the embedded v4 destination.
-///
+/// * **NAT64 local-use** `64:ff9b:1::a.b.c.d` (`64:ff9b:1::/48`, RFC 8215) —
+///   the range reserved for deployment-chosen NAT64 prefixes. A deployment that
+///   carves a `/96` out of it translates the low 32 bits exactly as above.
 /// * **IPv4-compatible** `::a.b.c.d` (top 96 bits zero) — the deprecated
 ///   (RFC 4291) sibling of the mapped form. Formally not auto-routed to IPv4
 ///   everywhere, but a defense-in-depth SSRF guard must not let `::169.254.169.254`
@@ -306,17 +311,42 @@ fn is_localhost_ipv6(ip: Ipv6Addr) -> bool {
 ///   segments 1..3; a 6to4 relay translates to the embedded destination.
 ///
 /// All carry the IPv4 in a fixed slot. `Ipv6Addr::to_ipv4_mapped()` handles the
-/// mapped form; the NAT64, IPv4-compatible, and 6to4 forms are matched
-/// explicitly. `::` (unspecified) and `::1` (loopback) are the two IPv4-compatible
-/// literals that must NOT be reinterpreted as embedded IPv4 — they have dedicated
-/// predicates (and `::1` must classify as loopback, not slip through the
-/// `is_lan_private_ipv4` allow path as `0.0.0.1`), so they are excluded here.
+/// mapped form; the other five are matched explicitly. `::` (unspecified) and
+/// `::1` (loopback) are the two IPv4-compatible literals that must NOT be
+/// reinterpreted as embedded IPv4 — they have dedicated predicates (and `::1`
+/// must classify as loopback, not slip through the `is_lan_private_ipv4` allow
+/// path as `0.0.0.1`), so they are excluded here.
+///
+/// **Known uncovered:** RFC 6052 also defines *variable-length* embeddings for
+/// network-specific prefixes of `/32`, `/40`, `/48`, `/56`, and `/64`, which
+/// scatter the v4 octets around a zero `u` byte at bits 64-71 instead of using
+/// the low 32 bits. Decoding those requires knowing the deployment's own NSP,
+/// which this crate has no way to learn, so only the low-32 layout is decoded.
+/// An operator behind a NAT64 gateway using a non-`/96` NSP can still reach a
+/// blocked v4 by hand-encoding it; that is an accepted limit of a
+/// prefix-agnostic guard, not an oversight.
 fn embedded_ipv4(ip: Ipv6Addr) -> Option<Ipv4Addr> {
+    /// Decode an IPv4 address from two adjacent IPv6 segments (big-endian).
+    fn v4_from(hi: u16, lo: u16) -> Ipv4Addr {
+        Ipv4Addr::new(
+            (hi >> 8) as u8,
+            (hi & 0xff) as u8,
+            (lo >> 8) as u8,
+            (lo & 0xff) as u8,
+        )
+    }
+
     // IPv4-mapped ::ffff:a.b.c.d
     if let Some(v4) = ip.to_ipv4_mapped() {
         return Some(v4);
     }
     let seg = ip.segments();
+    let low32 = v4_from(seg[6], seg[7]);
+    // IPv4-translated ::ffff:0:a.b.c.d (::ffff:0:0:0/96, RFC 2765): the 0xffff
+    // sits one segment left of the mapped form, low 32 bits carry the IPv4.
+    if seg[0] == 0 && seg[1] == 0 && seg[2] == 0 && seg[3] == 0 && seg[4] == 0xffff && seg[5] == 0 {
+        return Some(low32);
+    }
     // NAT64 well-known prefix 64:ff9b::/96 (RFC 6052): first 6 segments are
     // 0064:ff9b:0000:0000:0000:0000, low 32 bits carry the IPv4.
     if seg[0] == 0x0064
@@ -326,36 +356,27 @@ fn embedded_ipv4(ip: Ipv6Addr) -> Option<Ipv4Addr> {
         && seg[4] == 0
         && seg[5] == 0
     {
-        return Some(Ipv4Addr::new(
-            (seg[6] >> 8) as u8,
-            (seg[6] & 0xff) as u8,
-            (seg[7] >> 8) as u8,
-            (seg[7] & 0xff) as u8,
-        ));
+        return Some(low32);
+    }
+    // NAT64 local-use prefix 64:ff9b:1::/48 (RFC 8215). RFC 8215 reserves the
+    // whole /48 for NAT64, so no legitimate global destination lives here and a
+    // guard may decode the low 32 without inspecting segments 3..6 — erring
+    // toward blocking inside a reserved range is the safe direction.
+    if seg[0] == 0x0064 && seg[1] == 0xff9b && seg[2] == 0x0001 {
+        return Some(low32);
     }
     // IPv4-compatible ::a.b.c.d (top 96 bits zero), excluding :: and ::1 which
     // their own predicates handle. `to_ipv4()` also matches the mapped form, but
     // that already returned above, so a match here is the compatible form.
     if seg[0] == 0 && seg[1] == 0 && seg[2] == 0 && seg[3] == 0 && seg[4] == 0 && seg[5] == 0 {
-        let v4 = Ipv4Addr::new(
-            (seg[6] >> 8) as u8,
-            (seg[6] & 0xff) as u8,
-            (seg[7] >> 8) as u8,
-            (seg[7] & 0xff) as u8,
-        );
         // Exclude ::/::1 (0.0.0.0 / 0.0.0.1): unspecified + loopback.
-        if v4 != Ipv4Addr::UNSPECIFIED && v4 != Ipv4Addr::new(0, 0, 0, 1) {
-            return Some(v4);
+        if low32 != Ipv4Addr::UNSPECIFIED && low32 != Ipv4Addr::new(0, 0, 0, 1) {
+            return Some(low32);
         }
     }
     // 6to4 2002:a.b.c.d::/48 (RFC 3056): v4 in segments[1..3].
     if seg[0] == 0x2002 {
-        return Some(Ipv4Addr::new(
-            (seg[1] >> 8) as u8,
-            (seg[1] & 0xff) as u8,
-            (seg[2] >> 8) as u8,
-            (seg[2] & 0xff) as u8,
-        ));
+        return Some(v4_from(seg[1], seg[2]));
     }
     None
 }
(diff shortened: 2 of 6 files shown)


jtomek-strike48:
## Disclosure after self-review pass: gobuster breaking change (Lens 8)

An adversarial review flagged an undocumented behavioral change I should call out explicitly:

Gobuster's dir/vhost target validation previously read a separate `PENTEST_ENV` variable and **defaulted to Development (allow-all)** when unset. This PR unifies it onto the shared `PENTEST_ALLOW_PRIVATE_IPS` opt-in, which **defaults to Production (private IPs blocked)**.

**Net effect:** a deployment/script that scanned private-IP targets via **gobuster** without an opt-in — relying on the old `PENTEST_ENV`-defaulted allow-all — will now be refused ("Private IP addresses are not allowed in production mode"). This is intentional: it removes a weaker, inconsistent SSRF posture and matches how the connector host + nikto/ffuf/dirb already behave. But it is a behavioral change worth surfacing.

**Migration:** set `PENTEST_ALLOW_PRIVATE_IPS=true` to restore private-IP scanning via gobuster. `PENTEST_ENV` is no longer read.

cchao-devo (changes requested):
Review: Pick PR #296 — canonicalize IPv6-encoded IPv4 in SSRF guard; honor private-IP opt-in in tool targets

  Verdict: 🔧 Request changes (one HIGH, one MEDIUM — verified by running the validator). The two targeted vectors (IPv4-mapped, NAT64 /96) are correctly fixed, the shared-mode refactor is
  clean, and the gobuster tightening is right. But the fix is incomplete for the re-encoding class it set out to close: a third IPv6 encoding of the same metadata address is still
  allowed. Small, localized fix.

  Lenses run: the real pr-review-toolkit plugin agents — silent-failure-hunter (SSRF-bypass lead, ran the classifier) + code-reviewer (bit-math + refactor + gates) — plus my own
  independent execution of the validator against the encoded-metadata battery (the CLAUDE.md "run, don't read" mandate — decisive here). Verified in pick_08/pick; pick_07 untouched, no
  cluster.

  🔴 HIGH — IPv4-compatible ::169.254.169.254 reaches metadata (verified ALLOWED in Production)
  
  I ran validate_url on the exact hosts. ::169.254.169.254 (and its hex spelling ::a9fe:a9fe) classify as public → ALLOWED in both Production and PrivateNetwork, while the mapped
  (::ffff:…) and NAT64 (64:ff9b::…) forms of the identical address are correctly blocked. Mechanism: embedded_ipv4 uses to_ipv4_mapped() (matches only ::ffff:0:0/96) + the NAT64 /96
  prefix, so an IPv4-compatible ::a.b.c.d returns None → is_private_ipv6 checks only fe80/fc00/ff00/:: → returns false → Public. The PR's comment justifying the exclusion ("only a concern
  for ::1/::") is wrong in the under-blocking direction — it's also the concern for every other embedded address. This is the exact bug class the PR exists to close, shipped inconsistently
  (2 of 3 encodings of 169.254.169.254 blocked). It also defeats the DNS-rebinding guard for an AAAA record resolving to the compatible form.

  Fix (small): extend embedded_ipv4 to also extract the low-32 when the top 96 bits are zero, excluding :: and ::1 (which their own predicates handle) — the panel's suggested guard. Then
  re-run the probe: ::169.254.169.254 → blocked, ::10.0.0.1 → LAN-private (allowed in PrivateNetwork, blocked in Production).

  🟠 MEDIUM — 6to4 2002:a9fe:a9fe:: metadata allowed (verified ALLOWED in Production)

  Same class, different prefix (2002::/16 embeds the v4 in segments[1..3]). Lower real-world routability (6to4 relays deprecated, RFC 7526), but the same "re-encode a blocked v4" vector
  the fix targets and the fix doesn't touch it. Either decode it in embedded_ipv4 (segments[0]==0x2002 → v4 from segments[1..3]) or explicitly document it as a known-uncovered vector —
  don't leave it silently absent.

  Severity calibration (honest): IPv4-compatible and 6to4 addresses are formally deprecated and not universally auto-routed to IPv4 like mapped addresses are, so exploitability is
  environment-dependent and lower than the mapped/NAT64 forms the PR fixed. But this is a defense-in-depth control whose docstring explicitly promises it can't be bypassed by IPv6
  re-encoding, and it's a few lines to close — so I'd fix the compatible form before merge (HIGH) and at minimum document 6to4 (MEDIUM).

  What's correct (verified by execution + bit-math)

  - Mapped + NAT64/96 fixed: ::ffff:169.254.169.254, ::ffff:a9fe:a9fe (hex), 64:ff9b::a9fe:a9fe all blocked in both modes; mapped RFC-1918 ::ffff:10.0.0.1 correctly allowed in
  PrivateNetwork, blocked in Production. embedded_ipv4 bit-math verified (64:ff9b::a9fe:a9fe → 169.254.169.254). Applied consistently to all three IPv6 predicates.
  - DNS-rebinding all-IPs-must-be-private: correct — the resolver loop returns the first non-LAN IP (→ BlockedInternal), only promoting to LanPrivate if every resolved IP is LAN-private.
  (Inherits the compat-form gap for AAAA, closed by the HIGH fix.)
  - Shared-mode refactor: config.rs delegates to url_validation::resolve_validation_mode (logic byte-identical, secure-by-default preserved: only true/1 → PrivateNetwork, never
  Development). All 4 tools use target_validation_mode(); gobuster's weaker PENTEST_ENV path fully removed (grep-confirmed). Gobuster Development→Production tightening correctly disclosed.
  - Gates: clippy clean, url_validation 45/0, security_tests 55/0, config 37/0.

  Notes (non-blocking)
  
  - AWS IMDSv6 fd00:ec2::254: blocked in Production (ULA fc00::/7), allowed in PrivateNetwork — by design (PrivateNetwork admits all ULA for in-cluster targets; can't distinguish AWS's
  IMDSv6 ULA without hardcoding it). Worth a docstring note that PENTEST_ALLOW_PRIVATE_IPS=true on AWS exposes IMDSv6.
  - zap.rs has its own local validate_url (shell-metachar only, no SSRF IP check) — pre-existing separate path, not touched here; the "all web tools share one source of truth" framing
  shouldn't be over-read.
  - Debug builds: dirb/ffuf/nikto now inherit Development default in debug (were hardcoded Production); release is Production. Dev-only, and now consistent with the connector host.

  init-dev relevance (my cross-repo memory)

  Directly connected. This is the #289/PrivateNetwork work my memory tracks — the .svc DNS-rebinding case. The DNS-rebinding all-IPs-must-be-private semantics are correctly implemented,
  which is the property init-dev's .svc in-cluster targets rely on. Good. (The IPv6-compat bypass doesn't change init-dev's IPv4 .svc flow, but the fix should still land for the
  metadata-protection guarantee.)

jtomek-strike48:
Pushed `fc247c8` addressing both findings.

## HIGH (IPv4-compatible) + MEDIUM (6to4) — fixed

Extended `embedded_ipv4` to decode two more encodings of the same address, so all four spellings of `169.254.169.254` now canonicalize identically:

- **IPv4-compatible `::a.b.c.d`** (top 96 bits zero) — decoded, excluding `::` and `::1`. The exclusion is load-bearing: `::1` must keep classifying as loopback via its own predicate, not slip through the `is_lan_private_ipv4` allow path as `0.0.0.1` (its dedicated `ip == LOCALHOST` check at line 530 runs *after* the `embedded_ipv4` branch, so decoding `::1` there would shadow it).
- **6to4 `2002:a.b.c.d::/48`** — v4 decoded from segments 1..3.

Since all three IPv6 predicates share the helper, the DNS-rebinding guard inherits the coverage for AAAA records automatically. I also corrected the docstring — its prior "only a concern for `::1`/`::`" reasoning was wrong in the under-blocking direction, exactly as you noted.

## Verification (run, not read)

- **Guard-check (decisive):** neutered the two new branches -> `test_production_blocks_ipv4_compatible_and_6to4_metadata` goes RED with `wss://[::169.254.169.254]:443 must be BLOCKED in Production` (url_validation.rs:1200). Restored -> green. So the test genuinely guards the bypass.
- `::169.254.169.254`, `::a9fe:a9fe` (hex), and `2002:a9fe:a9fe::` are now blocked in **both** Production and PrivateNetwork; `::10.0.0.1` decodes to LAN-private (allowed in PrivateNetwork, blocked in Production) — matching the mapped/NAT64 behavior you verified.
- `::` and `::1` still return `None` from `embedded_ipv4` (assertions added).
- `cargo fmt --check` + `clippy -p pentest-core --lib` (`-D warnings`) clean; 42 `url_validation` lib tests pass.

Tests added: extended `test_embedded_ipv4_extraction` (compatible/6to4 vectors + `::`/`::1` exclusions) and a new `test_production_blocks_ipv4_compatible_and_6to4_metadata` covering both modes.

Ready for another look.

jtomek-strike48:
Both findings are fixed, and probing the class turned up two more encodings that I've also closed. Rebased onto `63d511f` (was 6 behind, so the previous green CI hadn't tested against current base).

## Your two findings

`fc247c8` (pushed before this round, after your review) closed both. Verified by execution, not by reading the diff:

| host | Production | PrivateNetwork |
|---|---|---|
| `::169.254.169.254` | BLOCKED | BLOCKED |
| `::a9fe:a9fe` | BLOCKED | BLOCKED |
| `2002:a9fe:a9fe::` | BLOCKED | BLOCKED |

Guard-checked per your "run, don't read" point: replacing both new branch conditions with `if false` turns `test_embedded_ipv4_extraction` and `test_production_blocks_ipv4_compatible_and_6to4_metadata` red (`44 passed; 2 failed`). Restored, green. The `::`/`::1` exclusion is in place so `::1` still classifies as loopback rather than slipping through the `is_lan_private_ipv4` allow path as `0.0.0.1`.

## Two more in the same class (new, this round)

Running the battery past the vectors you named, two others still reached metadata:

```
[::ffff:0:169.254.169.254]     prod=ALLOWED  private=ALLOWED   IPv4-translated ::ffff:0:0:0/96 (RFC 2765)
[64:ff9b:1::169.254.169.254]   prod=ALLOWED  private=ALLOWED   NAT64 local-use 64:ff9b:1::/48 (RFC 8215)
```

Both embed the v4 in the low 32 bits exactly as the mapped and well-known NAT64 forms do; `embedded_ipv4` just didn't match their prefixes. The local-use one is the more reachable of the two and I'd rate it above the 6to4 finding on exploitability: RFC 8215 reserves that `/48` specifically for deployment-chosen NAT64 prefixes, so a gateway carving a `/96` out of it translates to the embedded destination the same way the well-known prefix does. No deprecation caveat applies.

`629b5f7` decodes both. Post-fix, all six encodings block and the controls still pass:

```
169.254.169.254                  prod=BLOCKED  private=BLOCKED  bare v4
[::ffff:169.254.169.254]         prod=BLOCKED  private=BLOCKED  IPv4-mapped
[::ffff:0:169.254.169.254]       prod=BLOCKED  private=BLOCKED  IPv4-translated /96
[64:ff9b::169.254.169.254]       prod=BLOCKED  private=BLOCKED  NAT64 well-known /96
[64:ff9b:1::169.254.169.254]     prod=BLOCKED  private=BLOCKED  NAT64 local-use /48
[::169.254.169.254]              prod=BLOCKED  private=BLOCKED  IPv4-compatible
[2002:a9fe:a9fe::]               prod=BLOCKED  private=BLOCKED  6to4
[64:ff9b:2::a9fe:a9fe]           prod=ALLOWED  private=ALLOWED  control: outside reserved ranges
[2001:db8::a9fe:a9fe]            prod=ALLOWED  private=ALLOWED  control: global v6
example.com                      prod=ALLOWED  private=ALLOWED  control: public hostname
```

Guard-checked the same way: neutering both new branches turns all three new tests red; restored, green.

One design note worth your eye. The local-use branch checks only the `/48`, not a specific `/96` inside it. Since RFC 8215 reserves the entire range for NAT64, no legitimate global destination lives there, so erring toward blocking inside it is the safe direction — but it is deliberately broader than the other branches. `64:ff9b:2::` is outside both ranges and has a test asserting it does *not* decode.

I also added `test_translated_and_local_use_nat64_honor_private_opt_in`, which pins that an embedded RFC-1918 address stays allowed in `PrivateNetwork` and blocked in `Production`. Without it, a future change that blanket-blocked these prefixes would pass the metadata test while quietly breaking the in-cluster `.svc` targets you flagged as the init-dev dependency.

## Known-uncovered, now documented

RFC 6052 also defines variable-length embeddings for network-specific prefixes (`/32`, `/40`, `/48`, `/56`, `/64`) that scatter the v4 octets around the zero `u` byte at bits 64-71 instead of using the low 32 bits. Decoding those requires knowing the deployment's own NSP, which this crate has no way to learn. That's an accepted limit of a prefix-agnostic guard, and the docstring now says so — previously it enumerated its coverage as if exhaustive, which is the same overstatement you flagged in the original `::1`/`::` reasoning.

## Gates

`cargo fmt --all --check` clean. `cargo clippy --workspace --locked --features pentest-platform/desktop-pcap -- -D warnings` clean. `cargo test -p pentest-core --lib` 287 passed, 0 failed. All re-run after the rebase, not before it.

Not addressed from your non-blocking notes: the `PENTEST_ALLOW_PRIVATE_IPS`-on-AWS IMDSv6 docstring note and `zap.rs`'s separate local `validate_url`. Say the word if you want either in this PR rather than a follow-up.

jtomek-strike48:
Following up on your two non-blocking notes rather than letting them drop:

**`zap.rs`'s separate `validate_url`** -- verified and filed as #311 (P2). Confirmed it's scheme + shell-metacharacter only, no IP check, and its own test at `zap.rs:264` pins `http://10.0.0.5:8080/` as accepted. `ZapTool` is registered unconditionally with `target` as a required LLM-supplied string, so `http://169.254.169.254/latest/meta-data/` passes every check in that function.

While auditing I checked every `external/` tool taking a `target`, and scoped two categories out: `burp` takes a target but only formats instructional text (never fetches), and `nmap`/`masscan`/`rustscan`/`hydra`/`enum4linux` are LAN scanners where a URL-shaped SSRF guard is the wrong control. `zap` is the only *fetching* web tool missing the shared guard, which is what makes it the real inconsistency -- your "shouldn't be over-read" caveat was right, and it's now narrower than five-of-five.

Noted in #311 that it should land after this PR, since it wants `target_validation_mode()` rather than a hardcoded `Production` that then needs reworking.

**RFC 6052 variable-length NSP embeddings** -- filed as #310 (P3) with the reasoning for leaving it documented rather than decoded: those layouts scatter the octets around the zero `u` byte and can't be decoded without knowing the deployment's own prefix. A heuristic that tried all five lengths would over-block legitimate global addresses, which is the wrong failure direction.

Still open from your notes: the IMDSv6 docstring caveat for `PENTEST_ALLOW_PRIVATE_IPS` on AWS. That one is a comment-only change and I'm happy to fold it into this PR if you'd rather not have a separate round -- your call.

jtomek-strike48:
Thanks for the empirical review. The two flagged vectors are addressed in commits 996f9ab (IPv4-compatible `::a.b.c.d` + 6to4 `2002::/16`) and 629b5f7 (also IPv4-translated `::ffff:0:0:0/96` and NAT64 local-use `64:ff9b:1::/48`). `embedded_ipv4` now decodes all six embeddings; the variable-length RFC 6052 NSP layouts are documented as a known/accepted limit rather than left silent.

Guard-checked: neutering the decode branches turns `test_production_blocks_ipv4_compatible_and_6to4_metadata` red on `::169.254.169.254`; restoring it goes green. url_validation 48/0, security_tests 55/0, CI green. Ready for another look.

cchao-devo:
@jtomek-strike48, Apporved, thx for the great work.

  Re-reviewed at 629b5f7. Both findings are fixed, and I confirmed it by re-running the classifier myself (not reading the diff) — every metadata encoding is now blocked in
  both modes, with the decoder extracting the correct IPv4 in each case:

  ┌──────────────────────────────────────────────────┬────────────┬────────────────┐
  │                     encoding                     │ Production │ PrivateNetwork │
  ├──────────────────────────────────────────────────┼────────────┼────────────────┤
  │ ::169.254.169.254 (IPv4-compatible — my HIGH)    │ BLOCKED    │ BLOCKED        │
  ├──────────────────────────────────────────────────┼────────────┼────────────────┤
  │ 2002:a9fe:a9fe:: (6to4 — my MEDIUM)              │ BLOCKED    │ BLOCKED        │
  ├──────────────────────────────────────────────────┼────────────┼────────────────┤
  │ ::ffff:0:169.254.169.254 (IPv4-translated /96)   │ BLOCKED    │ BLOCKED        │
  ├──────────────────────────────────────────────────┼────────────┼────────────────┤
  │ 64:ff9b:1::169.254.169.254 (NAT64 local-use /48) │ BLOCKED    │ BLOCKED        │
  └──────────────────────────────────────────────────┴────────────┴────────────────┘

  Nice adversarial follow-through finding the extra two in the same class.

  I specifically stress-tested the NAT64 local-use /48 decode for over-blocking (my concern was it might extract the wrong octets for that layout): 64:ff9b:1::8.8.8.8
  correctly extracts 8.8.8.8 and stays ALLOWED, so the decode is right and legitimate global addresses through that prefix aren't falsely blocked. Also verified ::1/:: still
  return None (loopback/unspecified handled by their own predicates), and global/RFC1918 controls behave correctly — no over-blocking.

  Gates on my end: cargo fmt --check clean, clippy --workspace --all-targets -D warnings clean, url_validation 48/0, security_tests 55/0.

  The #310 (variable-length NSP) and #311 (zap.rs shared-guard, land after this) dispositions both look right — agree with documenting the NSP layouts rather than decoding
  them (a length-guessing heuristic would over-block global addresses, the wrong failure direction).

  LGTM to merge. The only open thread is the optional IMDSv6 docstring caveat — happy either way; fold it in now or take it as a follow-up, your call.

cchao-devo (approved):
LGTM.

Issue #208, opened 2026-07-06
Improve OS incompatibility error messages in tool execution


jtomek-strike48:
# Improve OS incompatibility error messages in tool execution

**Issue Type:** Enhancement
**Component:** Pick agent (`crates/core/src/tools.rs`)
**Effort:** Small
**Priority:** P3
**Parent:** #183 (Child A follow-up)
**Source:** PR #196 comprehensive review - UX improvement

## Problem Statement

When a user requests a tool that exists but is not supported on the current OS (e.g., `autopwn_crack` on Windows), the error handler returns a generic "Tool not found" message followed by a list of available tools. While the available-tools list is correctly OS-filtered, the error doesn't explain *why* the tool is unavailable (OS incompatibility vs. typo).

**Current behavior:**
```
Tool 'autopwn_crack' not found in registry

Available tools:
  - dns_enum
  - network_scan
  ...
```

**Desired behavior:**
```
Tool 'autopwn_crack' exists but is not supported on this OS
Required OS: [Linux]
Current OS: Windows

Available tools:
  - dns_enum
  - network_scan
  ...
```

## Acceptance Criteria

- [ ] Error handler distinguishes between "tool doesn't exist" and "tool exists but unsupported on this OS"
- [ ] OS-incompatible error shows required OS and current OS
- [ ] "Available tools" list remains OS-filtered (current behavior preserved)
- [ ] Error messages tested (add test case for unsupported tool execution)

## Implementation Notes

**Location:** `crates/core/src/tools.rs:778-796` (ToolRegistry::execute error handler)

**Suggested implementation:**
```rust
None => {
    // Check if tool exists but is unsupported
    if self.tools.contains_key(name) && !self.tools.get(name).unwrap().is_supported() {
        tracing::error!("Tool '{}' exists but is not supported on this OS", name);
        tracing::error!("Required OS: {:?}", self.tools.get(name).unwrap().supported_os());
        tracing::error!("Current OS: {:?}", DesktopOs::current());
    } else {
        tracing::error!("Tool '{}' not found in registry", name);
    }

    tracing::error!("");
    tracing::error!("Available tools:");
    let mut names: Vec<&str> = self.supported_names();
    // ... existing available-tools logic
}
```

**Testing:**
- Add test case verifying error message for unsupported tool
- Verify "tool doesn't exist" path still works correctly

## References

- PR #196 review findings - silent-failure-hunter identified this UX gap
- `crates/core/src/tools.rs:778-796` - error handler location

[...]

Issue #202, opened 2026-07-04
Child B1: fail loudly when platform-capability fallbacks are compiled out (network interfaces, mDNS)


jtomek-strike48:
**Parent epic:** #183. **Child:** B (slice B1).

## Problem

Two feature-gated fallback paths return a silent empty on non-default builds instead of failing loudly:

- `crates/platform/src/desktop/system.rs` (~:146) — `get_network_interfaces` fallback, when the `network-interface` feature is off, returns `Ok(Vec::new())` on non-Linux. An empty interface list flows into `check_wifi_connection_status` -> `total_adapters=0`, `connected_via_wifi=false`, `safe_to_scan=true`: a real (wrong) negative the agent trusts.
- `crates/platform/src/desktop/network.rs:219` — `mdns_discover` fallback (feature off) returns `Ok(Vec::new())`, which reads as "no services found." Its sibling `ssdp_discover` (`network.rs:197`) already returns `Err(Error::PlatformNotSupported(...))` — this makes mDNS match that established, honest pattern.

Both fallbacks are only reachable in a `--no-default-features`-style build; the default `desktop` feature ships the real cross-platform impls (`network-interface`, `mdns-sd`). So this is not a shipping-Windows regression — it is the epic's "fail loudly, no silent empties" principle applied to the misconfigured-build path.

## Fix

Return `Err(Error::PlatformNotSupported(...))` from both fallback arms, mirroring `ssdp_discover`. No change to the default build's behavior.

## Acceptance criteria

- [ ] Compiled-out network-interface fallback returns a typed error, not `Ok(vec![])`.
- [ ] `mdns_discover` fallback returns `PlatformNotSupported`, matching `ssdp_discover`.
- [ ] Default `desktop` build behavior unchanged.
- [ ] Test coverage for the typed-error arms where feasible.

**Recommended agent:** rust-reviewer

---
Sibling slices (scoped, not in this PR): **B2** real Windows DNS resolver enumeration in `safety_check/threat_intel.rs` (low priority, best-effort `Vec`); **B3** nmap raw-socket privilege probe in `external/nmap.rs` — that code currently lives only on the #141 branch, so B3 sequences after #141.

Issue #198, opened 2026-07-04
Child A follow-up: log when a tool is filtered from the advertised set


jtomek-strike48:
**Parent epic:** #183. Follow-up from the `/review-pr` pass on #196 (Child A).

## Problem

`ToolRegistry::supported_schemas()` / `supported_names()` filter with `.filter(|t| t.is_supported())`. When a tool is filtered out (misconfigured `supported_os`, or an exotic-target `DesktopOs::current()`), it silently disappears from the advertised set with no diagnostic. Given the epic's "fail loudly, no silent empties" philosophy, filtering should be observable.

## Proposed

Emit `tracing::debug!` (tool name, `Platform::current()`, `DesktopOs::current()`) when a tool is excluded, so an operator can see *why* an expected capability is absent.

## Acceptance criteria

- [ ] Filtering a tool out logs at debug with tool name + host platform/OS.
- [ ] No behavior change to the advertised set itself.

**Recommended agent:** rust-reviewer

PR #297, opened 2026-07-21
fix: OS-aware capability guardrails - fail loudly, no silent empties (#197 #198 #202 #208)

### LINKED ISSUE #197: Child A follow-up: guard against empty supported_os (supported-nowhere footgun)
**Parent epic:** #183. Follow-up from the `/review-pr` pass on #196 (Child A).

## Problem

`ToolSchema::os(vec![])` compiles and `supported_os: []` can deserialize from JSON. A tool that declares `Platform::Desktop` but an empty `supported_os` is then supported *nowhere* on desktop (`is_supported()` returns false for every `DesktopOs`), with no diagnostic. A tool author reaching for `.os(vec![])` when they meant `.platforms(vec![])` gets confusing silent behavior. The ~80 existing tools are safe (they take the `ALL_DESKTOP_OS` default); this is a hazard for new tools only.

## Options

- Add a `debug_assert!`/`tracing::warn!` in the `.os()` builder when the vec is empty.
- Or use a non-empty collection type to make the invariant type-enforced.

## Acceptance criteria

- [ ] An empty `supported_os` is either rejected, warned, or impossible to construct.
- [ ] Test covering the empty case.

jtomek-strike48:
## Summary

Four follow-ups from the #196 review (epic #183, Child A: OS-aware capability gating). All apply the epic's "fail loudly, no silent empties" principle. Closes #197, #198, #202, #208.

### #197 - empty `supported_os` footgun
`ToolSchema::os(vec![])` compiles and leaves a tool supported on *no* desktop OS (`is_supported()` false for every OS), so it silently vanishes from the advertised set - the likely cause being `.os(vec![])` where `.platforms(vec![])` was meant. The builder now `debug_assert!`s and `tracing::warn!`s on an empty vec (value left unchanged for back-compat). Test locks in that an empty `supported_os` is observably unsupported.

### #198 - observable filtering
`supported_schemas()`/`supported_names()` filtered unsupported tools out silently. New shared `is_supported_logged()` helper emits a `tracing::debug!` (tool name, `Platform::current()`, `DesktopOs::current()`) when a tool is excluded. No change to the advertised set itself.

### #208 - clear OS-incompatibility error (corrects the issue's suggested fix)
Tools are registered **unconditionally**, so on an unsupported host `get()` returns `Some` and `execute()` ran the tool anyway, failing cryptically deep in its body. The issue proposed a fix in the `None` arm, but that's **dead code** - `contains_key` is false exactly when `get()` returned `None`. The real fix is a `Some(tool) if !is_supported()` guard that returns `Error::PlatformNotSupported` naming the required + current OS *before* dispatch. A genuinely-absent tool still returns `ToolNotFound` (guarded by a separate test).

### #202 - typed errors for compiled-out fallbacks
Two feature-gated fallbacks returned `Ok(Vec::new())` - a silent lie:
- `get_network_interfaces` (non-Linux, `network-interface` off) fed an empty list into `check_wifi_connection_status` -> `total_adapters=0` -> `safe_to_scan=true` (a wrong "not on WiFi" negative the agent trusts).
- `mdns_discover` (`mdns-sd` off) read as "no services found."

Both now return `Error::PlatformNotSupported`, matching the sibling `ssdp_discover`. Shipping behavior is unchanged: the default `desktop` feature ships the real cross-platform impls. Note that both fallback arms are currently **unreachable in any buildable configuration** - `pub mod desktop` is `#[cfg(feature = "desktop")]`, and the `desktop` feature force-enables `network-interface` and `mdns-sd` (via `desktop-mdns`), so the arm needs the feature OFF while its module needs it ON. The change is directionally correct (matches `ssdp_discover`) and kept for a future feature-split, but it is dead code today, not exercised by any build.

## Test plan

- [x] 3 new `pentest-core` tests: execute rejects OS-unsupported tool with a clear error; unknown tool still `ToolNotFound`; empty `supported_os` unsupported
- [x] Mutation-verified per Lens 1: disabling the #208 gate makes the unsupported tool's `execute()` body panic fire (test goes RED); restoring is GREEN
- [x] `cargo test -p pentest-core -p pentest-platform --lib` - 271 + 45 passed
- [x] `cargo clippy -p pentest-core -p pentest-platform --all-targets -- -D warnings` - clean
- [x] `cargo fmt` - clean
- [x] Autopwn dispatch (`autopwn/toolchain/engine.rs`) now routes through `registry.execute()` so an OS-gated chain step (e.g. Linux-only `AutoPwnCrackTool` on macOS/Windows) hits the #208 guard instead of failing deep in the tool body; consumes the already-tested `execute()` path.
- [ ] The #202 fallback arms could not be compiled: as noted above, neither `#[cfg(not(feature = "mdns-sd"))]` (network.rs) nor `#[cfg(not(target_os = "linux"))]` (system.rs) is reachable in any feature configuration the workspace can build (the `desktop` module force-enables the features they gate off). Verified by reading + poison-token probe, not by compiling.

## Reviewer notes (honest scope)

- Both #202 arms are unreachable in any buildable feature config (see the #202 section): the `desktop` module gate force-enables the very features (`mdns-sd`, `network-interface`) the fallback arms require to be OFF. They are directionally-correct dead code matching `ssdp_discover`, verified by reading and a poison-token compile probe (an injected token in each arm never triggered a compile error under `desktop-pcap` or `--no-default-features`, while poisoning the compiled `mdns-sd`-ON arm did). Not verified by executing.
- #197's test constructs the schema via struct literal to bypass the new `debug_assert!` in `.os()` (which would otherwise panic the test in debug builds) while still asserting the `is_supported()` consequence.

### DIFF
--- a/crates/core/src/tools.rs
+++ b/crates/core/src/tools.rs
@@ -380,7 +380,28 @@ impl ToolSchema {
 
     /// Restrict the desktop OSes this tool runs on (e.g. `[DesktopOs::Linux]`
     /// for tools that shell out to Linux-only binaries). See #183.
+    ///
+    /// An empty vec is almost certainly a mistake: `is_supported()` would then
+    /// return false for *every* desktop OS, so the tool is advertised nowhere
+    /// and silently vanishes from the host tool list. (A likely cause is
+    /// reaching for `.os(vec![])` when `.platforms(vec![])` was meant.) We warn
+    /// loudly rather than change the value — the "supported nowhere" footgun in
+    /// #197. Tools that should run everywhere simply omit `.os(...)` and take
+    /// the [`ALL_DESKTOP_OS`] default.
     pub fn os(mut self, os: Vec<DesktopOs>) -> Self {
+        if os.is_empty() {
+            debug_assert!(
+                false,
+                "ToolSchema::os(vec![]) leaves '{}' supported on no desktop OS; \
+                 omit .os(...) for all-OS, or did you mean .platforms(vec![])?",
+                self.name
+            );
+            tracing::warn!(
+                tool = %self.name,
+                "ToolSchema::os(vec![]) sets an empty supported_os: this tool will be \
+                 advertised on NO desktop OS. Omit .os(...) for the all-OS default."
+            );
+        }
         self.supported_os = os;
         self
     }
@@ -835,9 +856,27 @@ impl ToolRegistry {
     pub fn supported_schemas(&self) -> Vec<ToolSchema> {
         self.tools
             .values()
-            .filter(|t| t.is_supported())
+            .filter(|t| self.is_supported_logged(t.as_ref()))
             .map(|t| t.schema())
             .collect()
+    }
+
+    /// `is_supported()` with an observability side-effect: when a tool is
+    /// filtered *out* of the advertised set, emit a `debug` line naming the
+    /// tool and the host platform/OS. Per the epic's "fail loudly, no silent
+    /// empties" principle (#198), an operator debugging a missing capability can
+    /// then see *why* it was excluded instead of it silently disappearing.
+    fn is_supported_logged(&self, tool: &dyn PentestTool) -> bool {
+        let supported = tool.is_supported();
+        if !supported {
+            tracing::debug!(
+                tool = %tool.name(),
+                platform = ?Platform::current(),
+                desktop_os = ?DesktopOs::current(),
+                "tool filtered from advertised set (unsupported on this host)"
+            );
+        }
+        supported
     }
 
     /// Get tool names
@@ -850,7 +889,7 @@ impl ToolRegistry {
     pub fn supported_names(&self) -> Vec<&str> {
         self.tools
             .values()
-            .filter(|t| t.is_supported())
+            .filter(|t| self.is_supported_logged(t.as_ref()))
             .map(|t| t.name())
             .collect()
     }
@@ -863,6 +902,26 @@ impl ToolRegistry {
         ctx: &ToolContext,
     ) -> Result<ToolResult> {
         match self.get(name) {
+            Some(tool) if !tool.is_supported() => {
+                // The tool is registered but not supported on this host's OS
+                // (tools are registered unconditionally; the OS gate is at
+                // advertise/execute time — see #183). Without this arm, an
+                // OS-incompatible tool would run and fail cryptically deep in
+                // its own logic. Fail early with a message that says *why*
+                // (required OS vs current OS), not a generic "not found" (#208).
+                let required = tool.supported_os();
+                let current = DesktopOs::current();
+                tracing::error!(
+                    tool = %name,
+                    ?required,
+                    ?current,
+                    "tool exists but is not supported on this OS"
+                );
+                Err(crate::error::Error::PlatformNotSupported(format!(
+                    "Tool '{name}' exists but is not supported on this OS \
+                     (required: {required:?}, current: {current:?})"
+                )))
+            }
             Some(tool) => tool.execute(params, ctx).await,
             None => {
                 // Find similar tool names for suggestions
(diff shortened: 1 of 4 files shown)


cchao-devo (approved):
Review: Pick PR #297 — OS-aware capability guardrails — fail loudly, no silent empties (#197 #198 #202 #208)

  Verdict: ✅ Approve. Correctly closes all four silent-failure follow-ups with no new silent failures introduced. Gates green, panel + independent verification agree. No blockers.

  Lenses run: the real pr-review-toolkit plugin agents — silent-failure-hunter (lead, this is a silent-empties PR) + code-reviewer — plus my own source verification of the two load-bearing
  claims and mechanical gates. Verified in pick_08/pick; pick_07 untouched, no cluster.

  Gates (on pr-297)
  
  - cargo fmt --all --check clean · cargo clippy --workspace --all-targets -- -D warnings clean · pentest-core 271/0, pentest-platform 43/0 (incl. all 10 os_capability tests, run
  empirically by the agents)

  The four fixes land correctly
  
  - #208 (execute guard): Some(tool) if !tool.is_supported() => Err(PlatformNotSupported) is ordered before the dispatch arm (verified — Rust top-to-bottom arm eval, can't be bypassed),
  and returns a typed error naming required-vs-current OS instead of running the tool body cryptically. The author correctly identified the issue's suggested fix (None arm) as dead code
  and fixed the real path.
  - #198 (observable filtering): is_supported_logged adds a debug line when a tool is filtered from the advertised set — pure observability, no behavior change.
  - #197 (empty supported_os): debug_assert! + tracing::warn!, value left unchanged for back-compat; observably unsupported (fail-closed). Confirmed debug_assert compiles out in release
  (just warns), so no release panic; the test uses struct-literal construction to bypass the assert while still asserting !is_supported().
  - #202 (typed-error fallbacks): both Ok(Vec::new()) → PlatformNotSupported, matching the ssdp_discover sibling.

  Two load-bearing properties — independently verified
  
  1. Advertise and execute use the identical predicate. supported_schemas/supported_names → is_supported_logged → tool.is_supported(), and the #208 execute-guard → tool.is_supported() —
  both resolve to the same trait method (tools.rs:801). So no tool can be advertised-but-rejected or hidden-but-executable. ✓
  2. The autopwn consumer change doesn't create a new silent failure. engine.rs now routes through registry.execute(); I read the Err arm (engine.rs:271-320): a PlatformNotSupported is
  recorded (execution.fail + FailedStep), logged (tracing + eprintln), driven into step.alternatives, and propagated as Err only if step.required (else recorded + continue). Surfaced, not
  swallowed — matches the comment. The Some(_) match keeps execute()'s None arm unreachable, so a present-but-unsupported tool is never misreported as ToolNotFound. ✓

  On the #202 dead-code disclosure (honest and accurate)

  The author is upfront that both #202 arms are unreachable in any buildable config: pub mod desktop is #[cfg(feature="desktop")] and desktop force-enables network-interface + mdns-sd
  (desktop-mdns), so the arms need the feature OFF while their module needs it ON. Verified: android/ios have their own independent impls (never reach the desktop fallbacks), and the
  desktop callers propagate via ?. So "shipping behavior unchanged" holds — the change is directionally-correct dead code for a future feature-split, matching ssdp_discover. Correctly
  labeled, correctly not claimed as tested-by-execution.

  Minor (non-blocking)
  
  - is_supported_logged(&self, ...) doesn't read self — a clippy::unused_self candidate, but that's pedantic-tier (not in -D warnings), and keeping it a method is fine for call-site
  ergonomics. The code-reviewer agent flagged it below threshold; agreed, not worth changing.

  init-dev relevance

  Low. This is OS-capability gating for the desktop build. init-dev's headless connector runs on Linux, so Linux-only tools are supported there and the #208 guard wouldn't reject them; the
  #202 arms are dead code regardless. No consumer impact.

PR #298, opened 2026-07-21
fix(nmap): accept -p- / ports="-" as full-port-range synonyms (#257)

### LINKED ISSUE #257: nmap: ports="all" / flags:"-p-" fails validation before reaching the -p- mapping
## Summary

A full-port-scan request expressed as the nmap CLI flag `-p-` (or, equivalently, `ports: "-"`) errors out during validation instead of running a full scan. Priority: LOW.

## Root cause

In `crates/tools/src/external/nmap.rs`, the `execute()` port dispatch maps the *string literal* `"all"` to the `-p-` flag:

```rust
match ports.as_str() {
    "top100" => builder = builder.arg("--top-ports", "100"),
    "top1000" => {}                       // default
    "all"     => builder = builder.flag("-p-"),
    _ => {
        let validated_ports = validate_port_spec(&ports)?;   // <-- "-" lands here
        builder = builder.arg("-p", &validated_ports);
    }
}
```

The legacy-arg normalizer turns `-p-` into `ports = "-"` (attached-port form: `-p` + spec `-`). `"-"` is not `"all"`, so it falls to `validate_port_spec("-")`, which splits on `-` into `["", ""]` and returns `Invalid port number`. The `"all" => -p-` branch is never reached.

## Why it matters now

- Identical on `main` (pre-existing, **not** a regression introduced by #230).
- PR #230 widens the surface where `-p-` appears: it accepts a `flags` value and the tool **description advertises `-p-`**, so the model is now more likely to emit `-p-` / `-p -` and hit this error.

## Repro

```jsonc
// any of these:
{ "target": "10.0.0.5", "ports": "-" }
{ "target": "10.0.0.5", "flags": "-p-" }
{ "target": "10.0.0.5", "flags": ["-p-"] }
```
-> `InvalidParams: Invalid port number:` instead of a full 1-65535 scan.

## Suggested fix (pick one)

- Normalize `-p-` / `"-"` -> `ports = "all"` in `apply_flag_tokens` (map the CLI spelling to the structured value), **or**
- Accept `"-"`/`"all"`/`"1-65535"` as equivalents in the `execute()` port match before calling `validate_port_spec`, **or**
- Teach `validate_port_spec` to accept the bare `-` full-range spelling and return `1-65535`.

## Tests to add

- `flags: "-p-"` (+ valid target) -> builder emits `-p-` (full scan), no validation error.
- `ports: "-"` structured -> same.
- Regression: `ports: "all"` still maps to `-p-`.

Found during review of #230 (cchao-devo). Explicitly deferred there as a separate follow-up.

jtomek-strike48:
## Summary

Closes #257. A full-port-scan request expressed as the nmap flag `-p-` (or `ports: "-"`) errored during validation instead of running a full 1-65535 scan.

## Root cause

The legacy-arg normalizer folds `-p-` into `ports = "-"` (attached-port form: `-p` + spec `-`). In `execute()`'s port dispatch, `"-"` is not `"all"`, so it fell through to `validate_port_spec("-")`, which splits on `-` into `["", ""]` and returns `Invalid port number`. The `"all" => -p-` branch was never reached.

This is **pre-existing on main** (not introduced by #230), but #230 widened the surface: the tool now folds a `flags` value and its description advertises `-p-`, so the model is more likely to emit `-p-` / `flags:["-p-"]` and hit the error.

## Fix

Extract the port-dispatch into a pure, unit-testable `resolve_port_args()` helper that treats all full-range spellings as equivalent — structured `"all"`, the bare `"-"` from the normalizer, and explicit `"1-65535"` all map to the bare `-p-` flag. Custom specs still go through `validate_port_spec`; genuinely-invalid input is still rejected.

```jsonc
// all of these now run a full scan instead of erroring:
{ "target": "10.0.0.5", "ports": "-" }
{ "target": "10.0.0.5", "flags": "-p-" }
{ "target": "10.0.0.5", "flags": ["-p-"] }
```

## Test plan

- [x] 6 new `pentest-tools` unit tests: all three full-range synonyms -> `-p-`; direct `ports="-"` regression; `top100`/`top1000`/custom pass-through; still-rejects-invalid (`not-a-port`, `99999`); end-to-end `flags:"-p-"` and `flags:["-p-"]` fold -> full-scan chain
- [x] Mutation-verified per Lens 1: reverting `resolve_port_args` to the `"all"`-only arm turns the 3 synonym tests RED; restoring is GREEN
- [x] `cargo test -p pentest-tools --lib external::nmap` - 60 passed
- [x] `cargo fmt` - clean

## Note

Chose the "normalize synonyms at the dispatch point" option (of the three the issue suggested) because it keeps `validate_port_spec` untouched (used by other callers) and makes the whole port decision pure and testable without executing nmap.

### DIFF
--- a/crates/tools/src/external/nmap.rs
+++ b/crates/tools/src/external/nmap.rs
@@ -868,14 +868,21 @@ fn estimate_host_count(target: &str) -> usize {
 /// `["",""]` and errored with "Invalid port number", so a model emitting `-p-`
 /// (which the tool description advertises) got a validation failure instead of
 /// a full scan.
+///
+/// The keyword synonyms (`top100` / `top1000` / `all`) are matched
+/// case-insensitively: the schema advertises them lowercase, but the LLM emits
+/// case variants (`"ALL"`, `"Top100"`), which would otherwise fall through to
+/// `validate_port_spec` and error. A custom numeric spec is left verbatim for
+/// `validate_port_spec` (digits/commas/dashes have no case to fold).
 fn resolve_port_args(ports: &str) -> Result<Vec<(&'static str, Option<String>)>> {
-    match ports.trim() {
+    let trimmed = ports.trim();
+    match trimmed.to_ascii_lowercase().as_str() {
         "top100" => Ok(vec![("--top-ports", Some("100".to_string()))]),
         "top1000" => Ok(vec![]), // nmap default: no port flag
         // Full-range synonyms -> -p- (scan all 65535 ports).
         "all" | "-" | "1-65535" => Ok(vec![("-p-", None)]),
-        spec => {
-            let validated = validate_port_spec(spec)?;
+        _ => {
+            let validated = validate_port_spec(trimmed)?;
             Ok(vec![("-p", Some(validated))])
         }
     }
@@ -1122,6 +1129,22 @@ mod tests {
         // The fix must not blanket-accept: a real bad spec still errors.
         assert!(resolve_port_args("not-a-port").is_err());
         assert!(resolve_port_args("99999").is_err());
+    }
+
+    #[test]
+    fn resolve_ports_keyword_synonyms_are_case_insensitive() {
+        // The schema advertises lowercase keywords, but the LLM emits case
+        // variants; they must resolve, not fall through to validate_port_spec
+        // and error (adversarial-review finding on #257).
+        assert_eq!(resolve_port_args("ALL").unwrap(), vec![("-p-", None)]);
+        assert_eq!(resolve_port_args("All").unwrap(), vec![("-p-", None)]);
+        assert_eq!(
+            resolve_port_args("TOP100").unwrap(),
+            vec![("--top-ports", Some("100".to_string()))]
+        );
+        assert_eq!(resolve_port_args("Top1000").unwrap(), vec![]);
+        // Whitespace around a keyword is also tolerated (trim happens first).
+        assert_eq!(resolve_port_args("  all  ").unwrap(), vec![("-p-", None)]);
     }
 
     #[test]
--- a/crates/tools/src/external/nmap.rs
+++ b/crates/tools/src/external/nmap.rs
@@ -772,11 +772,14 @@ fn calculate_timeout(
     // Estimate number of target hosts
     let host_count = estimate_host_count(target);
 
-    // Estimate number of ports
-    let port_count = match ports {
+    // Estimate number of ports. Fold the keyword/full-range synonyms the same
+    // way resolve_port_args does (#257) so a full scan requested as "-" /
+    // "1-65535" / "ALL" is sized as 65535 ports, not mis-estimated as 1 and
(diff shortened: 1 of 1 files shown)


cchao-devo (approved):
Review: Pick PR #298 — fix(nmap): accept -p- / ports="-" as full-port-range synonyms (#257)

  Verdict: ✅ Approve. Correct, well-tested fix that closes the #257 finding I filed during the #230 review. The synonym mapping doesn't over-widen (verified by running it against
  adversarial inputs), validate_port_spec is untouched for other callers, and the author caught + fixed a coupled second bug. No blockers.

  Lenses run: mechanical gates (fmt + workspace clippy + tests) + manual trace + empirical execution of resolve_port_args against production-shaped and adversarial inputs (the CLAUDE.md
  "run, don't read" mandate). Single-file pure-logic fix, so I verified directly rather than spinning the full panel. Verified in pick_08/pick; pick_07 untouched, no cluster.

  Gates (on pr-298)

  - cargo fmt --all --check clean · cargo clippy --workspace --all-targets -- -D warnings clean · nmap tests 62 passed

  The fix is correct and doesn't over-widen (verified by running it)

  I ran resolve_port_args on ~22 inputs beyond the test set. Actual outputs:
  - Full-range synonyms → -p-: "-", "all", "ALL", "1-65535", "  all  " ✓
  - Still rejects invalid (no blanket-accept — the key non-widening property): 1-65536, 0, 0-65535, 80-, -80, "", ",", "1,,2", "http", 999999... (overflow) all → Err. The "-"→full-range
  mapping loosened only the three intended synonyms. ✓ 
  - Custom specs pass through verbatim to the untouched validate_port_spec (80,443, 65535, 1-65535,7000). ✓
  - -p- as a literal ports value correctly errors — it only becomes a full scan via flag-folding (flags:"-p-" → ports="-"), verified end-to-end. ✓

  The coupled-bug catch (the standout)

  calculate_timeout had the same "all"-only literal match — so a full scan requested as "-"/"1-65535"/"ALL" would have been sized as 1 port, clamped to the 60s floor, and killed mid-scan.
  The author fixed both the dispatch and the timeout estimate with the identical synonym fold, and added full_range_synonyms_get_the_same_timeout_as_all to prove the composition. This is
  exactly the "two coupled values feeding one outcome — fix both, test the composed result at real inputs" discipline; missing the timeout half would have shipped a fix that still killed
  full scans. Well done.

  Other
  
  - validate_port_spec body unchanged (verified) — the 2 other callers (port_scan.rs, masscan.rs) are unaffected.
  - Case-insensitive keyword matching (ALL, Top100) + trim is a sensible robustness add for LLM-emitted case/whitespace variants; matched by tests.
  - Comments are accurate and explain the root cause + the case-fold rationale.
  - The extraction to a pure (flag, Option<value>) list is clean and made the whole thing unit-testable without executing nmap — good design.

  init-dev relevance

  None directly — nmap tool arg handling. (init-dev's headless connector runs sandbox-disabled; this is scan-scope arg logic, not a sandbox path.)

Issue #181, opened 2026-06-30
Determine available network size (scope-bounded, no unbounded scanning)


jtomek-strike48:
**Issue Type:** Feature
**Component:** Pick (`crates/tools/src/safety_check` network discovery; `crates/core/src` scope/recon)
**Effort:** Medium (2-3 days)
**Priority:** P2
**Source:** Operator requirement - "Pick should be able to figure out the network size that is available."
**Target repo:** `Strike48-public/pick`
**Pairs with:** pick#182 (`/24` netmask bug - prerequisite for accurate sizing)

> **Reality check (verified against code 2026-06-30):** Pick already has the building
> blocks - this is composition + scope-binding, not new scanning.
> - **Local discovery exists:** `crates/tools/src/safety_check/network_map.rs:37`
>   `discover_network()` detects the default interface/gateway/local IP
>   (`:176` `get_gateway_and_local_ip`), derives a subnet (`:200` `determine_subnet` -
>   **hardcoded /24**, the pick#182 bug), runs a bounded nmap `-sn` sweep
>   (`:213`), merges the ARP table (`:143`), returns a deduped host set (`:71`).
> - **It already reasons about size:** `safety_check/mod.rs:180` `cap_for_network_size`,
>   `BUSY_NETWORK_DEVICE_THRESHOLD = 20` (`:171`), `spans_multiple_subnets` (`:207`).
> - **Scope is enforced via injected exclude:** Matrix injects the engagement's
>   out-of-scope hosts into scan tools' `exclude` (`external/runner.rs:42`
>   `param_exclude_list`, re-validated `:67`; wired into `external/nmap.rs:104,249-252`
>   and `external/masscan.rs:75-77,121-134`) - issue #2524. CIDR is first-class
>   (`crates/core/src/validation.rs:250` `validate_cidr`, `:308` `validate_target`).
> - **Engagement target** lives at `crates/core/src/orchestrator.rs:34` `EngagementInfo.target`.
> - **Specialist pattern** for adding capability: `crates/core/src/specialist_spawner.rs`
>   (`SpecialistType` `:30`, indicator pattern `:454`, prompts in
>   `skills/claude-red/specialists/*`).
> - **Caveat:** `determine_subnet` assumes /24 (`network_map.rs:200`), IPv6 unsupported
>   (`:206`); `scope_violation` is a prompt convention, not enforced in `core`
>   (`evidence.rs` has no scope check) - hard scope enforcement is the injected
>   `exclude` list only.

---

[...]

PR #299, opened 2026-07-21
fix(safety_check): derive sweep subnet from real netmask, not hardcoded /24 (#182)

### LINKED ISSUE #182: Bug: determine_subnet hardcodes /24, ignoring the interface netmask
**Issue Type:** Bug
**Component:** Pick (`crates/tools/src/safety_check/network_map.rs`)
**Effort:** Small (2-4 hours)
**Priority:** P2
**Source:** Surfaced during grounding for pick#181 (network-size detection).
**Target repo:** `Strike48-public/pick`
**Pairs with:** pick#181 (accurate sizing depends on this fix)

> **Reality check (verified against code 2026-06-30):** `determine_subnet/1`
> (`crates/tools/src/safety_check/network_map.rs:196-210`) builds the subnet by
> **assuming /24** - it takes the first three octets and appends `.0/24`, and its own
> doc comment says so: *"Assumes /24 subnet for simplicity. In production, we'd use the
> netmask from the interface."* The real netmask **is already available**: the same
> module reads the interface's IPv4 via `default_net` (`network_map.rs:187-191`,
> `interface.ipv4.first()`), and `default_net`'s `Ipv4Net` carries the prefix/netmask -
> so the fix reads data already in hand. IPv6 currently bails (`:206-208`).

---

## Problem Statement

`determine_subnet` always returns a `/24`, regardless of the interface's actual netmask.
On any network that isn't /24 - a /16 corporate LAN, a /25 or /23 segment, a /22 - the
derived subnet is **wrong**:

- On a wider net (e.g. /16), discovery sweeps only 1/256th of the real subnet (the
  enclosing /24), **undercounting** hosts and network size.
- On a narrower net (e.g. /25), the assumed /24 spans **two** real subnets, potentially
  directing a sweep at addresses outside the agent's actual segment.

This directly corrupts the network-size signal (pick#181) and any host-count
reasoning built on the discovered set.

---

## Proposed Solution

Read the real prefix length from the interface instead of assuming /24.

**File:** `crates/tools/src/safety_check/network_map.rs`

`get_gateway_and_local_ip` (`:176`) already obtains the interface; extend it (or
`determine_subnet`) to carry the `default_net` `Ipv4Net` prefix length, and compute the
network address from `addr & netmask`:

```rust
// illustrative - use the prefix already provided by default_net's interface.ipv4
fn determine_subnet(net: &default_net::ip::Ipv4Net) -> anyhow::Result<String> {
    let prefix = net.prefix_len();              // real netmask, not assumed
    let network = net.addr & mask_from_prefix(prefix);
    Ok(format!("{}/{}", network, prefix))
}
```

- Compute the network base address by applying the netmask (don't just zero the last
  octet).
- Keep a **sane safety cap**: if the real prefix is very wide (e.g. < /16), do **not**
  silently expand a sweep to a huge range - clamp/skip and report (ties to pick#181's
  bounded-discovery + `cap_for_network_size`, `mod.rs:180`). A netmask fix must not become
  an accidental "scan a /8" enabler.

### IPv6
Either compute the IPv6 prefix properly or keep the explicit "not yet supported" bail
(`:206-208`) - but make it explicit, not a silently-wrong /24-style assumption.

---

## Testing Strategy

```rust
#[test]
fn derives_subnet_from_real_prefix_not_assumed_24() {
    assert_eq!(subnet_for(ipv4net("10.1.2.3", 16)), "10.1.0.0/16");
    assert_eq!(subnet_for(ipv4net("192.168.1.130", 25)), "192.168.1.128/25");
    assert_eq!(subnet_for(ipv4net("172.16.5.9", 23)), "172.16.4.0/23");
}

#[test]
fn computes_network_address_via_netmask() {
    // not just zeroing the last octet: /25 of .130 -> .128, not .0
}

#[test]
fn very_wide_prefix_is_capped_not_silently_swept() {
    // prefix < /16 -> clamped/reported, no huge implicit sweep
}
```

---

## Acceptance Criteria

- [ ] `determine_subnet` uses the interface's real prefix length (no hardcoded /24)
- [ ] Network base address computed via netmask (correct for /25, /23, /16, etc.)
- [ ] Very-wide prefixes are capped/reported, not turned into an implicit large sweep (ties to pick#181 bounds)
- [ ] IPv6 handled explicitly (computed or cleanly reported unsupported)
- [ ] Tests cover /16, /23, /24, /25 derivation + network-address math + the wide-prefix cap

---

## References

- `crates/tools/src/safety_check/network_map.rs:196-210` - `determine_subnet` (the /24 assumption)
- `network_map.rs:187-191` - interface IPv4 already read via `default_net` (prefix available here)
- `crates/tools/src/safety_check/mod.rs:180` - `cap_for_network_size` (use for the wide-prefix cap)
- pick#181 - network-size detection (consumer of the corrected subnet)

---

**Proposed labels:** `product/pick`, `type/bug`, `priority/P2`, `area/recon-agent`, `size/S`

**Risk:** Low - reads data already fetched; main care is the wide-prefix safety cap so the
fix doesn't widen scan scope.

jtomek-strike48:
## Summary

Closes #182. `determine_subnet` always returned `<a>.<b>.<c>.0/24` regardless of the interface's actual prefix, corrupting network discovery and the host-count/size signal (#181 consumer).

- On a wider net (e.g. /16), discovery swept only 1/256th of the real subnet.
- On a narrower net (e.g. /25), the assumed /24 spanned two real segments, potentially probing addresses outside the agent's own subnet.

## Root cause

The real prefix was already fetched but thrown away: `interface.ipv4.first()` yields a `default_net` `Ipv4Net` that carries `prefix_len` / `netmask`, but `get_gateway_and_local_ip` mapped it straight to a bare `IpAddr` (`net.addr`), dropping the prefix before `determine_subnet` ever saw it.

## Fix

- Thread the local IPv4 `prefix_len` out of `get_gateway_and_local_ip`.
- New pure `subnet_cidr_v4(addr, prefix)` computes the network base **via the netmask** (`addr & mask`), so `/25` of `.130` -> `.128/25`, not `.0/24`.
- **Wide-prefix safety cap:** a prefix wider than `/16` (`MIN_SUBNET_PREFIX`), or a degenerate `/0`, is clamped to `/16` and logged - a netmask fix must never silently widen a sweep to a `/8`-scale range. Host-count bounding for sizing still lives in `cap_for_network_size`.
- Missing prefix falls back to `/24` (prior behavior preserved); IPv6 keeps its explicit "not yet supported" bail rather than a silently-wrong assumption.

## Test plan (all AC covered)

- [x] Derivation for /16, /23, /24, /25, /26, /30, /32
- [x] Network address computed via netmask, not octet-zeroing (`.130 /25 -> .128`)
- [x] Very-wide prefix capped: /8 and /12 -> /16; /0 -> /16; /16 boundary unchanged
- [x] IPv6 explicitly unsupported; no-prefix falls back to /24
- [x] Mutation-verified per Lens 1: removing the cap fails the cap test; reverting to octet-zeroing fails 5 netmask-math tests
- [x] `cargo test -p pentest-tools --lib safety_check` - 40 passed
- [x] `cargo fmt` - clean

## Note

`subnet_cidr_v4` is a pure function (no interface/I/O), so the subnet math is fully unit-tested without touching real network state. Unblocks #181 (network-size detection), which depends on an accurate subnet.

### DIFF
--- a/crates/tools/src/safety_check/network_map.rs
+++ b/crates/tools/src/safety_check/network_map.rs
@@ -5,7 +5,7 @@ use super::types::{Device, NetworkMap, ThreatLevel};
 use anyhow::Context;
 use pentest_platform::NetworkOps;
 use std::collections::HashMap;
-use std::net::IpAddr;
+use std::net::{IpAddr, Ipv4Addr};
 use std::process::Stdio;
 use tokio::process::Command;
 
@@ -16,6 +16,16 @@ use tokio::process::Command;
 /// box). Kept conservative to avoid alarming a non-technical user over a
 /// normal NAS or printer.
 const SUSPICIOUS_PORT_THRESHOLD: usize = 10;
+
+/// Narrowest (largest) prefix we will ever widen a discovery sweep to.
+///
+/// The interface may legitimately report a very wide prefix (e.g. /8 on some
+/// VPN or container setups). Sweeping a /8 is 16M addresses - a netmask fix
+/// (#182) must never silently turn into "scan the whole /8". When the real
+/// prefix is wider than this, we clamp the sweep subnet to `/16` (still large,
+/// but bounded) and log the clamp so the narrowing is observable. The full
+/// host-count bounding for sizing lives in `cap_for_network_size` (mod.rs).
+const MIN_SUBNET_PREFIX: u8 = 16;
 
 /// Enrichment data for a single host, keyed by IP, drawn from the ARP table.
 struct ArpInfo {
@@ -39,7 +49,7 @@ pub async fn discover_network() -> anyhow::Result<NetworkMap> {
 
     // Get gateway and local IP
     let gateway_info = get_gateway_and_local_ip().await?;
-    let (gateway_ip, local_ip) = gateway_info;
+    let (gateway_ip, local_ip, local_prefix) = gateway_info;
 
     tracing::info!("Discovered: gateway={}, local={}", gateway_ip, local_ip);
 
@@ -52,7 +62,7 @@ pub async fn discover_network() -> anyhow::Result<NetworkMap> {
     // Discovery source 1: an nmap ping sweep of the local /24. Fast and finds
     // hosts we have not yet talked to. Best-effort - on a larger network this
     // only covers our own /24, which is why we also merge the ARP table.
-    let subnet = determine_subnet(&local_ip)?;
+    let subnet = determine_subnet(&local_ip, local_prefix)?;
     let nmap_ips = match run_nmap_arp_scan(&subnet).await {
         Ok(ips) => {
             tracing::info!("nmap discovered {} hosts in {}", ips.len(), subnet);
@@ -172,8 +182,14 @@ async fn collect_arp_info() -> HashMap<IpAddr, ArpInfo> {
     map
 }
 
-/// Get gateway and local IP address.
-async fn get_gateway_and_local_ip() -> anyhow::Result<(IpAddr, IpAddr)> {
+/// Get gateway and local IP address, plus the local IPv4 prefix length.
+///
+/// The prefix comes from `default_net`'s `Ipv4Net` (which carries `prefix_len`
+/// / `netmask` alongside the address); it is threaded out here so the sweep
+/// subnet reflects the real netmask instead of an assumed /24 (#182). `None`
+/// when there is no IPv4 (should not happen given the `.context` below, but the
+/// type keeps the IPv6-only future honest).
+async fn get_gateway_and_local_ip() -> anyhow::Result<(IpAddr, IpAddr, Option<u8>)> {
     let interface = tokio::task::spawn_blocking(default_net::get_default_interface)
         .await
         .context("Failed to spawn interface detection task")?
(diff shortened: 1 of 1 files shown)


cchao-devo (approved):
Review: Pick PR #299 — fix(safety_check): derive sweep subnet from real netmask, not hardcoded /24 (#182)
  
  Verdict: ✅ Approve. Correct netmask math (verified by running it across the full prefix range), a sensible wide-prefix safety cap that prevents scope-widening, clean contract threading.
  No blockers.

  Lenses run: mechanical gates (fmt + workspace clippy + tests) + manual review + empirical execution of subnet_cidr_v4 across all prefixes and edge cases (the CLAUDE.md "run, don't read"
  mandate — highest-leverage here since it's pure bit-math). Single-file pure-logic fix, so verified directly rather than spinning the full panel. Verified in pick_08/pick; pick_07
  untouched, no cluster.

  Gates (on pr-299)
  
  - cargo fmt --all --check clean · cargo clippy --workspace --all-targets -- -D warnings clean · safety_check 40 passed

  The netmask math is correct (verified by running subnet_cidr_v4 on 19 inputs)

  - Network base via netmask, not octet-zeroing: /25 of .130 → .128 ✓, /26 of .200 → .192 ✓, /23 of 172.16.5.9 → 172.16.4.0 ✓, /30 of .6 → .4 ✓. This is the actual #182 bug fix and it's
  right.
  - /31, /32 correct (single/pair addresses).
  - No UB: effective is clamped to [16,32], so the shift 32-effective ∈ [0,16] is always well-defined — the /0 shift-by-32 footgun is guarded by the clamp. u32::from/Ipv4Addr::from are
  both big-endian, so octet order is consistent.

  The safety cap prevents scope-widening (the security-relevant property)
  
  This is a pentest sweep range, so the dangerous direction is a fix that widens the scan. Verified the MIN_SUBNET_PREFIX = /16 cap holds: /0, /1, /8, /12, /15 all clamp to /16 (never a
  /8-scale 16M-address sweep), /16 boundary passes through unchanged, and the clamp is logged so the narrowing is observable. A defensive >32 clamp handles nonsensical input (/33, /255 →
  /32). Correct fail-safe design — the netmask fix can't turn into "scan the whole /8."

  Contract threading is clean
  
  - get_gateway_and_local_ip now returns (IpAddr, IpAddr, Option<u8>) — the prefix was already fetched from default_net's Ipv4Net and previously discarded; now threaded through.
  Self-contained: the only callers are the one production site (updated) and tests. No stale 1-arg/2-tuple callers (grep-confirmed).
  - Prior behavior preserved: missing prefix → /24 fallback; a real /24 still derives /24; IPv6 keeps its explicit "not yet supported" bail rather than a silently-wrong assumption. Good —
  no regression for the common case.

  Consumer note (accurate)
  
  The fix unblocks #181 (network-size detection), which depends on an accurate subnet/host-count. The PR correctly notes host-count sizing bounding still lives in cap_for_network_size
  (mod.rs) — so this change fixes the scope signal without touching the sizing cap. Both layers compose (the /16 cap bounds the sweep; cap_for_network_size bounds the sizing) — no
  double-cap conflict.

  init-dev relevance
  
  None — this is the safety_check network-discovery path for local network mapping; not a sandbox or connector-contract surface. init-dev's headless connector on Linux would get accurate
  netmask-derived subnets (an improvement), but nothing changes in the deployment wiring.

Issue #260, opened 2026-07-17
cargo check -p pentest-ui (isolated) fails with 154 errors in pentest-tools/wifi_scan_detailed.rs (invalid feature combo)


jtomek-strike48:
## Problem

Running `cargo check -p pentest-ui` (or `cargo clippy -p pentest-ui`, `cargo test -p pentest-ui`) **in isolation** fails with ~154 compile errors, all in `crates/tools/src/wifi_scan_detailed.rs` -- a file the UI crate does not touch. Example:

```
error[E0282]: type annotations needed
   --> crates/tools/src/wifi_scan_detailed.rs:158:17
    |
158 |   let mut networks = platform.get_wifi_networks(selected_adapter.clone()).await?;
    |       ^^^^^^^^^^^^
```

This is confusing and wastes time: someone iterating on UI code naturally runs `cargo check -p pentest-ui`, gets a wall of errors pointing at wifi/monitor-mode code they never edited, and has to discover on their own that the check is meaningless in isolation.

## Root cause

`crates/ui/Cargo.toml` wires its platform dependencies asymmetrically:

```toml
pentest-platform = { workspace = true }                        # default features -> desktop ON
pentest-tools    = { workspace = true, default-features = false } # desktop OFF
```

When `pentest-ui` is checked alone, Cargo's resolver activates only the features `pentest-ui` requests. `pentest-tools` compiles with `default-features = false` (its `desktop` feature off), which is a combination that never actually ships -- every real app (`pentest-desktop`, `pentest-headless`, `pentest-web`) activates `pentest-tools/desktop` via `pentest-platform`'s `desktop` default, and feature unification turns it on at the workspace level.

In that unsupported off-configuration, `wifi_scan_detailed.rs` calls `PlatformProvider` methods (`get_wifi_networks`, `enable_monitor_mode`) whose return types are `#[cfg(feature = "desktop")]`-gated in `crates/platform/src/lib.rs`, so type inference has nothing to resolve against -> E0282 x N.

The `default-features = false` on `pentest-tools` is presumably intentional so `pentest-ui` can also target android/ios/web, with the consuming app selecting the platform. The bug is that the **desktop-off combination does not typecheck at all**, and nothing signals that isolated-crate checking is invalid.

## Impact

[...]

PR #300, opened 2026-07-21
feat(health): report real connector transport/heartbeat state on /hea…

### LINKED ISSUE #295: /health should report real connector state (gRPC session + registration), not a static OK
> Surfaced during matrix#3216 (Pick → newdev) — Strike48/init-dev#334 review (Jonathan). Related: Strike48/sdk-rs#59 (connector auto-reconnect) — both are about the connector honestly reporting / recovering its own liveness.

## Summary

Pick's `/health` endpoint returns a static `"OK"` (`crates/ui/src/liveview_server.rs:318`, and the file-browser variant `bin/file_browser_connector.rs:357`). It reflects only that the LiveView HTTP server is up — **not** whether the connector is actually functional (gRPC session connected + tools registered). So polling `/health` — or a k8s liveness/readiness probe pointed at it — cannot distinguish a healthy connector from a **wedged** one.

## Why it matters

The failure mode that actually takes Pick down keeps the process (and the HTTP server) alive:

- post-approval OTT→JWT exchange fails and the connector wedges (stops retrying),
- the gRPC stream drops / heartbeat stops and the SDK does not auto-reconnect (Strike48/sdk-rs#59),
- the session terminates server-side (heartbeat timeout) and the tool registry empties.

In all of these, PID 1 is fine and `/health` returns `"OK"`, so:

- The k8s deployment probe (`kube-charts pick-connector` currently uses `test -e /proc/1/status`, and even switching it to `httpGet /health` would be equally blind) never restarts the pod — k8s shows a misleading `Running`.
- Any external health poll gets a false green.

The init-dev side works around this out-of-band (`task newdev:status` greps Studio's connector registry, and `task newdev:fix-pick-approval` recovers a wedge), but that's operator-driven; there's no signal a probe or monitor can act on.

## Ask

Make `/health` reflect the connector's **actual** operational state, so polling it gives an accurate picture. Suggested: return non-200 (or a JSON body with a boolean) when the connector is NOT in a working state, keyed on:

- gRPC stream to the Studio connectors gateway is connected,
- the connector has a valid session (post-approval JWT obtained, not stuck in pending/retry),
- tools are registered.

Optionally expose a richer body for observability, e.g.:

```json
{ "status": "ok" | "degraded", "grpc_connected": true, "session": "active" | "pending" | "wedged", "tools_registered": 116, "last_heartbeat_secs_ago": 12 }
```

That lets a k8s `readinessProbe`/`livenessProbe` on `/health` restart or de-route a wedged pod, and lets external monitors detect the dropped-session state directly.

## Context

- Chart probe today: `k8s/charts-local/pick-connector/templates/deployment.yaml` (init-dev) uses `test -e /proc/1/status` for both probes — a near-no-op. Once `/health` is meaningful, the chart can point the probes at it (tracked on the init-dev side).
- This is the connector-side counterpart to Strike48/sdk-rs#59: #59 makes the connector *recover* (auto-reconnect); this makes it *report* its state so k8s/monitoring can act when it can't.

cchao-devo:
…lth (#295)

  ## What & why

  Closes #295. Pick's `/health` returned a static `"OK"` — it only proved the LiveView HTTP
  server was up, not whether the connector was actually functional. This enriches `/health` to
  report the connector's real **transport/heartbeat state** from the SDK's
  `ConnectorRunner::get_stats()`, so polling it (or a k8s probe / `newdev:status`) gives an
  accurate picture instead of a false green. **No connector behavior changes** — the endpoint
  only reads existing SDK state.

  ## What `/health` now returns
  
  `GET /health` (unauthenticated; served over the connector's Unix socket, reached via
  `curl --unix-socket /tmp/pentest-agent-1.sock http://x/health`):

  - **`200`** `{"status":"healthy", "grpc_transport":"up", ...}` when the gRPC transport is live.
  - **`503`** with `status` ∈ `starting` / `stopped` / `connecting` / `reconnecting` otherwise.
  - Body also carries `ever_connected`, `last_disconnect_reason`, `reconnection_attempts`,
    `total_disconnects`, `heartbeat_rtt_last_ms`, `uptime_seconds`, `tools_loaded`, `tenant`,
    `connector`.

  Classification (pure `classify_health`, fully unit-tested):
  
  | Condition (from `get_stats()`) | status | HTTP |
  |---|---|---|
  | runner not yet created | `starting` | 503 |
  | `running == false` | `stopped` | 503 |
  | running, never connected (`last_connected_at_ms` null) | `connecting` | 503 |
  | running, connected, no newer disconnect | `healthy` | **200** |
  | running, `last_disconnected_at_ms > last_connected_at_ms` | `reconnecting` | 503 |

  ## Implementation

  - `api_routes.rs`: new `HealthState`, pure `classify_health(stats, tools_loaded, tenant, connector) -> (StatusCode, Value)`, `get_health` handler, `create_health_route` — on a **separate
  un-authed router** (a probe must work without the Bearer token), merged as a sibling so it never inherits `/api/*`'s auth middleware.
  - `mod.rs`: added a shared `runner: Arc<RwLock<Option<Arc<ConnectorRunner>>>>` on `LiveViewConnector`, reused in the `PickConnector` literal so the existing write populates the same Arc
  `/health` reads. Health route built before the runner exists → reports `starting` until connected.
  - `liveview_server.rs`: removed the static `/health` from the base router (now provided by the merged health route; a duplicate would panic axum at merge).

  ## Probe wiring (recommendation; NOT shipped here)
  
  Readiness SHOULD consume `/health` (503 de-routes a not-connected pod). **Liveness should NOT**
  — a transient `reconnecting`/503 would crash-loop a pod the SDK is already recovering. The
  init-dev consumer wiring (readiness probe + `newdev:status`) is tracked in **Strike48/init-dev#335**.

  ## Testing
  
  **Unit:** 14 `classify_health` tests (all 5 states + malformed-input fail-safe + a
  production-shape `get_stats()` fixture to catch SDK key/type drift + the `==` timestamp
  boundary + a body/status consistency regression). `cargo fmt --check`, `cargo clippy -D warnings`,
  full `pentest-ui` suite (93 tests) all green.

  **Live (deployed on newdev k3s via `build:pick`, verified end-to-end):**
  - `healthy`/200 with real SDK data (`heartbeat_rtt_last_ms`, `tools_loaded: 115`, identity). ✅
  - **Controlled drop** (scaled Studio to 0): `/health` → `reconnecting`/503, `grpc_transport: down`,
    real `last_disconnect_reason`, climbing disconnect counts. ✅
  - **Recovery** (Studio back, Pick untouched): `503 → 200`; transport reconnected AND tools
    re-registered on their own (`total_connectors 0 → 1`, ~50–60s). ✅
  - `/api/status` still returns `401` without token — confirms moving `/health` off the base
    router did NOT leak auth-bypass to `/api/*`. ✅

  ## Known boundaries (intentional; documented in code)

  - `/health` reflects **transport/heartbeat only**. It cannot report registration/approval/JWT
    state (private SDK fields, no accessor in `strike48-connector` 0.4.4), so it can't distinguish
    *approved-active* from *pending-approval* — the Studio Gateways UI / connector registry remain
    the source of truth for approval. `tools_loaded` is honestly named ("loaded locally", not
    "registered with server").
  - **Silent-stall blind spot:** a connector whose stream is dead but not marked disconnected keeps
    `running == true`, so `/health` would read `healthy`. `get_stats()` exposes heartbeat *RTT*, not
    *staleness*. Detecting this needs an SDK accessor — folded into **Strike48/sdk-rs#59** (which
    also covers the silent-stall auto-recovery; the *clean-drop* recovery already works, verified above).
  - `bin/file_browser_connector.rs`'s `/health` is left as static `"OK"` — separate binary, no
    connector/runner state.

  ## Related
  - Parent hub: matrix#3216
  - init-dev#335 — wire the k8s readiness probe + `newdev:status` to consume `/health`
  - sdk-rs#59 — SDK silent-stall recovery + `get_stats()` heartbeat-staleness/session accessors

### DIFF
--- a/crates/ui/src/liveview_connector/api_routes.rs
+++ b/crates/ui/src/liveview_connector/api_routes.rs
@@ -75,7 +75,9 @@ use pentest_core::aggression::AggressionLevel;
 use pentest_core::config::ConnectorConfig;
 use pentest_core::matrix::MatrixChatClient;
 use serde::{Deserialize, Serialize};
+use serde_json::Value;
 use std::sync::Arc;
+use strike48_connector::ConnectorRunner;
 use tokio::sync::RwLock;
 
 use super::ScanState;
@@ -86,6 +88,21 @@ pub struct ApiState {
     pub scan_state: Arc<RwLock<Option<ScanState>>>,
     pub config: Arc<RwLock<ConnectorConfig>>,
     pub matrix_client: Arc<RwLock<Option<MatrixChatClient>>>,
+}
+
+/// State for the unauthenticated `/health` route.
+///
+/// Deliberately separate from [`ApiState`] and its Bearer-token `auth_middleware`:
+/// a health/readiness probe must be reachable without credentials, and this state
+/// carries no secrets — only the shared handle to the SDK connector runner, whose
+/// `get_stats()` snapshot is the source of transport/heartbeat truth.
+///
+/// The runner is created later (inside `LiveViewConnector::connect_and_run`), so the
+/// `Option` is `None` from server start until the connection loop populates the SAME
+/// `Arc` this state holds — the pre-connect window reports `status: "starting"`.
+#[derive(Clone)]
+pub struct HealthState {
+    pub runner: Arc<RwLock<Option<Arc<ConnectorRunner>>>>,
 }
 
 /// Request body for aggression level adjustment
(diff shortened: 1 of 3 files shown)


jtomek-strike48 (commented):
## Review: verified empirically (checked out 39f74d4, built, ran tests, mutation-tested)

Reviewed against the 8-lens rubric. Checked out the PR head into an isolated worktree, built with the real feature set, ran the tests, and neutered the guards to confirm they fail. Leading with what I verified sound so the (non-blocking) notes are in context.

### Verified sound

- **Tests pass at head:** 14 `classify_health` tests green with `--features connector,desktop` (the set the headless/desktop apps resolve).
- **Guards are genuine (Lens 1, mutation-tested):**
  - Inverting the `disconnected_at > connected_at` reconnecting branch turned `health_reconnecting_when_dropped` + `health_503_state_asserts_full_body` RED.
  - Making the `running` parse fail dangerous (`unwrap_or(true)`) turned `health_malformed_running_fails_safe_to_stopped` RED.
  - Restored: all 14 green.
- **Runs in required CI (Lens 2):** `pentest-ui` is a workspace member and every app enables it with `connector,shell-ws`, so `cargo test --workspace` (Linux Test lane) executes these tests.
- **Claims trace to code (Lens 3):** the `health_full_shape_get_stats_payload_is_healthy` fixture matches `strike48-connector` 0.4.4 `get_stats()` field-for-field (connector.rs:1849-1885), so it is a real SDK-drift catcher. The shared-Arc, lock-drop-before-await, and axum-merge doc-comments all check out against the code.
- **Consumer trace (Lens 5):** `self.runner` is one Arc shared by `HealthState` and `PickConnector`, populated at mod.rs:759, so `/health` observes the runner once connected (no "permanently starting" bug). `get_health` drops the outer read guard before awaiting `get_stats()`. The init-dev headless connector compiles this path (`apps/headless` = connector,shell-ws,desktop), so it is relevant there.
- **Fail-safe (silent-failure sweep):** every malformed/missing stats field defaults toward 503; no `.unwrap_or` produces a false-healthy 200. The documented `running==true` silent-stall blind spot is an SDK 0.4.4 limitation (no accessor), not a parsing bug here.
- **AC traceability (Lens 6):** transport/heartbeat state met; registration/approval/JWT state honestly scoped out and tracked in sdk-rs#59.
- **Hygiene (Lens 8):** 0 commits behind main; no PII/secret leakage in the diff; leaving `file_browser_connector.rs` static is a disclosed, defensible boundary.

Note: the isolated `cargo test -p pentest-ui` failure some may hit is pre-existing #260 (invalid feature combo in `pentest-tools/wifi_scan_detailed.rs`), reproduces on base main, and is unrelated to this PR.

### Non-blocking notes (optional test-coverage polish)

1. The `get_health` handler's `Some(runner)` path (session accessors -> body) is not asserted end-to-end; only the pure `classify_health` and the `None` branch are. A wrong session value would ship untested.
2. Type-malformed optional metrics (`uptime_seconds:"5"`, negative `reconnection_attempts`, string `heartbeat_rtt_last_ms`) default correctly via `unwrap_or` but aren't pinned by a test.
3. No explicit concurrency test for a runner swap during the `get_stats()` await (the lock-drop pattern is sound; just unproven by test).

### Recommendation

Approve. Correct, fail-safe, honestly scoped, comments trace to code, tests are mutation-verified guards, and it is live-verified on newdev per the PR body. The three notes are optional follow-ups, not change requests.

jtomek-strike48 (approved):
Reviewed at head `39f74d4`, base `86cf62d`. Checked out into an isolated worktree, built `pentest-ui --features connector,desktop`, and ran the suite. Approving; findings below are all non-blocking.

**Verified sound (empirically, not by reading):**

- **Tests genuinely guard the fix.** 14/14 health tests pass. I mutation-tested three ways and confirmed RED each time: flipping the reconnecting boundary `>` to `<` reds `health_reconnecting_when_dropped`; flipping the healthy arm 200 to 503 reds both healthy-path tests; replacing `ever_connected`'s `as_u64().is_some()` with a type-agnostic present-check reds `health_malformed_connected_ts_is_consistent_body_and_status`.
- **CI actually runs them.** `cargo tree -e features` confirms a `--workspace` build activates `pentest-ui`'s `connector`/`liveview` features via the app crates, so the required Linux `Test` job compiles and runs these. The macOS lane is package-scoped and correctly unaffected.
- **SDK contract matches exactly.** Every key `classify_health` reads exists in `strike48-connector` 0.4.4 `get_stats()` with the asserted types; the production-shape fixture is faithful.
- **No duplicate-route panic, auth scoping correct.** `/health` is merged as a separate un-authed router; `auth_middleware` is layered on the api router pre-merge, so `/health` never inherits the Bearer gate. Matches your live 401 check on `/api/status`.
- **Boundary shapes hold.** `last_connected_at_ms: 0` and `last_disconnected_at_ms: 0` both classify `healthy`/200 correctly.
- **Limitation claims are honest.** Confirmed the silent-stall blind spot in the SDK (stream marked disconnected only after `heartbeat_timeout`, so `running` stays true until then) and that registration state has no accessor on `ConnectorRunner`.

**SHOULD-FIX (non-blocking, fast-follow ok): no logging on the 503 path.** `get_health` returns 503 in four states with no local trail. A `tracing::warn!` carrying `status` / `last_disconnect_reason` / `reconnection_attempts` would help operators correlate probe de-routes with the underlying cause. One caveat: a readiness probe polls frequently, so log on state transition (or throttle) rather than on every poll to avoid spam during a long reconnect.

**LOW:** (1) The silent-stall blind spot is documented in the PR body but not in the `classify_health` boundary doc-comment; worth a line so it survives the merge (tracked in sdk-rs#59 anyway). (2) No test pins the `*_at_ms: 0` boundaries; behavior is correct, so this is coverage-completeness only.

jtomek-strike48 (approved):
## Hermes Review
**Verdict: LGTM**

Clean, well-reasoned change. The `/health` route is correctly split onto its own unauthenticated, state-carrying router (never inherits the Bearer `auth_middleware`), the runner `Arc` is genuinely shared with `PickConnector` so both observers see the same snapshot, and the old unconditional `"OK"` route was removed from `liveview_server.rs` to avoid the axum duplicate-route panic. `classify_health` is pure and side-effect-free, and the test coverage is excellent — the full transport/heartbeat state table plus boundary (`==` timestamps), malformed-input fail-safe, and body/status-consistency regressions are all pinned.

### Warnings
- **liveview_server.rs:318 (behavior change)** — `/health` now returns `503` for starting/stopped/connecting/reconnecting, whereas it previously always returned `200 "OK"`. If any existing k8s manifest, load balancer, or monitor currently uses `/health` as a **liveness** probe, this becomes a breaking change: a transient `reconnecting` would trigger pod restarts while the SDK is already retrying with backoff. The PR doc calls this out (readiness-only, no manifest ships here) — please confirm no live liveness probe points at `/health` before/at rollout, or the semantics flip silently in prod.

### Suggestions
- **api_routes.rs (`get_health`)** — the body is unauthenticated and exposes `tenant`, `connector`, and `last_disconnect_reason`. Identity is low-risk, but `last_disconnect_reason` is a passthrough of an SDK error string that could surface internal hostnames/TLS detail to an uncredentialed caller. Consider whitelisting/normalizing it to a coarse enum if the endpoint is ever network-reachable beyond the pod.

### Looks Good
- Lock discipline in `get_health`: clones the inner runner `Arc` and drops the read guard before awaiting `get_stats()`, so `/health` never blocks a runner swap.
- `ever_connected` deliberately keyed on `as_u64()` to match the classifier exactly — the self-contradictory-payload regression test (`health_malformed_connected_ts_is_consistent_body_and_status`) is a nice touch.
- Boundary (pick#295) documented honestly: `/health` reflects transport + heartbeat only, not approval/JWT-session state, with the sdk-rs#59 follow-up noted.

Issue #158, opened 2026-06-24
Add Safety Check feature to Quick Actions for public WiFi security validation


jtomek-strike48:
## Summary

Add a **Safety Check** feature to the Quick Actions dashboard that validates whether the operator's local network environment is secure. Target audience: operators with low security expertise working from public WiFi (airports, coffee shops) who need simple "am I safe?" validation.

**Demo value:** Visual network map with threat indicators, interactive graphs, and actionable security recommendations make this an excellent feature for showcasing Pick's capabilities.

---

## User Story

**As an operator** working from a coffee shop or airport,  
**I want** a quick safety check of my network environment,  
**So that** I know if it's safe to conduct penetration testing or if I should enable VPN/take other precautions.

---

### 1. DNS Integrity Check - HIGH PRIORITY
**Why:** Detects DNS hijacking/poisoning common in malicious hotspots.

**Implementation:**
- Query 5-10 known-good domains (google.com, github.com, cloudflare.com, 1.1.1.1, dns.google)
- Compare returned IPs against expected values (hardcoded or from trusted source)
- Detect captive portals (HTTP redirects on HTTPS requests)

**Output:** [PASS] DNS is safe | [WARN] DNS may be compromised | [FAIL] DNS hijacking detected

**Tools:**
- `std::net::ToSocketAddrs` for basic DNS resolution
- `reqwest` for HTTPS requests to detect captive portals
- `trust-dns-resolver` for more advanced DNS queries if needed

---

### 2. Router Threat Intelligence - HIGH PRIORITY
**Why:** Flags if you're behind a known malicious router/gateway.

**Implementation:**
- Get default gateway IP via `ip route` or platform APIs
- Get DNS server IPs from `/etc/resolv.conf` or platform APIs
- Lookup both IPs in threat intelligence feeds:
  - VirusTotal API (already available via MCP - `mcp__virustotal__get_ip_report`)
  - AbuseIPDB (available via MCP - `mcp__threatintel__abuseipdb_check`)
  - Optional: GreyNoise, AlienVault OTX via MCP

**Output:**
- [PASS] Router is clean (no threat intel hits)
- [WARN] Router has medium threat score (X detections)
- [FAIL] Router flagged as malicious

**Visual:** Show router IP + geolocation on world map with threat score

**Tools:**
- `default-net` crate for gateway detection (already in project)
- MCP threat intelligence tools via Strike48 SDK
- `reqwest` for direct API calls as fallback

---

[...]

PR #304, opened 2026-07-24
fix(safety_check): stop reporting the host's own footprint as a compromised network

### LINKED ISSUE #303: safety_check false-positives: reports host's own footprint as a compromised network
## Summary

`safety_check` (the pre-engagement "is this network safe to operate from" check) renders alarming verdicts that read to a user on their own machine as "you have been compromised" - when in fact it is seeing **Pick's own host footprint**. Three distinct mechanisms:

1. **Multi-interface self-reflection** (`network_map.rs`): device discovery excluded only the single default-route local IP. On a normal operator box with VPN/container/mesh interfaces (docker0, virbr0, Tailscale, a second NIC), the host's *other* addresses were enumerated as foreign "devices" on the network.
2. **DNS check over-firing** (`dns_check.rs`): any unexpected-IP or resolution anomaly mapped straight to `Failed/Critical` -> overall `Unsafe`. A VPN, split-horizon DNS, Pi-hole/NextDNS, or corporate resolver - all benign, all common - tripped a hard "compromised" verdict.
3. **Verdict copy** (`report.rs`): headline "UNSAFE" plus "this network is likely malicious or severely compromised" gave no signal that the verdict is about the *network environment*, not the user's device.

There is also a **latent** fourth issue: `classify_threat` flags a host with >=10 open ports as Suspicious, but `open_ports` is empty today. When per-host port enrichment lands, Pick will scan its own box and flag the listeners its installed tools opened (Metasploit handlers, BloodHound's Neo4j, etc.).

## Root cause

An offensive connector that also self-assesses network safety must **subtract its own footprint** before rendering a verdict. This is deconfliction (the same pattern BAS platforms use to tag simulated activity), not a reason to adopt a separate DFIR product.

## Empirical repro

`cargo test -p pentest-tools safety_check::tests::live_report -- --ignored --nocapture` on a multi-interface host shows the host's own docker/VM/mesh addresses among "devices," and the hard `Unsafe` verdict originates from the DNS path on VPN/split-DNS networks.

## Acceptance criteria

- [ ] Host's own interface addresses (all interfaces, not just default) never appear as foreign devices.
- [ ] A single uncorroborated DNS anomaly yields `Caution`, not `Unsafe`; captive-portal / total-DNS-loss remain `Critical`.
- [ ] Verdict copy is scoped to the network environment and never implies the host itself is compromised.
- [ ] Regression tests that fail against the pre-fix behavior.
- [ ] Breadcrumb for the latent port-scan self-flag before port enrichment ships.

**Recommended agent:** rust-reviewer (security-sensitive verdict logic).

jtomek-strike48:
Closes #303.

## Summary

- `safety_check` is a pre-engagement "is this network safe to operate **from**" check, but on a normal operator machine it read as "you have been compromised" because it was seeing **Pick's own host**, not a threat. This fixes the three mechanisms behind that false positive so the feature stays useful instead of being suppressed.
- **Self-attribution:** device discovery now excludes *all* local interface IPs (via the platform's existing `SystemInfo::get_network_interfaces`), not just the single default-route address. A box with VPN/container/mesh interfaces (docker0, virbr0, Tailscale, a second NIC) no longer surfaces its own addresses as foreign "devices."
- **DNS severity calibrated to reality:** a lone uncorroborated anomaly (VPN, split-horizon DNS, Pi-hole/NextDNS, corporate resolver) is now `Caution`, not `Unsafe`. Hard `Critical` is reserved for corroborated signals (total DNS loss, captive portal).
- **Verdict copy** is scoped to the network environment and no longer implies the host itself is compromised.

## Security trade-off (please scrutinize)

The DNS change means a genuine lone DNS-hijack (single unexpected IP, no captive portal, connectivity otherwise fine) now surfaces as `Caution` rather than `Unsafe`. This is deliberate: on real operator machines the benign causes (VPN/split-DNS/filtered resolvers) vastly outnumber hijacks, and a real hijack typically corroborates via the captive-portal or connectivity signals that still fire `Critical`. Flagging it here so the reviewer weighs it explicitly rather than finding it buried in the diff.

## Not in scope (deliberate)

The latent per-host port-scan self-flag (`classify_threat`, currently inert because `open_ports` is always empty) is left with a `TODO(self-attribution)` breadcrumb rather than built out now (YAGNI). The TODO points at the same self-filter pattern so Pick-owned listeners get subtracted before port enrichment ships.

## Tracked follow-ups

- #305 - escalate the DNS verdict to Critical when a *majority* of known-good domains resolve to unexpected IPs. Deferred from this PR to keep scope tight; it hardens the trade-off called out above without reintroducing the false positive.
- Parent feature: #158 (introduced `safety_check`).

## Test plan

- [x] `cargo clippy --workspace --locked --features pentest-platform/desktop-pcap -- -D warnings` clean (full CI-equivalent workspace lane; exit 0, no error lines)
- [x] `cargo test -p pentest-tools --lib safety_check` - **55 passed, 0 failed, 2 ignored** (48 original + 7 added during self-review)
- [x] Guard tests verified to **fail against the pre-fix behavior** (mutation-checked): neutering `filter_foreign_hosts` to ignore `own_ips` turns both interface-exclusion tests RED; restored -> green. The DNS test asserts `Warning`/`!=Critical` where the old code returned `Failed/Critical`.
- [x] Self-review (`/pre-review`) run: 3 independent review agents converged on `collect_own_ips()` being silently degrading + untested -> hardened (warn-on-failure, pure testable helper, loopback/IPv6-link-local defense, 5 new unit tests). See commit 90d570c.
- [ ] Reviewer: confirm the DNS severity trade-off above is acceptable.
- [ ] Optional live check on a multi-interface/VPN host: `cargo test -p pentest-tools safety_check::tests::live_report -- --ignored --nocapture`

### DIFF
--- a/crates/tools/src/safety_check/dns_check.rs
+++ b/crates/tools/src/safety_check/dns_check.rs
@@ -74,7 +74,10 @@ pub async fn check_dns_integrity() -> anyhow::Result<CheckResult> {
         }
     }
 
-    // Detect captive portal: all domains resolve to same IP
+    // Detect captive portal: all domains resolve to same IP. This is a
+    // *corroborated* signal (multiple independent domains collapsing to one
+    // address), so unlike a lone unexpected-IP it is strong enough to fail.
+    let mut captive_portal = false;
     if connectivity_resolutions.len() >= 2 {
         let first_ip = connectivity_resolutions
             .values()
@@ -86,6 +89,7 @@ pub async fn check_dns_integrity() -> anyhow::Result<CheckResult> {
                 .values()
                 .all(|ips| ips.contains(&first));
             if all_same {
+                captive_portal = true;
                 issues.push(format!(
                     "Captive portal detected: all domains resolve to {}",
                     first
@@ -94,14 +98,15 @@ pub async fn check_dns_integrity() -> anyhow::Result<CheckResult> {
         }
     }
 
-    // Determine status
-    let (status, severity) = if !connectivity_ok || !issues.is_empty() {
-        (CheckStatus::Failed, Severity::Critical)
-    } else if validated_domains < 3 {
-        (CheckStatus::Warning, Severity::Medium)
-    } else {
-        (CheckStatus::Passed, Severity::Info)
-    };
+    // Classify the outcome into a (status, severity). Extracted to a pure
+    // helper so the severity calibration is unit-testable without live DNS.
+    let has_only_soft_anomalies = connectivity_ok && !captive_portal && !issues.is_empty();
+    let (status, severity) = classify_dns_outcome(
+        connectivity_ok,
+        captive_portal,
+        !issues.is_empty(),
+        validated_domains,
+    );
 
     let details = if issues.is_empty() {
         format!(
@@ -110,6 +115,13 @@ pub async fn check_dns_integrity() -> anyhow::Result<CheckResult> {
         )
     } else if !connectivity_ok {
         "No DNS connectivity - unable to resolve any domains. Check network connection.".to_string()
+    } else if has_only_soft_anomalies {
+        format!(
+            "DNS returned unexpected results for some domains. This is common on VPNs, \
+             filtered DNS (Pi-hole/NextDNS), or corporate networks and is usually benign, \
+             but verify if unexpected:\n{}",
+            issues.join("\n")
+        )
     } else {
         format!("DNS integrity issues detected:\n{}", issues.join("\n"))
     };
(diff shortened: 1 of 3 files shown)


cchao-devo (approved):
Review: Pick PR #304 — fix(safety_check): stop reporting the host's own footprint as a compromised network (#303)

  Verdict: ✅ Approve — with the DNS trade-off explicitly acknowledged (it's sound and correctly bounded). Well-factored self-attribution fix with no threat-masking path, and a
  deliberate DNS severity recalibration that's narrowly scoped and tracked. Gates green. No blockers.

  Lenses run: the real pr-review-toolkit silent-failure-hunter (both failure directions) + code-reviewer, plus my own empirical runs of classify_dns_outcome, own_ips_from_interfaces,
  filter_foreign_hosts. Verified in pick_08/pick; pick_07 untouched.

  Gates (on pr-304)

  - cargo fmt --all --check clean · cargo clippy --workspace --all-targets -- -D warnings clean · safety_check 55/0 (2 ignored = live --ignored tests)

  Self-attribution fix — sound, no threat-masking (the dangerous direction, checked)

  The exclude set is strictly this-host's-own: own_ips = platform.get_network_interfaces() (our NICs) + local_ip (default route). Verified by running the helpers — there's no code path
  by which an attacker-controlled address enters own_ips (an attacker can't add an interface to our host). The one theoretical masking case (a foreign device whose IP literally equals
  one of our interface IPs) requires an on-LAN IP conflict with our own box — not a realistic attacker position, and correctly scoped. Nice details: CIDR-suffix stripping, loopback +
  IPv6-link-local defense-in-depth (even if the backend mislabels is_loopback), malformed addrs dropped with a breadcrumb (no panic), extracted into pure testable helpers. The
  IPv4-mapped-v6 own address doesn't collision-suppress a plain-v4 peer (errs toward keeping peers visible — safe direction).

  Silent degradation — safe

  collect_own_ips returns empty on interface-enum failure and warns (appropriate — an incomplete self-filter silently weakens the check, so it must be observable).
  own_ips.insert(local_ip) runs unconditionally after, so a failure degrades to exactly the old behavior (exclude gateway + default IP), never excludes-nothing, never crashes.

  DNS severity recalibration — the deliberate trade-off, assessed and accepted

  classify_dns_outcome (pure, testable) confirmed by running it:
  - Corroborated signals still hard-fail: total DNS loss → (Failed, Critical), captive portal → (Failed, Critical). ✓
  - Lone uncorroborated anomaly → (Warning, Medium) → Caution. This is the downgrade.
  - No bad→Safe path: has_anomalies forces at least Warning; the gap is strictly Unsafe→Caution, never masked to Safe.

  My assessment of the trade-off the author asked me to weigh: accept it. A lone DNS anomaly on an operator machine is overwhelmingly benign (VPN/split-DNS/Pi-hole/corporate resolver);
  a false "compromised" verdict gets the whole feature suppressed — worse than a Caution. The one true-positive weakening is a full DNS hijack (connectivity fine, anomalies but not
  all-collapse-to-one-IP → not captive portal) which now reads Caution instead of Unsafe. That's real but low-likelihood (a real hijack usually corroborates via
  captive-portal/connectivity signals that still fire Critical), the detail copy is honest about the ambiguity, and the escalation (majority-of-domains-unexpected → Critical) is
  explicitly filed as #305. Reasonable to ship with that follow-up tracked — but worth confirming in the merge that the "full-hijack = Caution until #305" window is acceptable to
  whoever owns the safety posture. (That's the author's flagged question; my answer is yes.)

  TODO(self-attribution) on classify_threat — genuinely inert

  Verified open_ports is always Vec::new() (no per-host port-scan wiring feeds it), so classify_threat always returns Safe today — the latent self-flag can't fire. Leaving it as a TODO
  gated on future port-enrichment (with a provenance-based mitigation pointer + memory-note reference) is safe and correct YAGNI.

  init-dev relevance

  Low — safety_check is the pre-engagement network-safety check. The self-reflection false-positive would've fired on operator machines generally; not a connector-contract or
  cluster-deployment surface. init-dev's headless connector isn't the primary consumer.

Issue #162, opened 2026-06-24
Add Authorization & Multi-Audience Path Testing to WebApp/API Specialists


jtomek-strike48:
# Add Authorization & Multi-Audience Path Testing to WebApp/API Specialists

**Issue Type:** Enhancement
**Component:** Pick Core (WebApp specialist + Api specialist; shared authz engine)
**Effort:** Large (1-2 weeks)
**Priority:** P1
**Dependencies:** WebApp taint/variant (pick#147); complements browser automation (matrix#2377)
**Source:** Customer engagement requirement - "frontend, both client-facing and internal, same URL different paths"
**Existing Issues:** Extends `pick#99` WebApp/Api specialists; parent spec `matrix#2381`; integration target `matrix#2422` (vulnerable web range)

---

## Problem Statement

A customer needs black-box + penetration testing of a frontend that serves
**two audiences from the same host on different paths** - e.g. `app.example.com/`
(client-facing) and `app.example.com/admin` or `/internal` (privileged/internal).
The core risk in this topology is **broken access control**: can a low-privileged
(or unauthenticated) user reach internal paths, internal APIs, or other tenants'
data?

Pick's current specialists do not address this. They reason about injection-class
bugs (SQLi/XSS/SSRF/SSTI/XXE) on **discovered** endpoints, largely from a single
session's perspective. They do **not**:

- Test the **same endpoint across multiple identities** (unauth vs user vs admin
  vs tenant-A vs tenant-B) to detect authorization boundary failures.
- Systematically discover **internal/privileged paths** that share a host with the
  public app (the exact "same URL, different paths" case).
- Detect **BOLA/IDOR** (object-level authz), **BFLA** (function-level authz),
  forced browsing, or vertical/horizontal privilege escalation.

Broken Access Control is OWASP Web Top 10 #1 (A01:2021) and the basis of OWASP
API Top 10 #1 (BOLA) and #5 (BFLA). It is currently a coverage hole.

---

## Proposed Solution

Add a shared **multi-identity authorization testing engine** used by both the
WebApp and Api specialists. The engine replays requests across a matrix of
identities and diffs the responses to find authorization boundary violations.

### Part A: Multi-Identity Session Matrix

**File:** `crates/core/src/specialist/authz/identity_matrix.rs` (new)

```rust
/// A test identity (credential set + resulting session) used for differential authz testing.
#[derive(Debug, Clone)]
pub struct TestIdentity {
[...]

PR #317, opened 2026-07-25
feat(#162): differential-authz identity model, store, orchestrator advertisement + specialist prompts

jtomek-strike48:
## Summary

Delivers the Pick-side reference flow for differential authorization testing (BOLA/BFLA/forced-browsing) behind pick#162 and the pick#164 customer-attack-surface epic. Operator-provided test identities now load into a connector-side store, are advertised (references only) to the Red Team orchestrator, and are consumed by the WebApp/API specialist prompts - all while keeping the customer's credentials off the Matrix/LLM boundary.

## Re-scope note (please read before reviewing)

The issue proposed a Rust differential-authz *engine* (`crates/core/src/specialist/authz/*.rs`). Pick's specialists are **prompt-driven markdown skills**, not Rust attack logic - so this PR delivers the pieces that were actually missing (identity intake, connector-side secret custody, orchestrator advertisement, prompt wiring) and keeps execution in the LLM + existing tools. Full rationale on pick#162. This is deliberately **not** the literal module layout from the issue.

## What's in this PR (3 commits)

1. **Identity model + connector-side store** - `TestIdentity { label, role, tenant }` reference on `SpecialistContext` (serde-defaulted); new `identity.rs` with `SessionMaterial` (secret; non-`Serialize`; redacts in Debug/Display) + `IdentityStore` on `ToolContext` (resolve by label; Debug shows labels, never material).
2. **Standalone file population** - loads a gitignored `identities.json` at startup (missing = no-op; malformed/unknown-role = loud error); kept out of `settings.json` by design; `identities.example.json` documents the schema.
3. **Orchestrator advertisement (Hop A) + prompt wiring (Hop C)** - connector advertises identity *references* via `InstanceMetadata` (mirrors `host_interfaces`; consumed by matrix#3354); WebApp/API prompts consume the `identities` matrix via `identity_ref`, run the differential replay + BOLA object/tenant swap + co-hosted `/admin` discovery, treat 401/403 as a working control, and report reduced coverage below two identities.

## Security design

- `SessionMaterial` is non-serializable and never written to `settings.json` (plaintext) - credentials rest only in the operator-owned, gitignored `identities.json`.
- Guard tests: `TestIdentity` never serializes a secret; the advertised metadata value carries labels/roles but no session material; redaction holds under nested Debug.
- Type 1 (operator-provided) only; Type 2 (discovered) is a separate policy (pick#313).

## Cross-repo dependency

- **matrix#3354** - the Red Team orchestrator must read the advertised references and forward them to `spawn_specialist`. This is the keystone that activates the feature end-to-end; the metadata contract is spelled out there.

## Deferred (tracked)

pick#312 (StrikeKit provisioning) - pick#314 (execute_command injection) - pick#315 (webwright injection) - pick#316 (Settings UI) - pick#313 (Type 2 flagging)

## Test plan

- [x] `cargo check --workspace --locked --features pentest-platform/desktop-pcap`
- [x] `cargo clippy --workspace --locked --features pentest-platform/desktop-pcap -- -D warnings`
- [x] `cargo test --workspace ...` - all suites pass (see caveat below)
- [x] identity unit tests: redaction (incl. nested Debug), store resolve/miss, file loader (missing/valid/malformed/unknown-role), metadata refs-carry-no-secret
- [x] serde back-compat: context without `identities` deserializes
- [x] prompt-wiring guard: WebApp/API prompts reference the identity matrix (`identities`/`identity_ref`/reduced-coverage)
- [x] gitignore: `identities.json` ignored, `identities.example.json` trackable

### Caveat (Lens 4 - pre-existing, not introduced)

The full `--workspace` run intermittently hit a **pre-existing flaky SIGABRT** ("stack smashing detected") in the pcap-backed `tool_integration::port_scan_requires_host_flat` under heavy parallel load. It passes **30/30 in isolation**; this branch touches no pcap/port_scan/platform code. Tracked as pick#318.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -44,6 +44,8 @@ logs/
 
 # Operator test-identities for differential-authz testing (pick#162).
 # Holds real target credentials (cookies/JWT/headers) - never commit.
+# The loader (identities_file_path) prefers a cwd `identities.json` when present,
+# so a file copied from the template next to it here is both loaded AND ignored.
 # The committed template `identities.example.json` is deliberately re-included.
 identities.json
 identities.*.json
--- a/.gitignore
+++ b/.gitignore
@@ -42,6 +42,13 @@ logs/
 .env.bak-*
 .env.*.bak
 
+# Operator test-identities for differential-authz testing (pick#162).
+# Holds real target credentials (cookies/JWT/headers) - never commit.
+# The committed template `identities.example.json` is deliberately re-included.
+identities.json
+identities.*.json
+!identities.example.json
+
 # AI tooling local overlays — private, operator-specific, never committed.
 # See CLAUDE.md "Local Overrides" section.
 CLAUDE.local.md
--- a/.gitignore
+++ b/.gitignore
@@ -44,6 +44,11 @@ logs/
 
 # Operator test-identities for differential-authz testing (pick#162).
 # Holds real target credentials (cookies/JWT/headers) - never commit.
+# ROLLOUT WARNING: injected credentials are passed to tools via argv
+# (e.g. `curl -H "Authorization: Bearer <token>"`), so they are readable from
+# `ps` / `/proc/<pid>/cmdline` by any local user for the child's lifetime. Do
+# NOT provision real credentials on a shared or multi-tenant connector host
+# until pick#335 moves the header off argv (`curl -H @file`).
 # The loader (identities_file_path) prefers a cwd `identities.json` when present,
 # so a file copied from the template next to it here is both loaded AND ignored.
 # The committed template `identities.example.json` is deliberately re-included.
--- a/crates/core/src/evidence.rs
+++ b/crates/core/src/evidence.rs
@@ -336,10 +336,21 @@ mod tests {
             ProbeCommand::from_exact("nmap -sV 192.168.1.1"),
             "Nmap scan report",
         );
-        let node = fixture().with_provenance(prov.clone());
+        let node = fixture().with_provenance(prov);
         let wire = serde_json::to_value(&node).unwrap();
         let back: EvidenceNode = serde_json::from_value(wire).unwrap();
-        assert_eq!(back.provenance, Some(prov));
+        // The wire-facing provenance fields survive the round-trip. The raw
+        // `command` is intentionally not serialized (#317 review) — it can carry
+        // an injected credential — so it comes back empty while the redacted
+        // `effective_command` is intact.
+        let back_prov = back.provenance.expect("provenance present");
+        assert_eq!(back_prov.underlying_tool, "nmap");
+        assert_eq!(back_prov.tool_version, "7.95");
+        assert_eq!(back_prov.probe_commands[0].command, "");
+        assert_eq!(
+            back_prov.probe_commands[0].effective_command,
+            "nmap -sV 192.168.1.1"
+        );
     }
 
     #[test]
--- a/crates/core/src/identity.rs
+++ b/crates/core/src/identity.rs
@@ -1,0 +1,230 @@
+//! Connector-side identity store for differential authorization testing
+//! (pick#162).
+//!
+//! The differential-authz method needs to replay the same request as several
+//! operator-provided identities (unauth / user / admin / tenant-A/B) and diff
+//! the responses. Those identities are authenticated by **secrets** - cookies,
+//! bearer tokens, or raw header bundles the operator supplies at engagement
+//! setup.
+//!
+//! # Why the secret lives here and not in [`crate::specialist_spawner::TestIdentity`]
+//!
+//! Pick runs standalone *and* through StrikeKit/Matrix. When spawned via Matrix,
+//! the specialist's [`SpecialistContext`](crate::specialist_spawner::SpecialistContext)
+//! is serialized and sent to the agent (the LLM reasons server-side; only tool
+//! *execution* returns to the connector). If the raw session material rode in
+//! that context it would transit to and rest on Matrix and sit in the LLM
+//! window - a leak surface for the *customer's* credentials.
+//!
+//! So the context carries only **references** (`TestIdentity { label, role,
+//! tenant }`), and the raw material stays connector-side in this
+//! [`IdentityStore`], keyed by `label`. The LLM references an identity by label
+//! in its tool calls; the connector resolves the label here and injects the
+//! auth locally at execution time. The secret never crosses the Matrix/LLM
+//! boundary.
+//!
+//! This handles **Type 1** credentials only: operator-*provided* identities for
+//! authorized testing. Type 2 (credentials *discovered* mid-engagement) are
+//! findings/evidence - logged and flagged, not redacted - and are out of scope
+//! here (see pick#313).
+//!
+//! Population is two-pathed: standalone (operator config/UI) ships now; the
+//! StrikeKit/Matrix provisioning path is pick#312.
+
+use std::collections::HashMap;
+
+/// Opaque session material (cookies, bearer token, or a raw header bundle) that
+/// authenticates a test identity when a specialist replays requests.
+///
+/// This is a secret. Its `Debug` and `Display` impls **redact** the value so it
+/// cannot leak into logs, `tracing` fields, or error messages - the
+/// secret-hygiene rule shared with pick#179. It is intentionally **not**
+/// `Serialize`: unlike the pattern-based [`crate::provenance::redact`] (which
+/// scrubs secrets out of tool output), this type keeps the secret off every
+/// serialized surface entirely. It is never placed in
+/// [`SpecialistContext`](crate::specialist_spawner::SpecialistContext); it lives
+/// only in the connector-side [`IdentityStore`]. Reach the raw value only via
+/// [`SessionMaterial::expose`], and only on the authentication path.
+#[derive(Clone, Default, PartialEq, Eq)]
+pub struct SessionMaterial(String);
+
+impl SessionMaterial {
+    /// Wrap operator-provided session material.
+    pub fn new(material: impl Into<String>) -> Self {
+        Self(material.into())
+    }
+
+    /// Borrow the raw material. This is the authentication path only - callers
+    /// that route the result into logs, evidence nodes, or tool output violate
+    /// the secret-hygiene contract.
+    pub fn expose(&self) -> &str {
+        &self.0
+    }
+
+    /// Whether any material was supplied. `true` for an anonymous identity.
+    pub fn is_empty(&self) -> bool {
+        self.0.is_empty()
+    }
+}
+
+impl std::fmt::Debug for SessionMaterial {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "SessionMaterial(<redacted {} bytes>)", self.0.len())
+    }
+}
+
+impl std::fmt::Display for SessionMaterial {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "<redacted {} bytes>", self.0.len())
+    }
+}
+
+/// Connector-local map from identity `label` to its [`SessionMaterial`].
+///
+/// Held on [`ToolContext`](crate::tools::ToolContext) so any tool can resolve a
+/// `label` referenced by the LLM to the real auth material and inject it at
+/// execution time. Never serialized; the whole point is that it does not leave
+/// the connector.
+#[derive(Clone, Default)]
+pub struct IdentityStore {
+    by_label: HashMap<String, SessionMaterial>,
+}
+
+impl IdentityStore {
+    /// An empty store (no identities provisioned).
+    pub fn new() -> Self {
+        Self::default()
+    }
+
+    /// Build a store from `(label, material)` pairs. Later entries win on a
+    /// duplicate label.
+    pub fn from_pairs<I, L>(pairs: I) -> Self
+    where
+        I: IntoIterator<Item = (L, SessionMaterial)>,
+        L: Into<String>,
+    {
+        let by_label = pairs
+            .into_iter()
+            .map(|(label, material)| (label.into(), material))
+            .collect();
+        Self { by_label }
+    }
+
+    /// Insert or replace the material for `label`.
+    pub fn insert(&mut self, label: impl Into<String>, material: SessionMaterial) {
+        self.by_label.insert(label.into(), material);
+    }
+
+    /// Resolve a label referenced by the LLM to its session material.
+    ///
+    /// Returns `None` when the label is unknown (typo, or an identity that was
+    /// never provisioned) - callers must treat a miss as "cannot authenticate
+    /// as this identity" and fail loudly rather than silently running
+    /// unauthenticated, which would misreport an authz result.
+    pub fn resolve(&self, label: &str) -> Option<&SessionMaterial> {
+        self.by_label.get(label)
+    }
+
+    /// Number of provisioned identities.
+    pub fn len(&self) -> usize {
+        self.by_label.len()
+    }
+
+    /// Whether no identities have been provisioned.
+    pub fn is_empty(&self) -> bool {
+        self.by_label.is_empty()
+    }
+}
+
+/// The store never reveals its secrets through `Debug` - it reports only the
+/// set of known labels (non-sensitive) and a count, so it is safe to include in
+/// a `#[derive(Debug)]` parent (e.g. `ToolContext`).
+impl std::fmt::Debug for IdentityStore {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        let mut labels: Vec<&str> = self.by_label.keys().map(String::as_str).collect();
+        labels.sort_unstable();
+        f.debug_struct("IdentityStore")
+            .field("labels", &labels)
+            .field("count", &self.by_label.len())
+            .finish()
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn session_material_redacts_in_debug_and_display() {
+        let secret = "super-secret-session-token";
+        let session = SessionMaterial::new(secret);
+
+        let debug = format!("{session:?}");
+        let display = format!("{session}");
+        assert!(
+            !debug.contains(secret),
+            "Debug must not leak the secret: {debug}"
+        );
+        assert!(
+            !display.contains(secret),
+            "Display must not leak the secret: {display}"
+        );
+        // Only the raw exposure path yields the value.
+        assert_eq!(session.expose(), secret);
+    }
+
+    #[test]
+    fn anonymous_identity_has_empty_material() {
+        let default = SessionMaterial::default();
+        assert!(default.is_empty());
+        assert_eq!(default.expose(), "");
+    }
+
+    #[test]
+    fn store_resolves_known_label() {
+        let store = IdentityStore::from_pairs([
+            ("user_a", SessionMaterial::new("Cookie: sid=aaa")),
+            ("admin", SessionMaterial::new("Authorization: Bearer zzz")),
+        ]);
+        assert_eq!(store.len(), 2);
+        assert_eq!(
+            store.resolve("admin").map(SessionMaterial::expose),
+            Some("Authorization: Bearer zzz")
+        );
+    }
+
+    #[test]
+    fn store_miss_returns_none_not_empty() {
+        // A miss must be distinguishable from "resolved to empty" so callers can
+        // fail loudly instead of silently replaying unauthenticated.
+        let store = IdentityStore::from_pairs([("user_a", SessionMaterial::new("Cookie: sid=a"))]);
+        assert!(store.resolve("nonexistent").is_none());
+    }
+
+    #[test]
+    fn insert_replaces_existing_label() {
+        let mut store = IdentityStore::new();
+        store.insert("user_a", SessionMaterial::new("old"));
+        store.insert("user_a", SessionMaterial::new("new"));
+        assert_eq!(store.len(), 1);
+        assert_eq!(
+            store.resolve("user_a").map(SessionMaterial::expose),
+            Some("new")
+        );
+    }
+
+    #[test]
+    fn store_debug_never_leaks_material_but_shows_labels() {
+        let store = IdentityStore::from_pairs([
+            ("user_a", SessionMaterial::new("secret-aaa")),
+            ("admin", SessionMaterial::new("secret-zzz")),
+        ]);
+        let debug = format!("{store:?}");
+        assert!(
+            !debug.contains("secret-aaa") && !debug.contains("secret-zzz"),
+            "Debug must not leak material: {debug}"
+        );
+        // Labels are non-sensitive and useful for diagnostics.
+        assert!(debug.contains("user_a") && debug.contains("admin"));
+    }
+}
--- a/crates/core/src/identity.rs
+++ b/crates/core/src/identity.rs
@@ -31,7 +31,49 @@
 //! Population is two-pathed: standalone (operator config/UI) ships now; the
 //! StrikeKit/Matrix provisioning path is pick#312.
 
+use crate::error::{Error, Result};
+use serde::Deserialize;
 use std::collections::HashMap;
+use std::path::{Path, PathBuf};
+
+use crate::specialist_spawner::{IdentityRole, TestIdentity};
+
+/// Environment variable naming the identities file, overriding the default
+/// location. Lets an operator (or `run-pentest.sh`) point at an engagement-
+/// specific file without touching the config dir.
+pub const IDENTITIES_FILE_ENV: &str = "PICK_IDENTITIES_FILE";
+
+/// One entry in the operator-provided identities file.
+///
+/// This DTO is the **one** boundary where session material legitimately enters
+/// the process from a persisted source: the file is operator-owned and
+/// gitignored (see `.gitignore` `.env*` / dedicated secrets convention). The
+/// raw `session` string is read here and immediately wrapped in the
+/// non-serializable [`SessionMaterial`]; it is never re-serialized.
+#[derive(Debug, Deserialize)]
+struct IdentityFileEntry {
+    label: String,
+    role: IdentityRole,
+    #[serde(default)]
+    tenant: Option<String>,
+    /// Raw session material (cookie/JWT/header bundle). Omitted/empty for an
+    /// anonymous identity.
+    #[serde(default)]
+    session: String,
+}
+
+/// Parsed identities file: the reference set (labels/roles/tenants) plus the
+/// connector-side store of their secrets.
+///
+/// Splitting the two is deliberate: `references` is safe to hand to the
+/// specialist context, `store` stays connector-local.
+#[derive(Debug, Default)]
+pub struct LoadedIdentities {
+    /// Reference set to place in [`SpecialistContext`](crate::specialist_spawner::SpecialistContext).
+    pub references: Vec<TestIdentity>,
+    /// Connector-local secret store.
+    pub store: IdentityStore,
+}
 
 /// Opaque session material (cookies, bearer token, or a raw header bundle) that
 /// authenticates a test identity when a specialist replays requests.
@@ -136,6 +178,68 @@ impl IdentityStore {
     }
 }
 
+/// Resolve the identities-file path: the [`IDENTITIES_FILE_ENV`] override if
+/// set, otherwise `identities.json` in the connector's config dir. The path is
+/// not required to exist - a missing file simply means no identities.
+pub fn identities_file_path() -> PathBuf {
+    if let Ok(p) = std::env::var(IDENTITIES_FILE_ENV) {
+        if !p.trim().is_empty() {
+            return PathBuf::from(p);
+        }
+    }
+    crate::settings::settings_dir().join("identities.json")
+}
+
+/// Load operator-provided identities from a JSON file.
+///
+/// A **missing** file is not an error - it yields empty [`LoadedIdentities`], so
+/// standalone runs without identities behave exactly as before. A file that
+/// exists but is malformed (bad JSON, unknown role) **fails loudly** rather than
+/// silently degrading to zero identities, which would misreport authz coverage.
+///
+/// The file schema is a JSON array of objects:
+/// `[{"label":"user_a","role":"user","tenant":null,"session":"Cookie: sid=..."}]`
+/// (`role` is one of `anonymous` | `user` | `privileged`; `tenant`/`session`
+/// optional). The file is operator-owned and must be gitignored - it holds real
+/// credentials.
+pub fn load_identities_from_file(path: &Path) -> Result<LoadedIdentities> {
+    let contents = match std::fs::read_to_string(path) {
+        Ok(c) => c,
+        Err(e) if e.kind() == std::io::ErrorKind::NotFound => {
+            return Ok(LoadedIdentities::default());
+        }
+        Err(e) => {
+            return Err(Error::Config(format!(
+                "failed to read identities file {}: {e}",
+                path.display()
+            )));
+        }
+    };
+
+    let entries: Vec<IdentityFileEntry> = serde_json::from_str(&contents)
+        .map_err(|e| Error::Config(format!("malformed identities file {}: {e}", path.display())))?;
+
+    let mut references = Vec::with_capacity(entries.len());
+    let mut store = IdentityStore::new();
+    for entry in entries {
+        if !entry.session.is_empty() {
+            store.insert(entry.label.clone(), SessionMaterial::new(entry.session));
+        }
+        references.push(TestIdentity {
+            label: entry.label,
+            role: entry.role,
+            tenant: entry.tenant,
+        });
+    }
+
+    Ok(LoadedIdentities { references, store })
+}
+
+/// Convenience: load from the default [`identities_file_path`].
+pub fn load_default_identities() -> Result<LoadedIdentities> {
+    load_identities_from_file(&identities_file_path())
+}
+
 /// The store never reveals its secrets through `Debug` - it reports only the
 /// set of known labels (non-sensitive) and a count, so it is safe to include in
 /// a `#[derive(Debug)]` parent (e.g. `ToolContext`).
@@ -226,5 +330,87 @@ mod tests {
         );
         // Labels are non-sensitive and useful for diagnostics.
         assert!(debug.contains("user_a") && debug.contains("admin"));
+    }
+
+    // ---- file loader ----
+
+    fn write_temp(name: &str, contents: &str) -> PathBuf {
+        // Unique-per-name temp path; tests use distinct names so no clash.
+        let path = std::env::temp_dir().join(format!("pick-identities-test-{name}.json"));
+        std::fs::write(&path, contents).unwrap();
+        path
+    }
+
+    #[test]
+    fn missing_file_yields_empty_not_error() {
+        let path = std::env::temp_dir().join("pick-identities-does-not-exist-xyz.json");
+        let _ = std::fs::remove_file(&path);
+        let loaded = load_identities_from_file(&path).expect("missing file is not an error");
+        assert!(loaded.references.is_empty());
+        assert!(loaded.store.is_empty());
+    }
+
+    #[test]
+    fn valid_file_splits_references_from_secrets() {
+        let path = write_temp(
+            "valid",
+            r#"[
+                {"label":"unauth","role":"anonymous"},
+                {"label":"user_a","role":"user","tenant":"t1","session":"Cookie: sid=aaa"},
+                {"label":"admin","role":"privileged","session":"Authorization: Bearer zzz"}
+            ]"#,
+        );
+        let loaded = load_identities_from_file(&path).expect("valid file loads");
+
+        // References carry no secret and include all three identities.
+        assert_eq!(loaded.references.len(), 3);
+        let admin = loaded
+            .references
+            .iter()
+            .find(|r| r.label == "admin")
+            .unwrap();
+        assert_eq!(admin.role, IdentityRole::Privileged);
+
+        // The store holds only the two with session material.
+        assert_eq!(loaded.store.len(), 2);
+        assert_eq!(
+            loaded.store.resolve("user_a").map(SessionMaterial::expose),
+            Some("Cookie: sid=aaa")
+        );
+        std::fs::remove_file(&path).ok();
+    }
+
+    #[test]
+    fn anonymous_entry_has_reference_but_no_store_slot() {
+        let path = write_temp("anon", r#"[{"label":"unauth","role":"anonymous"}]"#);
+        let loaded = load_identities_from_file(&path).expect("loads");
+        assert_eq!(loaded.references.len(), 1);
+        assert!(
+            loaded.store.resolve("unauth").is_none(),
+            "anonymous identity carries no session material"
+        );
+        std::fs::remove_file(&path).ok();
+    }
+
+    #[test]
+    fn malformed_file_fails_loudly() {
+        let path = write_temp("malformed", r#"{ not an array"#);
+        let err = load_identities_from_file(&path).expect_err("malformed file must error");
+        assert!(
+            err.to_string().contains("malformed identities file"),
+            "unexpected error: {err}"
+        );
+        std::fs::remove_file(&path).ok();
+    }
+
+    #[test]
+    fn unknown_role_fails_loudly_not_silently_dropped() {
+        let path = write_temp(
+            "badrole",
+            r#"[{"label":"x","role":"superuser","session":"s"}]"#,
+        );
+        let err = load_identities_from_file(&path).expect_err("unknown role must error");
+        assert!(err.to_string().contains("malformed identities file"));
+        std::fs::remove_file(&path).ok();
     }
 }
--- a/crates/core/src/identity.rs
+++ b/crates/core/src/identity.rs
@@ -182,12 +182,40 @@ impl IdentityStore {
 /// set, otherwise `identities.json` in the connector's config dir. The path is
 /// not required to exist - a missing file simply means no identities.
 pub fn identities_file_path() -> PathBuf {
-    if let Ok(p) = std::env::var(IDENTITIES_FILE_ENV) {
+    let env_override = std::env::var(IDENTITIES_FILE_ENV).ok();
+    let cwd_candidate = PathBuf::from("identities.json");
+    let cwd_present = cwd_candidate.is_file();
+    resolve_identities_path(
+        env_override.as_deref(),
+        cwd_present,
+        cwd_candidate,
+        crate::settings::settings_dir().join("identities.json"),
+    )
+}
+
+/// Pure resolution logic for [`identities_file_path`], split out so the
+/// precedence rules are testable without mutating process-global cwd/env.
+///
+/// Precedence: explicit `PICK_IDENTITIES_FILE` (if non-blank) > a cwd
+/// `identities.json` that exists > the per-user settings-dir path. Preferring
+/// the cwd file when present is what makes the natural "copy the template and
+/// edit it in place" workflow load, and aligns the loaded path with the one the
+/// repo `.gitignore` protects (#317 H3).
+fn resolve_identities_path(
+    env_override: Option<&str>,
+    cwd_present: bool,
+    cwd_path: PathBuf,
+    settings_path: PathBuf,
+) -> PathBuf {
+    if let Some(p) = env_override {
         if !p.trim().is_empty() {
             return PathBuf::from(p);
         }
     }
-    crate::settings::settings_dir().join("identities.json")
+    if cwd_present {
+        return cwd_path;
+    }
+    settings_path
 }
 
 /// Load operator-provided identities from a JSON file.
@@ -222,9 +250,14 @@ pub fn load_identities_from_file(path: &Path) -> Result<LoadedIdentities> {
     let mut references = Vec::with_capacity(entries.len());
     let mut store = IdentityStore::new();
     for entry in entries {
-        if !entry.session.is_empty() {
-            store.insert(entry.label.clone(), SessionMaterial::new(entry.session));
-        }
+        // Every provisioned label gets a store slot - anonymous identities with
+        // an empty `session` get an empty [`SessionMaterial`]. This keeps a
+        // resolvable-but-empty identity distinct from an unknown label: the
+        // former injects nothing and runs (that IS the anonymous identity), the
+        // latter still misses and fails loud. Skipping empty entries here made
+        // the LLM-instructed `identity_ref: "unauth"` unreachable (it resolved
+        // to `None` and errored "not a provisioned test identity").
+        store.insert(entry.label.clone(), SessionMaterial::new(entry.session));
         references.push(TestIdentity {
             label: entry.label,
             role: entry.role,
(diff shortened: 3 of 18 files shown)


cchao-devo (changes requested):
Review: Pick PR #317 — differential-authz identity model, store, orchestrator advertisement + specialist prompts (#162)
  
  Verdict: 🔴 Request changes — one CONFIRMED CRITICAL credential leak. The type-level secret custody is excellent and the production wiring is real (not a #331-style no-op), but injecting an identity
  causes the customer credential to cross the exact Matrix/LLM boundary the feature exists to protect — via the raw, un-redacted ProbeCommand.command field. This is self-defeating and blocks. The fix is
  small and localized.

  Lenses run: the real pr-review-toolkit plugin agents (silent-failure-hunter + code-reviewer) with the empirical mandate, plus my own source verification of every link in the leak chain + running 
  redact/from_exact. Verified in pick_08/pick.

  🔴 CRITICAL — injected credential leaks to the agent via raw ProbeCommand.command (verified by execution)
  
  When a specialist runs execute_command with identity_ref, build_identity_injection splices the raw material.expose() (Authorization: Bearer <customer-secret> / Cookie: …) into the argv. That argv becomes
  full_command → ProbeCommand::from_exact(full_command). I confirmed at source + by running it:
  - from_exact (provenance.rs:35) stores the raw string in .command and redacts only .effective_command.
  - ProbeCommand.command is pub command: String with no skip_serializing — it goes on the wire (its own doc says "May contain secrets; internal use only," but nothing enforces "internal").
  - Serialized wire JSON (ran it): "command":"curl -H Authorization: Bearer sk-SECRETraw https://t/" — secret present — next to the redacted effective_command.
  - ToolResult.provenance is serialized and returned to the agent by the production PickConnector::execute_with_context (pick_connector.rs:391).

  So on every identity-ref call, the customer credential the feature is designed to keep off the LLM/Matrix boundary sits in plaintext in provenance.command crossing that boundary — directly contradicting
  identity.rs:12-24 ("the secret never crosses the Matrix/LLM boundary"), the very reason SessionMaterial is non-Serialize. redact does catch the header shapes (I ran it — effective_command is safe), which
  makes this specifically the un-redacted sibling field, not both.

  Fix (pick one): (a) redact full_command before ProbeCommand::from_exact on identity-injected runs; (b) #[serde(skip_serializing)] on ProbeCommand.command (verify no wire consumer needs it — the round-trip
  test at tools.rs:1311 asserts it's present, so that test encodes the current leak and would need revisiting); or (c) strip probe_commands[].command at the connector before returning. Plus a guard test
  asserting the serialized ToolResult for an injected run contains no substring of the material. Design note: relying on downstream redact regexes is fragile — the injection point should keep the secret out
  by construction, not lean on a best-effort scrubber.

  🟠 HIGH (residual, related) — reflected credential in data.stdout/stderr

  If the target reflects the request header (echo/verbose/error page, or curl -v on stderr), the injected secret lands in result.stdout/stderr → serialized into data → returned to the agent. Only
  raw_response_excerpt is redacted; the full data.stdout is not (and #331's output sanitizer isn't merged/works only on a test connector). Until output-side scrubbing lands, at minimum redact the
  stdout/stderr when identity_ref was injected, or gate this feature behind the output-scrub work.

  🟠 HIGH (residual) — plaintext credential in logs + process table

  The platform layer (command.rs:70,84) logs args/full_cmd at info/debug — now secret-bearing — so any log sink (Quickwit/journald) captures the credential. And -H "Authorization: Bearer <secret>" as argv
  is visible in ps//proc/<pid>/cmdline. The logging predates this PR but this PR makes it secret-carrying; suppress/redact those tracing args and document the process-table exposure (inherent to
  header-via-argv; curl -H @file avoids it).

  What's excellent (so the rework is targeted)
  
  - Type-level custody is airtight (verified): SessionMaterial non-Serialize + redacting Debug/Display, expose() the sole raw accessor called at exactly one non-test site (the auth path), IdentityStore
  redacting Debug, SpecialistContext.identities: Vec<TestIdentity> (references only — structurally can't hold material since it's Serialize and SessionMaterial isn't), ToolContext not Serialize. The
  metadata advertisement carries no material (test-proven).
  - Production wiring is REAL (unlike #331): load_default_identities() at startup → Arc<IdentityStore> into the production PickConnector → ctx.with_identities() in the live execute_with_context.
  resolve_identity hits a populated store in production.
  - Fail-closed is correct: unknown identity_ref → error (no unauth run); unsupported binary → error (no cred-dropped run); anonymous → inject nothing. binary_basename bypass attempts (curl; evil, CURL,
  /tmp/curl) all fail closed. Flag shapes per binary are right; argv-exec (not shell) means no cross-binary misinjection.
  - File loader fails loud on malformed/unknown-role, missing = no-op. Gates green (fmt/clippy clean, identity 19/0, execute_command 26/0).

  init-dev relevance
  
  This is the production PickConnector path init-dev deploys — so the leak would be live in our newdev environment once an operator provisions identities.json. But the feature is inert until the cross-repo
  keystone (matrix#3354, orchestrator forwarding) lands, so there's runway to fix the leak first.

jtomek-strike48:
Pushed b0b4768 for the CRITICAL. Confirmed your leak chain at source before fixing — traced identity_ref → build_identity_injection (.expose()) → format_full_command → ProbeCommand::from_exact → raw string in .command → serialized in provenance → returned by production PickConnector::execute_with_context. Real, and self-defeating for the feature. Thank you.

**CRITICAL — fixed by construction.** Marked `ProbeCommand.command` `#[serde(skip_serializing, default)]` (`crates/core/src/provenance.rs`). The raw command stays in memory for local traceability; only the redacted `effective_command` crosses any wire/agent/report boundary. Chose option (b) over the redact-at-injection option (a) deliberately, per your design note: this keeps the secret off the wire for **every** tool that ever puts a credential in argv (identity runs, `curl -u user:pass`, …), not just the identity path, and doesn't lean on the redact regex. Verified no production consumer reads `.command` off the wire — report rendering uses `effective_command` (chat_panel/constants.rs:710); only tests referenced `.command`.

- New guard `probe_command_raw_command_never_serializes`: a secret-bearing command must not appear in serialized JSON. Mutation-verified per Lens 1 — removing `skip_serializing` turns it RED, green on restore.
- Your option-b note was right that the round-trip test encoded the leak. Updated three tests that asserted full-struct equality / `.command` on the wire (`tool_result_with_provenance_round_trips_through_json`, evidence `provenance_attaches_cleanly_and_round_trips`, orchestrator `seed_message_round_trips_through_json`) to assert the wire contract instead: raw command absent, redacted `effective_command` intact.

Validation: `cargo test -p pentest-core --lib` 291/0; identity 19/0; execute_command 29/0; `cargo clippy --workspace --features pentest-platform/desktop-pcap --all-targets -- -D warnings` clean.

**The two HIGHs — acknowledged, deliberately deferred (not fixed here), because both are cross-PR:**

- **Reflected credential in `data.stdout/stderr`:** this is exactly what #331 fixes — I moved that PR's sanitizer onto the production `PickConnector::execute_with_context`, so once #331 lands, target-reflected secrets in `data` are scrubbed at the same choke point. Duplicating it here would drift from #331's single-choke-point design and make this PR depend on an unmerged one. Recommend sequencing #331 ahead of / alongside this, or gating the identity feature on #331 (it's inert until the matrix#3354 keystone anyway, so there's runway).
- **Plaintext credential in logs + process table:** the `command.rs:70,84` arg logging predates this PR but this PR makes it secret-bearing, and header-via-argv is inherently visible in `/proc/<pid>/cmdline`. Both warrant their own change (redact the tracing args; consider `curl -H @file` indirection). Happy to file a tracked follow-up issue and take it next rather than expand this PR's scope.

Want me to file the logging/process-table follow-up, or fold the log-redaction in here?

jtomek-strike48:
Filed the logging + process-table HIGHs as #335 (feature/security, P2, S) so they don't get lost — the CRITICAL fix in b0b4768 stands on its own; #335 covers the residual log/proc-table surface.

jtomek-strike48:
Thanks - the CONFIRMED CRITICAL leak-chain trace (raw `ProbeCommand.command` on the wire) was exactly right. Status of all three findings, verified by running:

**CRITICAL - credential on the wire via `provenance.command`. FIXED (commit b0b4768).** Marked `ProbeCommand.command` `#[serde(skip_serializing, default)]` - your option (b), keeping the secret off every wire/agent/report boundary by construction. Verified:
- 291/291 core lib tests GREEN, including the new guard `probe_command_raw_command_never_serializes` and the three round-trip tests updated to assert the wire-facing contract (raw command absent, redacted `effective_command` intact).
- Mutation-confirmed: removing `skip_serializing` turns the guard RED, serializing `"command":"curl -H \"Authorization: Bearer sk-...\""` next to the redacted `effective_command` - the exact leak. Restored -> GREEN.
- No production consumer reads `.command` off the wire (report rendering uses `effective_command`); `default` lets wire-deserialized values round-trip to "".

**HIGH (logs + process table). FIXED (commit 228580e, this push).** The desktop platform layer logged raw `args` (debug, host path) and `args`/`full_cmd` (info, sandbox path) at `command.rs`, so an identity-run credential hit any log sink in plaintext. Both sites now redact via the single-source `pentest_core::provenance::redact` before logging (`cmd` is just the binary name; the sandbox-manager/`set_use_sandbox` lines carry no args, left as-is). Added a regression test asserting `redact` strips the credential from both log-input shapes. The residual process-table exposure (`ps` / `/proc/<pid>/cmdline`, inherent to header-via-argv; `curl -H @file` avoids it) is documented inline and tracked for the differential-authz rollout.

**HIGH (reflected credential in `data.stdout/stderr`). Tracked, not in this PR.** If the target reflects the injected header, it lands in stdout/stderr -> `data` -> agent. This is a separate surface already tracked by **#320** and fixed by the output sanitizer in **#331** (which I've separately verified runs at the production `execute_with_context` choke point and recurses `data` leaves through `redact`, so a reflected `Authorization:`/`Cookie:` is scrubbed there). Deferring here keeps the scrub in one place rather than forking a second redactor into execute_command.

**Runway note (your own point):** the feature is inert until the cross-repo keystone matrix#3354 (orchestrator forwarding) lands, so none of these can leak in production before #320/#331 merge. Gates green locally (fmt clean; the one clippy `const_is_empty` warning is pre-existing in `proot.rs:377`, untouched by this PR, and only on rustc 1.91).

Re-requesting review.

jtomek-strike48:
Thanks for the empirical review, @cchao-devo. All three findings are addressed. Pushed as of head `4a5405f` (this also merged `origin/main` into the branch, which is what brings the #331 sanitizer in -- see the HIGH-2 note).

## CRITICAL -- raw credential in `ProbeCommand.command` crosses the wire

Fixed by construction in `b0b4768`: `#[serde(skip_serializing, default)]` on `ProbeCommand.command`. The raw command stays in memory for local traceability; only the redacted `effective_command` crosses any wire/agent/report boundary. New guard `probe_command_raw_command_never_serializes` asserts a secret-bearing command never appears in the serialized JSON -- mutation-verified (removing the attribute turns it red). Three round-trip tests that encoded the old leaky behavior were updated to assert the wire-facing contract instead.

## HIGH (logs + process table)

Log half fixed in `228580e`: both secret-bearing sites in `crates/platform/src/desktop/command.rs` now route `args`/`full_cmd` through `pentest_core::provenance::redact` before the `tracing::info!`/`debug!`, with a regression test. The residual process-table exposure (`-H "Authorization: Bearer <secret>"` visible in `ps` / `/proc/<pid>/cmdline`) is inherent to header-via-argv and is documented inline; it's tracked in #335 (P2), whose durable fix is `curl -H @file` indirection. #335's log half is now closed by `228580e`, so it narrows to the process-table item on merge.

## HIGH (residual) -- reflected credential in `data.stdout`/`stderr`

This one is worth being precise about, because your review predates the fix landing. The remedy is #320/#331 (`sanitize_agent_result`). At review time #331 wasn't merged and you flagged that it looked test-connector-only. It merged (`5c54b8c`) and this branch now carries it after the `origin/main` merge. Two things resolve the concern:

1. It's on the production path, not a test connector. `sanitize_agent_result` is called in `PickConnector::execute_with_context` (`crates/ui/src/liveview_connector/pick_connector.rs:404`) -- the connector the SDK runner actually drives -- and recursively scrubs secret shapes out of `data` (so `data.stdout`/`stderr`) via the same `provenance::redact` set, before the result reaches the agent.

2. New empirical guard in `4a5405f`: `execute_with_context_scrubs_reflected_credential_from_stdout` drives the real production path with a target reflecting `Authorization: Bearer <secret>` into stdout and asserts the secret survives nowhere in the agent-facing result. Mutation-verified: neutering the `sanitize_agent_result` call turns it red; the benign-passthrough test stays green.

Coverage note (so it can't silently vanish): the `liveview_connector` tests compile only under the `connector` feature (which implies `liveview`). CI's `cargo test --workspace --locked --features pentest-platform/desktop-pcap` pulls it in via feature unification; to run locally, `cargo test -p pentest-ui --features desktop,connector`.

Design note on your "keep the secret out by construction" point: agreed as the durable direction. The wire/agent boundary is now closed by construction (skip-serialize) plus defense-in-depth (the sanitizer); the one remaining argv-broadcast surface is the process table on the operator's own host, tracked in #335 with the `-H @file` fix. Feature is still inert until matrix#3354 lands, so there's runway.

Re-requesting review.

ngdevo:
HIGH

  H1. Anonymous identities are functionally unreachable via execute_command.
  crates/tools/src/execute_command.rs:build_identity_injection calls ctx.resolve_identity(label).ok_or_else(...)? before the empty-material fast-path. But the file loader
  (crates/core/src/identity.rs:277) only inserts into the store if !entry.session.is_empty() — anonymous identities are added to references but not to IdentityStore. Result: the LLM does
  identity_ref: "unauth" (exactly what both specialist prompts instruct), resolve_identity returns None, and the tool errors with "identity_ref 'unauth' is not a provisioned test identity; refusing
  to run unauthenticated". The "anonymous injects nothing" unit test only passes because ctx_with("anon", "") inserts a placeholder into the store — the production loader never does that. Fix:
  either insert anonymous entries into the store (empty material) or teach build_identity_injection to consult the reference list before failing.

  H2. Wire-contract break for ProbeCommand.command is unversioned.
  crates/core/src/provenance.rs now marks command as #[serde(skip_serializing, default)]. This is the right hardening for injected credentials, but every downstream consumer of an evidence/manifest
  JSON (report agent, exports, any historical evidence file replayed) will see command == "". The evidence and orchestrator tests are updated in-place — good — but there is no schema-version bump,
  no migration note, and no back-compat check for evidence files persisted before the deploy. If evidence files are cached to disk or replayed cross-version this will surface as a silent regression.

  H3. .gitignore protects the wrong path.
  .gitignore (line 12) ignores identities.json at repo root; the committed template identities.example.json is placed at repo root. But load_default_identities() reads from
  settings::settings_dir().join("identities.json") (i.e. ~/.config/pentest-connector/identities.json). A user who follows the natural "copy the example, drop it next to the example" workflow puts
  their file at the repo root — where the loader will not find it. Only PICK_IDENTITIES_FILE env or a manual copy into settings_dir() works. Either move identities.example.json into the settings
  directory template, or teach identities_file_path() to prefer a repo-root file when present. As-is, the gitignore rule protects a path the code doesn't actually load — which is a false sense of
  security if someone drops a real file in the settings dir with a different name.

  MEDIUM

  M1. Header-via-argv exposes secret in /proc/<pid>/cmdline.
  crates/platform/src/desktop/command.rs acknowledges this in a comment. curl -H "Authorization: Bearer <token>" — the token is visible to any local user for the child's lifetime. The PR's tracking
  note is honest, but this is the exact class of leak the differential-authz feature exists to close. curl -H @file (or piping stdin) would remove the exposure; the current design does not. Not a
  blocker if a follow-up is scheduled, but should not ship to customers unmarked.

  M2. IdentityStore::insert silently replaces on duplicate label.
  crates/core/src/identity.rs:207-209. An operator who accidentally repeats a label in identities.json gets exactly one identity of that label, with no warning. The unit test asserts this as
  intended behavior, but the loader (line 274) iterates entries without a dedup check. For a file where labels drive the entire authorization matrix, a duplicate is almost certainly a mistake.
  Consider erroring on duplicate labels the same way the loader errors on unknown roles.

  M3. Loader path is not validated for symlink/permissions.
  load_identities_from_file reads whatever $PICK_IDENTITIES_FILE points at. No check that the file is not world-readable, not a symlink out of a trusted directory, not on a network mount. The file
  is documented as holding real customer credentials.

  LOW

  L1. IdentityFileEntry treats missing session and "session": "" identically; a "session": null errors loudly. This is inconsistent; matches what the tests exercise, but null is the more idiomatic
  "no material" in JSON.

  L2. references_metadata_value returns None on serde_json::to_string failure ("infallible for these fields"). Unreachable in practice, but a debug_assert! would catch a future field change.

  L3. build_identity_injection prepends injection before user args. If a user manually passes -H "Authorization: <different>", curl accepts both headers — but for single-valued flags (-A, -b) the
  last-wins rule means the user's value overrides the identity. Minor precedence surprise.

  L4. The header-flag whitelist (curl | ffuf | wget | sqlmap) is closed. Every future tool that grows an identity-injection need requires editing this map. Not a bug, but a real growth pressure.

jtomek-strike48:
Thanks @ngdevo — all three HIGH findings confirmed at source and fixed in `1793dcf`. Verified empirically, not by reasoning.

**H1 (anonymous unreachable) — confirmed, was a genuine functional break.** You were exactly right, and the passing test was lying. Traced it at source:
- Loader (`identity.rs:225`): `if !entry.session.is_empty()` pushed anonymous to `references` but never to `store`.
- Injection (`execute_command.rs:288`): `resolve_identity(label).ok_or_else(...)?` errors *before* the `material.is_empty()` fast-path — so `identity_ref: "unauth"` (what both specialist prompts instruct) resolved to `None` and hit "not a provisioned test identity."
- The green `build_injection_empty_material_injects_nothing` only passed because `ctx_with("anon", "")` inserts a placeholder the production loader never creates.

Fix: every provisioned label now gets a store slot (empty material for anonymous), so anonymous resolves to `Some(empty)` → existing fast-path injects nothing and runs; an unknown label still misses → fails loud. Added an **end-to-end guard through the real loader** (`build_injection_anonymous_from_real_loader_injects_nothing`) — not a hand-built store. **Mutation-verified:** re-adding the skip-empty guard turns both the loader test and the e2e test red (the e2e one with the literal `identity_ref 'unauth' is not a provisioned test identity` error you predicted). Restored → green.

**H3 (.gitignore guards the wrong path) — confirmed.** `identities_file_path()` read `settings_dir()/identities.json`; the gitignore rule + `identities.example.json` both live at repo root. Fix: the resolver now prefers a cwd `identities.json` when present (env override still wins), so the "copy the template, edit in place" workflow both loads and stays ignored. Split precedence into a pure `resolve_identities_path` so it's tested without mutating process-global cwd/env (4 precedence tests).

**H2 (unversioned wire-contract break) — verified lighter than feared; no bump needed.** `ProbeCommand` already carries a serialized redacted `effective_command` (the report renders *that* — `constants.rs:710`), and grepping every prod reader of raw `.command` back from JSON turns up only comments — nothing consumes it. `skip_serializing` doesn't affect *reads*, so old-format evidence files (with a `command` key) still deserialize cleanly via `#[serde(default)]`. Added `probe_command_deserializes_legacy_wire_with_command_present` covering both legacy and current wire shapes, plus a back-compat note on the field doc.

**Verification:** `cargo test -p pentest-core` 330/0, `-p pentest-tools` 354/0; `cargo fmt --check` clean; `cargo clippy -p pentest-core -p pentest-tools --all-targets` zero warnings.

**On your MEDIUMs** (deferred, tracked — not in this commit): M1 (header-via-argv in `/proc/cmdline`) and the reflected-credential-in-stdout residual you flagged in the prior round are the same argv/output-scrub class; agreed they should be gated behind the output-scrub work (#331) or moved to `-H @file` before this ships to customers. The feature stays inert until the orchestrator keystone (matrix#3354) lands, so there's runway. I'll file a follow-up for the argv/stdout exposure rather than inline it here. M2 (duplicate-label silent replace) and M3 (loader path perms/symlink) — happy to fold into that follow-up too if you'd prefer them tracked.

Re-requesting your review.

jtomek-strike48:
@cchao-devo — ready for re-review at `1793dcf`. All findings from your 2026-07-27 review are addressed; CI is fully green (the earlier red `Test` job was the pre-existing `evidence_producer::multiple_push_single_drain` shared-buffer flake — it passed on rerun, and this branch doesn't touch `evidence_producer.rs`).

Verified empirically, not by reasoning:

- **CRITICAL — raw `ProbeCommand.command` on the wire** → fixed by construction in `b0b4768`: `#[serde(skip_serializing, default)]` on `.command`. New guard `probe_command_raw_command_never_serializes` passes, and I **mutation-verified** it — removing the attr turns the guard red with the secret visible in the serialized JSON, so it genuinely guards. Three round-trip tests that encoded the old leaky behavior were updated; a legacy-wire deserialize test covers back-compat.
- **HIGH — reflected credential in `data.stdout/stderr`** → covered by the #320/#331 sanitizer now merged onto this branch, sitting on the production `PickConnector::execute_with_context` path (refuting the "test connector only" concern); guard added in `4a5405f`.
- **HIGH — plaintext credential in logs + process table** → tracing args redacted in `228580e`; the inherent process-table exposure of header-via-argv is tracked as #335.

ngdevo's three follow-up HIGHs (anonymous identity unreachable, gitignore/path mismatch) are fixed in `1793dcf`, mutation-verified end-to-end through the real loader.

Re-requesting your review — thanks for the leak-chain trace, it was exactly right.

ngdevo:
Secret custody in the type system is genuinely good: SessionMaterial is not Serialize, Debug/Display redact, TestIdentity carries no secret by construction, and IdentityStore's Debug shows labels only. The problems are at
  the boundaries where the secret leaves Rust's control, and one blast-radius issue that has nothing to do with identities.

  HIGH

  1. The credential rides in argv and is world-readable from the process table. header_flags_for/2 returns ["-H", "Authorization: Bearer …"] (curl/ffuf) or --header=… (wget/sqlmap), spliced into the child's argv. Any local
  user can read it from ps / /proc/<pid>/cmdline for the child's lifetime. The PR acknowledges this in a code comment in crates/platform/src/desktop/command.rs — "inherent to header-via-argv; curl -H @file avoids it. Tracked
  for the differential-authz feature rollout (gated on matrix#3354)" — and ships anyway. It's inherent to this design, not to the goal: curl and ffuf both read headers from a file, and sqlmap takes -r <requestfile>. Shipping
  operator production credentials this way on a shared or multi-tenant connector host is the single riskiest thing in the PR.

  2. #[serde(skip_serializing, default)] on ProbeCommand.command silently empties the field for every tool, and the agent prompts still tell agents to read it. The change is repo-wide, not identity-scoped: after any
  serialize→deserialize hop, command comes back "". Both pre-existing round-trip tests were amended to assert command == "" (crates/core/src/evidence.rs, crates/core/src/tools.rs) rather than an alternative being explored.
  Meanwhile crates/ui/src/components/chat_panel/constants.rs — untouched by this PR — still documents probe_commands[].command in the JSON schema shown to the Report and Validator agents (lines ~693, ~866) and rule #2 at line
  710 reads: "Render effective_command from each finding's probe_commands[].command field verbatim." That instruction now resolves to an empty string on every finding. Either scope the skip to identity runs, or update the
  prompts and any other .command reader in the same PR.

  3. effective_command redaction is not guaranteed to cover operator-authored session material. SessionMaterial is a free-form header line. redact()'s regexes catch Authorization:, Cookie:, Bearer, long hex/base64, and
  *api[_-]?key\s*[=:]. A perfectly plausible identity like X-Session-Token: s7f2q9 (short, non-hex, non-matching key name) survives redaction — and effective_command is serialized and lands in customer-facing reports. The
  custody guarantee is type-enforced on one side and regex-hoped on the other. Constrain the accepted header names, or mark the injected fragment for unconditional replacement at splice time rather than pattern-matching it
  afterward.

  MEDIUM

  4. The loud-failure contract is defeated at its only production call site. load_identities_from_file's doc states a malformed file "fails loudly rather than silently degrading to zero identities, which would misreport authz
  coverage," and two unit tests assert the Err. In crates/ui/src/liveview_connector/mod.rs the sole caller does Err(e) => { tracing::warn!("Ignoring identities file: {e}"); LoadedIdentities::default() } — producing precisely
  the silent degradation. A typo in the operator's JSON yields a connector that starts clean, advertises no identities, and lets specialists report "reduced coverage" as if that were the intended configuration. Fail startup,
  or surface it to the operator in the UI.

  5. No de-duplication against user-supplied args. Injected flags are prepended, so if the model also passes -H "Cookie: …", curl honours the later occurrence — the user arg silently overrides the identity, and the
  differential result is attributed to the wrong principal. That is the exact silently-wrong failure mode build_identity_injection's doc comment says it exists to prevent. Detect a conflicting header in args and error.

  6. Nothing records which identity produced which response. identity_ref doesn't reach Provenance (and per #2 the raw command that would have shown it is now blank on the wire). A differential-authz finding is only meaningful
  with the identity attached; as written, a reviewer cannot verify that user_a and user_b responses came from different principals.

  7. Secrets are cloned per tool call with no zeroization. ctx = ctx.with_identities((*self.identities).clone()) on every execute_with_context, and SessionMaterial is a plain String with no Drop scrub, so copies accumulate in
  freed heap for the connector's lifetime. Arc the store into the context, or use a zeroizing wrapper.

  LOW

  8. cwd-preferring path resolution is a foot-gun. resolve_identities_path prefers ./identities.json when present. Launch the connector from an unexpected directory and it loads a different (or attacker-planted, in a shared
  workspace) identities file; the .gitignore protection only covers the repo root. Log the resolved path at startup.

  9. references_metadata_value swallows serialize errors with .ok(). Unreachable today, but the failure mode is "identities silently not advertised," which is #4 again in miniature.

  10. Injection supports only curl/ffuf/wget/sqlmap — none of the actual web scanners (nikto, zap, nuclei). Worth stating in the PR as a known scope limit, since the specialist prompts describe differential authz as a general
  capability.

jtomek-strike48:
Thanks @ngdevo — the secret-custody-at-the-boundary framing was exactly right. Three fixed in-PR (`3ed88bb`), verified empirically (ran the modules, mutation-checked each guard); the rest are tracked in pick#353. Two small corrections noted below where the code differs from the finding.

**HIGH #3 (redaction gap) — FIXED.** You're right that pattern redaction can't be the guarantee for arbitrary operator header material. Since the injected secret is *known*, I scrub it by exact substring **before** `redact()` runs (new `ProbeCommand::from_exact_redacting_secret`) — security-by-construction, not regex-hope. One correction: your `X-Session-Token: s7f2q9` example is actually caught (the `env_secret` regex alternation includes `token`). The real surviving shape is something like `-H "X-Api-Id: ab12cd"` (no keyword, short, non-hex) — that's the regression test I added, and it leaks pre-fix / is scrubbed post-fix.

**MED #6 (attribution) — FIXED.** The non-secret identity label now rides on `ProbeCommand.description` ("authenticated as test identity: <label>"), which serializes onto the wire, so a reviewer can attribute a response to its principal even though the raw command (which carried the header) is `skip_serializing`.

**HIGH #2 — the serde change is intentional/correct; the real defect was the stale prompt.** `skip_serializing` on `ProbeCommand.command` is deliberate (keeps the secret off the wire by construction). The open bug was that `chat_panel/constants.rs` rule #2 still told the Report agent to render `effective_command` from `probe_commands[].command` — now always `""`. Repointed it to `probe_commands[].effective_command` and dropped the never-serialized `command` key from the two schema examples. (The "Reproduce it" step and Validator rules already used `effective_command` correctly.)

**Verification:** pentest-core 36/0, pentest-tools 29/0. Each guard mutation-verified — neuter the value-scrub → exotic-shape test RED; drop the attribution wiring → end-to-end test RED; drop the execute-level scrub → RED. `cargo fmt` + `clippy -D warnings` clean; workspace checks with `--features pentest-platform/desktop-pcap` (CI parity).

**Tracked as follow-up in pick#353** (feature is inert until matrix#3354): #4 loud-failure-defeated-at-caller, #5 header dedup vs user args, #7 zeroize/Arc-the-store. One correction on #5 — `build_identity_injection`'s doc doesn't actually claim to dedup, but the gap (curl sends both headers) is real and worth closing. #1 (argv/process-table) stays in #335. #9/#10 are no-action (documented-unreachable / already stated as a known limit with a loud error).

Re-requesting review from @ngdevo and @cchao-devo.

ngdevo:
Verified at source:

  - provenance.rs:56-79 — from_exact_redacting_secret scrubs the known secret by exact substring before redact() runs. This is the right inversion: security-by-construction instead of pattern-matching after the fact. The
  empty-secret path correctly falls through to pattern-only rather than replace("") blanking the string.
  - execute_command.rs:249-275 — IdentityInjection carries injected_secret with a manual Debug that prints has_secret: bool, never the value. A derived Debug would have defeated the whole custody model in logs and panic
  output; getting this right matters.
  - execute_command.rs:163-175 — attribution flows through ProbeCommand.description, which does serialize (verified by the round-trip test asserting the label on the wire and the secret absent from it).
  - chat_panel/constants.rs:710 — rule #2 now reads probe_commands[].effective_command; both schema examples (:694, :867) dropped the never-serialized command key. The stale-prompt bug is genuinely closed.
  - Follow-up tracking is real, not hand-waved: pick#353 is open and its §4/§5/§6 describe findings #4/#5/#7 accurately, with file/line and a proposed fix each. pick#335 is open for the argv exposure.

  Two developer corrections I accept. My X-Session-Token: s7f2q9 example is caught — the env_secret regex alternation includes token. The surviving shape is X-Api-Id: ab12cd (no keyword, short, non-hex), which is exactly the
  regression test added. And build_identity_injection's doc comment does not claim to dedup; my characterization overstated it. The gap is still real, just not a broken promise.

  Remaining issues

  HIGH — finding #1 (argv exposure) ships unfixed. The credential is readable from ps / /proc/<pid>/cmdline for the child's lifetime. It's tracked in pick#335 and acknowledged in a code comment, and the feature is inert until
  matrix#3354 lands — but this is the exact leak class the differential-authz feature exists to close, and curl -H @file / ffuf -H from file / sqlmap -r all avoid it. It is inherent to this design, not to the goal.

  MEDIUM — finding #4 unfixed (liveview_connector/mod.rs:751-752 still warn! + LoadedIdentities::default()). Mitigating: an empty store makes a later identity_ref fail loud per-tool at first use, so a typo'd file does surface
  — just later and less clearly than the loader's own doc promises.

  Verdict: Approve, conditional on one answer

  The three in-PR fixes are correct and well-tested, the custody model is genuinely strong (SessionMaterial not Serialize, redacting Debug/Display, manual Debug on the new injection type), CI is fully green, and the deferrals
  are tracked with real issues rather than promises.

  The one thing I want confirmed before approving: is the feature actually inert until matrix#3354? If yes, approve — finding #1 is a rollout gate, not a merge gate, and pick#335 is the right place for it. If any path can
  reach build_identity_injection on a customer connector before matrix#3354, then #1 blocks: don't put operator production credentials in argv on a shared host.

  I'd also ask for a one-line startup log of the resolved identities path (finding #8) — trivial, and it makes the cwd-preferring resolution auditable.

ngdevo (approved):
Verdict: Approve, conditional on one answer

  The three in-PR fixes are correct and well-tested, the custody model is genuinely strong (SessionMaterial not Serialize, redacting Debug/Display, manual Debug on the new injection type), CI is fully green, and the deferrals
  are tracked with real issues rather than promises.

  The one thing I want confirmed before approving: is the feature actually inert until matrix#3354? If yes, approve — finding #1 is a rollout gate, not a merge gate, and pick#335 is the right place for it. If any path can
  reach build_identity_injection on a customer connector before matrix#3354, then #1 blocks: don't put operator production credentials in argv on a shared host.

  I'd also ask for a one-line startup log of the resolved identities path (finding #8) — trivial, and it makes the cwd-preferring resolution auditable.

jtomek-strike48:
@ngdevo — thanks for the re-review and the conditional approve. Answering your gating question directly.

**Is the feature inert until matrix#3354? Yes — verified at source, not by reasoning.**

The inertness doesn't rely on the LLM refraining from emitting `identity_ref` (it could). It's enforced by the store failing closed: `build_identity_injection` (`crates/tools/src/execute_command.rs:346-351`) resolves the label via `ctx.resolve_identity(label)`, and on an **unprovisioned connector that returns `None`** → hard error `"is not a provisioned test identity … refusing to run unauthenticated"`. The call site (`:128`) maps that to `return Ok(ToolResult::error(...))`, so the tool aborts **before** any secret is spliced into argv.

`resolve_identity` only returns material when an operator has placed `identities.json` on the connector. Until matrix#3354 forwards identities *and* an operator provisions the file, there is no path that reaches argv with a credential — finding #1's `ps`/`/proc/<pid>/cmdline` exposure cannot occur because the argv is never constructed with a secret. So #1 is a **rollout gate** (don't provision operator prod creds on a shared host until #335 moves the header off argv), correctly tracked in pick#335, not a merge gate. Finding #4 (pick#353) stays a tracked MEDIUM as you noted.

**Finding #8 (startup log of the resolved identities path):** agreed it's trivial and makes the cwd-preferring resolution auditable. Happy to add a one-line `info!` at loader startup — say the word and I'll push it here, or I'll fold it into the pick#353 loader-hardening work, whichever you prefer for keeping this PR's diff tight.

Re-requesting @cchao-devo to clear the stale 2026-07-27 change-request — all three of their original findings are fixed (CRITICAL by-construction in `b0b4768`, the two HIGHs in `4a5405f`/`228580e` + tracked in #335) and CI is fully green.

jtomek-strike48:
Thanks for the thorough review — all three findings are addressed on the current head (`c6163b1`). Verified by running the guard tests, not by reasoning.

**CRITICAL — raw `ProbeCommand.command` leak (b0b4768)**
`command` is now `#[serde(skip_serializing, default)]`, so it is retained in-memory for local traceability but never crosses the wire. Added `from_exact_redacting_secret()` so an injected identity is scrubbed by value at construction (security-by-construction, not a downstream regex). Guard test `provenance::tests::probe_command_raw_command_never_serializes` asserts the serialized `ToolResult` contains no substring of the secret and no `command` key — removing the `skip_serializing` turns it red. A back-compat test confirms legacy wire JSON with a `command` key still deserializes.

**HIGH — reflected credential in `data.stdout/stderr`**
The #320/#331 output sanitizer has since merged to main (pulled in here via `a0c5576`) and now runs on the production `PickConnector::execute_with_context`, not a test connector. `4a5405f` adds a guard on that production path.

**HIGH — plaintext credential in logs + process table (228580e)**
Both arg-bearing log sites in `desktop/command.rs` (debug host path, info sandbox path) now redact via the single-source `provenance::redact` before logging; the other `info!` sites log only the binary name. The process-table exposure (argv visible in `ps` / `/proc/<pid>/cmdline`) is documented inline as inherent to header-via-argv, with `curl -H @file` noted as the avoidance. Guard test asserts `redact` strips the credential from the exact `format!` shapes the log statements build.

Gates green (fmt/clippy clean; identity + provenance + specialist_spawner suites pass). Re-requesting review.

ngdevo:
Minor review comments. Approving the PR. @cchao-devo 's blocking comments are also addressed.

Result: All 19 checks green on c6163b1 (2 skipped by design).

 Residual findings (new — not blockers)

 MEDIUM — the by-value secret scrub covers only the report boundary; logs and agent output are still regex-only

 The PR's own test (execute_command.rs:740) asserts that X-Api-Id: ab12cd survives provenance::redact —
 that precondition is why from_exact_redacting_secret exists. But the injected secret is known only at the
 ProbeCommand construction site. The other two boundaries fall back to pattern matching:

 - Logs: crates/platform/src/desktop/command.rs:74,98 redact via provenance::redact only → such material
 lands verbatim in journald/Quickwit on every identity run.
 - Agent output: crates/core/src/sanitize.rs (sanitize_agent_result, the #331 choke point) also reuses
 redact → a target that reflects the header puts the same material in data.stdout in front of the LLM.

 This is the same finding class as the HIGH #3 that was fixed, on two surfaces instead of one, and it is currently
 untracked: pick#335 was narrowed to the process-table item after 228580e closed "the log half".

 Cheap fix, in the file that already holds the secret: applied_identity is in scope inside the timed closure in
 execute_command.rs, so result.stdout/stderr can be exact-substring-scrubbed there before data is built
 (same construction as from_exact_redacting_secret). The log surface needs the secret threaded to the platform
 layer, or the log line reduced to the redacted effective_command form — better as a #335 item.

 LOW — the exact-substring scrub degrades silently when format_full_command escapes the material

 format_full_command (execute_command.rs:279-293) rewrites " as \" inside quoted args, so session material
 containing a double quote (e.g. Authorization: Digest username="admin", …) no longer matches the raw secret
 in command.replace(secret, REDACTION) — the scrub silently no-ops and only the regex pass remains. Harmless for
 shapes redact catches, a leak for shapes it does not. Worth a debug_assert!(!effective_command.contains(secret))
 in from_exact_redacting_secret so a future no-op is loud rather than silent.

 LOW (process, not code) — operator-facing rollout warning has no home

 The ps exposure is documented in a code comment; the operator-facing artifacts (identities.example.json,
 the .gitignore block) say "never commit" but not "provisioning real credentials puts them in argv, readable by
 any local user, until pick#335 lands". One line in the example file or the gitignore block would close it.

ngdevo (approved):
LGTM. See last comment.

jtomek-strike48:
Pushed `50d04e3` addressing @ngdevo's 2026-08-14 residual findings. Verified empirically (ran the guards + mutation-checked), not by reasoning.

## MEDIUM — by-value scrub only covered the report boundary; logs + agent output were regex-only

Agreed, and it's the same class as the HIGH #3 that was fixed, on another surface. Closed the **agent-output** surface by construction:

- Extracted `build_run_output()` from the run closure (`crates/tools/src/execute_command.rs`) so the result-building path is testable without a live subprocess, and scrub the **known** injected secret by value from `stdout`/`stderr` **before** `data` and the provenance excerpt are built. So a target that reflects the header (`curl -v`, verbose echo, error page) can't put the credential in front of the agent regardless of shape — no reliance on the regex pass.
- This also value-scrubs the provenance **excerpt** (previously regex-only, as cchao noted in the original review).
- Factored the scrub into shared `provenance::redact_known_secret()`, reused by `from_exact_redacting_secret` so the marker + semantics have one source.

New guard `run_output_scrubs_reflected_identity_secret_from_output` drives the production path with `X-Api-Id: ab12cd` (the exact shape the PR's own `execute_command.rs` test proves `redact()` misses) reflected into both stdout and stderr, and asserts the secret survives nowhere in `data` or the excerpt. **Mutation-verified**: removing the scrub turns it RED with the literal leak (`reflected secret leaked into data.stdout: ... X-Api-Id: ab12cd`); restored → GREEN. Plus passthrough + anonymous-identity guards so the scrub can't over-reach or corrupt on empty material.

**The log surface stays in #335, deliberately.** The `command.rs` sites already regex-`redact()` (landed in `228580e`), but closing the exotic-shape gap there by value needs the secret **threaded to the platform layer** — a bigger, separate change that belongs with the process-table work in #335 (updated there). Not folding it in keeps this diff tight.

## LOW — scrub degrades silently when `format_full_command` escapes the material

Added the `debug_assert!` you suggested in `from_exact_redacting_secret`: trips loudly in debug/test if the by-value scrub no-ops and the secret survives verbatim into `effective_command`. Honest scope: it catches the plain no-op; it does **not** catch the escaped-quote variant (raw substring absent, escaped form present) — documented inline as the known LOW edge, with the durable fix being scrub-before-escape.

## LOW — operator rollout warning had no home

Added the argv / `/proc/<pid>/cmdline` exposure warning (don't provision real creds on a shared host until #335 moves headers off argv) to the identities `.gitignore` block. (Couldn't live in `identities.example.json` — strict JSON array; the loader rejects a non-entry comment object.)

Gates green locally: `fmt --check` clean; `clippy -p pentest-core -p pentest-tools --all-targets --features pentest-platform/desktop-pcap -- -D warnings` clean; `pentest-core` 334/0, `pentest-tools` 359/0. All four new tests run in the required Linux lane (no `#[cfg]` gating).

@cchao-devo — this also hardens the exact boundary from your 2026-07-27 review. All three of your original findings remain addressed (CRITICAL by-construction in `b0b4768`; logs `228580e` + #335; reflected-output now closed by value here). Could you clear the stale change-request when you get a chance? CI is fully green.

cchao-devo (changes requested):
Review of pick PR #317 (@ 50d04e3) — feat(#162): differential-authz identity model, store, orchestrator advertisement + specialist prompts

  Verdict: REQUEST CHANGES — excellent design and test discipline, but one CRITICAL authz-integrity defect and two HIGH silent-degradations must be fixed first.

  Reviewed with the full panel (code-review, silent-failure, test-coverage, type-design) plus local gates; I independently verified every blocker against the code. The central security invariant
  holds and is impressive: SessionMaterial is compile-enforced to never cross the Matrix/LLM boundary (not Serialize, redacting Debug/Display), the only .expose() call routes to argv + by-value
  scrubbing, TestIdentity/IdentityRole are provably the only identity data that can enter SpecialistContext, and provenance redaction has genuine negative-control tests. Gates green: fmt +
  clippy --workspace clean; tests pass (identity 24, provenance 37, execute_command 32) and run on the ubuntu CI lane.

  The blockers are all in the degraded paths — where an identity ends up unauthenticated but is still reported under its privileged label, which for a pentest tool is a false finding, the worst
  failure mode.
  1. CRITICAL — a non-anonymous identity with empty session material runs unauthenticated and is reported as that identity. IdentityStore is keyed label → SessionMaterial and drops the role
  entirely (identity.rs); the loader (identity.rs:252-266) inserts every entry with no check that role != Anonymous ⟹ non-empty session; and the injection gate keys off material.is_empty() alone
  (execute_command.rs:394-399), which cannot tell an anonymous identity from a mis-provisioned privileged one. So {"label":"admin","role":"privileged"} with an omitted session → resolve returns
  Some(empty) → inject nothing → the command runs with no auth, yet applied_identity attributes the response to "admin" → the differential-authz result reads "admin-only endpoint reachable as
  admin" when the request carried no credential. This is exactly the mis-auth-as-valid-authz class the module exists to prevent, it slips past because resolve returned Some (not None), and no
  test exercises it (every privileged test supplies a session). Fix: fail loud at the loader (role != Anonymous && session.is_empty() → Error::Config); defense-in-depth, carry the role into the
  store and take the empty-inject fast-path only when role == Anonymous.
  2. HIGH — a non-string identity_ref is silently dropped and the command runs unauthenticated. execute_command.rs:113 uses .and_then(|v| v.as_str()), so identity_ref: 123 / {...} → None →
  injection skipped → unauth run reported as success. Absent/null (legit "no identity") is indistinguishable from a malformed ref. Given the fail-closed ethos (LLM tool-calls violate schemas),
  distinguish None|Null (run as-is) from String (inject) from anything else (fail loud).
  3. HIGH — a malformed identities file is swallowed, silently evaporating the authz matrix. load_identities_from_file deliberately returns Err on bad JSON / unknown role, documented
  (identity.rs:224-226) as failing loud "rather than silently degrading to zero identities, which would misreport authz coverage." But the sole caller (liveview_connector/mod.rs:766-769)
  converts that into warn! + LoadedIdentities::default() (zero identities). An operator typo → connector boots with no identities → specialists still run differential-authz prompts → "no broken
  access control" reported when nothing was tested. Untested arm. Fix: propagate / refuse to start (or surface a blocking, operator-visible health error) and reconcile the two layers' contracts.

  Should-fix (Medium)

  4. A "-containing exotic-shape secret can leak into the published report. format_full_command escapes embedded " (execute_command.rs:323-326) while from_exact_redacting_secret scrubs the raw
  secret (provenance.rs:65) — a mismatch, so the by-value scrub no-ops and pattern-only redact() misses exotic shapes → verbatim leak into effective_command. The debug_assert! (provenance.rs:77)
  anticipates this but is release-compiled-out and every test secret is quote-free. Fix: scrub against the escaped form too (or redact before escaping) and add a quote-containing-secret test to
  pin release behavior.
  5. Injected credential can leak to logs for non-standard header shapes. At the platform log layer (command.rs:98-103) only pattern-redact() is available (no known value there), so an exotic
  X-Api-Id: ab12cd-style secret hits journald/Quickwit in plaintext; the guard test only covers Authorization: Bearer. Bounded (standard shapes caught) and adjacent to the accepted pick#335
  process-table residual. Fix: thread injected_secret down for a value-scrub, or log only the binary name, or explicitly extend the documented residual.

  Low / design nits

  - references_metadata_value uses .ok() (identity.rs) — infallible today; prefer .expect("TestIdentity serialization is infallible") so a future regression panics loudly rather than silently
  dropping advertisement.
  - Duplicate label in the file: references.len() diverges from store.len() (last-wins) → ambiguous matrix advertised. Add a loader dedupe/reject + test.
  - Type design: after .expose() the secret becomes a plain String in AppliedIdentity.secret, guarded only by hand-written/absent Debug — fragile; consider keeping SessionMaterial until argv.
  LoadedIdentities' public fields let references/store disagree — prefer a constructor / non-pub fields. Drop the unused Eq on SessionMaterial (non-constant-time equality on secret bytes).

  Verified good

  Central invariant compile-enforced (non-Serialize secret; references-only context); resolve() None fails loud before any subprocess spawns; provenance by-value scrub with regex-defeating
  negative controls; ProbeCommand.command #[serde(skip_serializing)]; IdentityRole total; fmt/clippy --workspace clean; new tests run on the ubuntu CI lane.

jtomek-strike48:
Pushed a71dbb7 addressing the CRITICAL + both HIGH findings. Verified locally with 1.92.0: `cargo test -p pentest-core -p pentest-tools` (identity + execute_command suites green, incl. new guards), `cargo fmt --check` clean, and `cargo clippy --workspace --locked --features pentest-platform/desktop-pcap -- -D warnings` clean.

1. CRITICAL (privileged-empty runs unauth, reported as privileged): two layers.
   - Loader: `load_identities_from_file` now rejects `role != Anonymous && session.is_empty()` with `Error::Config`. Guard: `privileged_identity_without_session_fails_loudly`, `user_identity_with_empty_session_string_fails_loudly`.
   - Defense-in-depth: `IdentityStore` now stores the role (`StoredIdentity`), and the injection gate takes the empty-inject fast-path only when `role == Anonymous`; any other role with empty material fails loud. Guard: `build_injection_nonanon_empty_material_fails_loud` (green with fix; the old code returned Ok(inject-nothing), so it genuinely guards).

2. HIGH (non-string identity_ref silently dropped): `identity_ref` is now discriminated - absent/null = run as-is, string = inject, any other shape = fail loud. Guards: `execute_non_string_identity_ref_fails_without_running`, `execute_null_identity_ref_runs_as_is`.

3. HIGH (malformed identities file swallowed to zero identities): the connector now returns `Err` from `connect_and_run` on a malformed/unreadable file (refuse to start); a missing file remains a non-error.

The Medium/Low items (quote-containing-secret release-scrub #4, exotic log-shape #5, `references_metadata` `.expect`, duplicate-label reject, `LoadedIdentities` non-pub fields) are not yet addressed - happy to fold them into this PR or track separately, your call.

ngdevo:
Verdict: ✅ Code is merge-ready — but don't click merge yet (one process blocker)

  Every merge-blocking finding across the PR's entire review history is now resolved and test-guarded. I verified the final commit's diff directly rather than trusting the author's summary. The only thing standing between you and merge is a stale review state, not the code.

  The actual blocker

  cchao-devo's CHANGES_REQUESTED is still active and was filed against 50d04e3 — one commit before the fix a71dbb7. cchao has not re-reviewed. Both of your (ngdevo) approvals also predate the current head. Under normal branch protection, that outstanding change-request blocks the merge until cchao re-reviews or it's dismissed.

  cchao's 3 blockers — all fixed in a71dbb7 (verified in the diff, each mutation-guarded)

  ┌────────────┬─────────────────────────────────────────────────────────────────┬────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
  │     #      │                             Finding                             │                                                                    Fix                                                                     │
  ├────────────┼─────────────────────────────────────────────────────────────────┼────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
  │ 🔴         │ Privileged identity with empty material ran unauth but was      │ Loader rejects role != Anonymous && session.is_empty(); StoredIdentity now carries role so empty-inject fires only for Anonymous. Tests    │
  │ CRITICAL   │ attributed to its privileged label (false "reachable as admin") │ privileged_identity_without_session_fails_loudly, build_injection_nonanon_empty_material_fails_loud                                        │
  │ #1         │                                                                 │                                                                                                                                            │
  ├────────────┼─────────────────────────────────────────────────────────────────┼────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
  │ 🟠 HIGH #2 │ Non-string identity_ref silently dropped → unauth run reported  │ Discriminated: absent/null = run as-is, string = inject, else = fail loud. Tests execute_non_string_identity_ref_fails_without_running,    │
  │            │ as a probe                                                      │ execute_null_identity_ref_runs_as_is                                                                                                       │
  ├────────────┼─────────────────────────────────────────────────────────────────┼────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
  │ 🟠 HIGH #3 │ Malformed identities file swallowed to zero identities → "no    │ Connector now returns Err (refuses to start); missing file still a no-op                                                                   │
  │            │ broken access control" reported without testing                 │                                                                                                                                            │
  └────────────┴─────────────────────────────────────────────────────────────────┴────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘

  Earlier CRITICAL/HIGH leaks (raw ProbeCommand.command serialization, log + stdout/stderr credential leaks) were resolved in prior commits and test-guarded. The custody model (SessionMaterial non-Serialize + redacting
  Debug/Display) is confirmed solid by both reviewers.

  Residual items — legitimately deferrable, not merge gates

  The feature is confirmed inert until matrix#3354 (unprovisioned connector fails closed before any secret reaches argv), so these are rollout gates:
  - 🟡 MEDIUM — exotic header shapes (X-Api-Id: …) still regex-only at the platform log layer → journald/Quickwit. Tracked in pick#335.
  - 🟡 MEDIUM — a quote-containing secret defeats the by-value scrub (format_full_command escapes "); guarded only by a release-compiled-out debug_assert!. This one I could not confirm has a tracked issue — pick#353 covers the
  loader/design nits but not the scrub-before-escape fix specifically.
  - ⬜ Design nits (duplicate-label reject, path perms, Arc+zeroize on SessionMaterial) — tracked in pick#353.

  What to do before merging

  1. Get cchao-devo to re-review a71dbb7 and clear the stale change-request.
  2. Re-approve as ngdevo against the current head (your LGTM is 2 commits stale).
  3. Confirm you're OK deferring the two MEDIUMs — and file/point-to an issue for the quote-escaped-secret scrub (cchao #4) if one doesn't exist, since I couldn't find it tracked.

cchao-devo (approved):
Review of pick PR #317 (re-review @ a71dbb7) — feat(#162): differential-authz identity model

  Verdict: APPROVE — clears my earlier CHANGES_REQUESTED. All three blockers I filed against 50d04e3 are resolved and test-guarded.

  Re-verified the delta 50d04e3..a71dbb7 directly and ran the gates locally (Rust 1.92): cargo fmt --all --check clean, cargo clippy --workspace --locked --features 
  pentest-platform/desktop-pcap -- -D warnings clean, and the new guard tests pass (pentest-core identity 26/0, pentest-tools execute_command 35/0).

  My blockers — all resolved

  1. CRITICAL (mis-provisioned non-anon identity runs unauth, reported as privileged): fixed at the layer I recommended, plus defense-in-depth. The store now holds StoredIdentity {
  role, material } (the role travels with the secret); load_identities_from_file rejects role != Anonymous && session.is_empty() with Error::Config; and build_identity_injection
  takes the empty-inject fast-path only when role == Anonymous, else fails loud — covering direct store-population paths that bypass the file loader. Guards:
  privileged_identity_without_session_fails_loudly, user_identity_with_empty_session_string_fails_loudly, build_injection_nonanon_empty_material_fails_loud (all verified passing).
  2. HIGH (non-string identity_ref silently dropped → unauth): now matches None|Null (run as-is) vs String (inject) vs any other shape (fail loud, naming the JSON type). Guards:
  execute_non_string_identity_ref_fails_without_running (a numeric ref → Failed + "must be a string"), execute_null_identity_ref_runs_as_is.
  3. HIGH (malformed identities file swallowed → authz matrix evaporates): the connector now return Err("refusing to start … unreadable or malformed") instead of warn! + empty
  store, with a correct note that a missing file is still fine (loader returns empty for NotFound).

  Non-blocking residuals (not gating this approval)

  - The by-value secret scrub covers the report and agent-output boundaries; the platform log surface (command.rs) and a secret that is simultaneously quote-containing and an
  exotic non-regex shape still fall back to pattern-redact() (guarded by a release-compiled-out debug_assert!). Bounded — standard Cookie:/Bearer shapes are always caught. Worth a
  tracked follow-up, but consistent with the accepted residuals; does not block merge.

  Thanks for the thorough, test-guarded fixes. Approving; my stale CHANGES_REQUESTED can be dismissed.

PR #332, opened 2026-07-27
fix(ssrf): route zap target through the shared SSRF guard (#311)

### LINKED ISSUE #289: Tool target validation hardcodes ValidationMode::Production — ignores PENTEST_ALLOW_PRIVATE_IPS (nikto/ffuf reject private targets in non-prod)

### LINKED ISSUE #311: SSRF: zap tool has its own validate_url with no IP check, bypassing the shared guard
## Summary

`crates/tools/src/external/zap.rs:158` defines a **local** `validate_url` that checks only the URL scheme and shell metacharacters. It performs no SSRF IP validation, so the `zap` tool will happily scan the cloud-metadata service and RFC-1918 hosts regardless of the operator's `PENTEST_ALLOW_PRIVATE_IPS` setting.

The four sibling web tools that make outbound requests -- `nikto`, `ffuf`, `dirb`, `gobuster` -- all call the shared `pentest_core::url_validation::validate_url`, which classifies the resolved IP and blocks link-local / metadata / private ranges per the active `ValidationMode`. `zap` is the only fetching tool that does not.

## Evidence

`zap.rs:158-183` in full is a scheme prefix check plus a shell-metacharacter reject:

```rust
fn validate_url(raw: &str) -> Result<String> {
    // ... rejects empty
    if !(url.starts_with("http://") || url.starts_with("https://")) { /* err */ }
    // rejects ; & | ` $ < > \n \r \ " '
    Ok(url.to_string())
}
```

There is no `IpAddr` parse, no DNS resolution, and no `ValidationMode`. Compare `nikto.rs:94`:

```rust
let target = validate_url(&target, ValidationMode::Production, None)?;
```

The divergence is visible in the tool's own tests: `zap.rs:264` asserts that `" http://10.0.0.5:8080/ "` is **accepted**, pinning the permissive behavior as intended. `http://169.254.169.254/latest/meta-data/` satisfies every check in the local function.

`ZapTool` is registered unconditionally (`crates/tools/src/lib.rs:156`) for `Platform::Desktop` and `Platform::Tui`, and `target` is a required, LLM-supplied string parameter (`zap.rs:61`). So the reachable path is: agent emits a metadata or private-range URL as `target` -> local guard passes it -> `zap-baseline.py -t <url>` fetches it.

## Scope note

I audited every tool in `crates/tools/src/external/` that takes a `target` parameter. Two categories are **not** part of this issue:

- `burp` takes a `target` but never makes a request -- `build_guidance()` only formats instructional text for a human operator. Correctly unvalidated.
- `nmap`, `masscan`, `rustscan`, `hydra`, `enum4linux` are LAN/host scanners whose entire purpose is private-range targets. A URL-shaped SSRF guard is the wrong control for them; they need scope/authorization controls, which is a separate design question. Not in scope here.

`zap` is specifically a **web DAST** tool in the same category as `nikto`/`ffuf`/`dirb`/`gobuster`, which is why the inconsistency matters.

## Why P2 and not higher

This is an operator-facing pentest tool, so "it scanned a private IP" is often the intended use, not a breach. The severity is that the control is *inconsistent and undocumented*: an operator who sets `PENTEST_ALLOW_PRIVATE_IPS=false` (or leaves it unset in a release build) reasonably expects the metadata service to be off-limits across web tools, and gets that for four of five. The metadata-service path is the genuinely sensitive one, since the connector may run on a cloud host with an instance role.

## Proposed fix

Replace the local function's body with a call to the shared guard, keeping the shell-metacharacter check as a defensive layer:

- [ ] `zap.rs` calls `pentest_core::url_validation::validate_url` with the mode from the shared `PENTEST_ALLOW_PRIVATE_IPS` resolver, matching how `nikto`/`ffuf`/`dirb`/`gobuster` do it.
- [ ] Retain the metacharacter reject (defense in depth even though args go out as an argv vector, never a shell string).
- [ ] Update the test at `zap.rs:264` -- it currently asserts private IPs are accepted, which will invert. This is a deliberate behavior change and should be called out in the PR body, since an operator relying on `zap` against RFC-1918 targets now needs `PENTEST_ALLOW_PRIVATE_IPS=true`.
- [ ] Add a test asserting `http://169.254.169.254/` is blocked in `Production` and in `PrivateNetwork`.
- [ ] Guard-check it: reverting the call must turn the new test red.

## Dependency

Should land **after** PR #296, which introduces `target_validation_mode()` as the shared `PENTEST_ALLOW_PRIVATE_IPS` resolver for exactly this purpose. Building on `main` today would mean hardcoding `ValidationMode::Production` and then reworking it.

## References

- Raised as a non-blocking note by the reviewer on PR #296 ("`zap.rs` has its own local `validate_url` (shell-metachar only, no SSRF IP check) -- pre-existing separate path, not touched here; the 'all web tools share one source of truth' framing shouldn't be over-read"), then verified independently.
- Related: #232, #236, #289, #310.

**Recommended agent:** security-reviewer

jtomek-strike48:
## Summary

- `zap` defined a **local** `validate_url` that checked only URL scheme + shell metacharacters, with no IP classification. It was the only *fetching* web tool skipping the shared SSRF guard, so it would scan the cloud-metadata service (`169.254.169.254`) and RFC-1918 hosts regardless of `PENTEST_ALLOW_PRIVATE_IPS`. Closes #311.
- Compose two layers behind a new `validate_target(raw, mode)` helper: (1) `sanitize_target_url` (the former local scheme + metachar pre-check, renamed, kept as defense-in-depth), then (2) the shared `pentest_core::url_validation::validate_url`, matching how `nikto`/`ffuf`/`dirb`/`gobuster` gate.
- Extracting a pure, mode-parameterized helper is what makes the **wiring** testable: the production `execute` path only ever sees `Development` mode in debug builds, which would mask the guard.

## Stacking

- **Base is `fix/ssrf-ipv6-canonicalization-232-236-289` (PR #296), not `main`** — this depends on `target_validation_mode()`, which #296 introduces. Merge #296 first, then this retargets to `main` automatically. Building on `main` today would mean hardcoding `ValidationMode::Production` and reworking it (per the #311 dependency note).

## Behavior change (called out per Lens 8)

`zap` now **blocks RFC-1918 targets in Production** (release default). An operator must set `PENTEST_ALLOW_PRIVATE_IPS=true` to scan private targets — same contract as the sibling web tools. The old test asserting a private IP was accepted is inverted accordingly.

## Test plan

- [x] `cargo test -p pentest-tools --lib zap` — 11 pass (incl. 2 new #311 guards)
- [x] **Guard-verified by mutation:** neutering layer 2 inside `validate_target` turns `zap_blocks_metadata_via_shared_guard` + `zap_blocks_private_ip_in_production` red; restored → green. (First attempt tested `validate_url` directly and stayed green under mutation — caught and fixed by pointing the tests at the composed wiring.)
- [x] `cargo test -p pentest-tools --lib` — 325 pass, 6 ignored, 0 failed
- [x] `cargo fmt --all` clean; `cargo clippy -p pentest-tools --all-targets -- -D warnings` clean
- [x] Updated two stale doc-comments referencing the renamed `zap::validate_url` (`metasploit.rs`, `ad/common.rs`)

## Closes #289

Per an acceptance-criteria audit: #289 ("have per-tool target validation honor PENTEST_ALLOW_PRIVATE_IPS — audit all validate_url callers in crates/tools/src/external/") is code-complete once this PR lands. PR #296 converted nikto/ffuf/dirb/gobuster to target_validation_mode() and "addresses #289", but zap was the one remaining caller not using the shared guard — the exact gap this PR closes. With both merged, every fetching web tool honors the opt-in, satisfying #289's audit-all-callers criterion.

### DIFF
--- a/crates/tools/src/external/ad/common.rs
+++ b/crates/tools/src/external/ad/common.rs
@@ -1,13 +1,13 @@
 //! Shared validation helpers for the Active Directory attack tools.
 //!
 //! These keep the four AD wrappers DRY: a single metacharacter guard (copied
-//! from `zap::validate_url`'s check) and a conservative path/flag allowlist used
+//! from `zap::sanitize_target_url`'s check) and a conservative path/flag allowlist used
 //! for free-form positional values such as wordlist paths and `extra_args`.
 
 use pentest_core::error::Result;
 
 /// Reject shell metacharacters defensively even though args are passed as an
-/// argv vector (never a shell string). Mirrors `zap::validate_url` and
+/// argv vector (never a shell string). Mirrors `zap::sanitize_target_url` and
 /// `metasploit::reject_metacharacters`.
 ///
 /// Passwords may legitimately contain symbols, so callers validating a password
--- a/crates/tools/src/external/metasploit.rs
+++ b/crates/tools/src/external/metasploit.rs
@@ -331,7 +331,7 @@ fn validate_lport(lport: u64) -> Result<()> {
 }
 
 /// Reject shell metacharacters defensively even though args are passed as an
-/// argv vector (never a shell string). Mirrors `zap::validate_url`.
+/// argv vector (never a shell string). Mirrors `zap::sanitize_target_url`.
 fn reject_metacharacters(value: &str, field: &str) -> Result<()> {
     if value.chars().any(|c| {
         matches!(
--- a/crates/tools/src/external/zap.rs
+++ b/crates/tools/src/external/zap.rs
@@ -17,6 +17,7 @@ use pentest_core::tools::{
     execute_timed_with_provenance, ExternalDependency, ParamType, PentestTool, Platform,
     ToolCategory, ToolContext, ToolParam, ToolResult, ToolSchema,
 };
+use pentest_core::url_validation::{target_validation_mode, validate_url, ValidationMode};
 use pentest_platform::{get_platform, CommandExec};
 use serde_json::{json, Value};
 use std::time::Duration;
@@ -83,7 +84,7 @@ impl PentestTool for ZapTool {
             let platform = get_platform();
 
             let target = param_str_or(&params, "target", "");
-            let target = validate_url(&target)?;
+            let target = validate_target(&target, target_validation_mode())?;
             let scan_type = param_str_or(&params, "scan_type", "baseline");
             let timeout_secs = param_u64(&params, "timeout", 300);
 
@@ -153,9 +154,32 @@ impl PentestTool for ZapTool {
     }
 }
 
-/// Validate that the target is a well-formed http(s) URL and contains no shell
-/// metacharacters. Returns the normalized URL string.
-fn validate_url(raw: &str) -> Result<String> {
+/// Two-layer target validation for the zap tool (#311).
+///
+/// 1. [`sanitize_target_url`] — a local scheme + shell-metacharacter pre-check.
+/// 2. The SHARED SSRF guard [`pentest_core::url_validation::validate_url`],
+///    which classifies the resolved IP and blocks link-local / cloud-metadata /
+///    private ranges per `mode`.
+///
+/// zap was previously the only *fetching* web tool that skipped layer 2, so it
+/// would scan `169.254.169.254` regardless of `PENTEST_ALLOW_PRIVATE_IPS`; the
+/// sibling web tools (nikto/ffuf/dirb/gobuster) all gate this way. Extracted as
+/// a pure, mode-parameterized helper so the wiring itself is unit-testable
+/// (the production `execute` path only ever gets `Development` in debug builds,
+/// which would mask the guard).
+fn validate_target(raw: &str, mode: ValidationMode) -> Result<String> {
+    let pre = sanitize_target_url(raw)?;
+    validate_url(&pre, mode, None)
+}
+
+/// Defense-in-depth pre-check: the target must be a well-formed http(s) URL and
+/// contain no shell metacharacters. Returns the trimmed URL string.
+///
+/// This does NOT perform SSRF/IP classification — that is layer 2 in
+/// [`validate_target`]. Kept as a separate layer so a bad scheme or an injected
+/// metacharacter is rejected with a specific message before DNS resolution, and
+/// so the metachar guard survives even if the shared call is ever refactored.
+fn sanitize_target_url(raw: &str) -> Result<String> {
     let url = raw.trim();
     if url.is_empty() {
         return Err(pentest_core::error::Error::InvalidParams(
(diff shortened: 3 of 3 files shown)


jtomek-strike48:
Superseded by #334: after #296 merged (and its branch was deleted), this PR auto-closed. Rebased the zap commit onto `main` and reopened as #334.

PR #333, opened 2026-07-27
fix(port_scan): distinguish unreachable hosts from closed ports (#306)

### LINKED ISSUE #306: port_scan reports unreachable hosts as scanned-and-clean
## Problem

`probe_port` collapses every connection failure into `open: false`, so a port that is genuinely closed and a host that was never reachable produce identical output. In a pentest report those are very different facts.

`crates/platform/src/common.rs:48`:

```rust
let open = TcpStream::connect_timeout(&socket_addr, timeout).is_ok();
```

`EHOSTUNREACH`, `ENETUNREACH`, `EACCES` (blocked by a sandbox or a missing capability) and a genuine timeout all become `open: false` with no error recorded.

The multi-host caller in `crates/tools/src/port_scan.rs` builds an `errors[]` array, but it is only populated from host-level `Err` returns. The per-port TCP path never produces one, so the array looks like error handling while remaining empty for this whole class of failure.

## Impact

Scanning an unreachable subnet yields `hosts_scanned: 254, hosts_with_open_ports: 0, errors: []`. The agent reads that as a clean network and writes a report saying nothing is exposed, when in fact no packet ever reached the target. For a tool whose output an operator may hand to a client, "we checked and found nothing" versus "we could not check" must be distinguishable.

## Reproduction

```bash
# Point at an unroutable address; note the successful, empty result.
# probe_port returns open:false for every port rather than surfacing EHOSTUNREACH.
rg -n 'is_ok\(\)' crates/platform/src/common.rs
```

The relevant assertion is that no code path between `connect_timeout` and the tool's JSON output inspects `io::Error::kind()`.

## Suggested fix

Have `probe_port` return the `io::ErrorKind` alongside the result, and classify:

- `ConnectionRefused` is a genuine closed port, keep `open: false`.
- `TimedOut` is indeterminate, report separately from closed.
- `HostUnreachable` / `NetworkUnreachable` / `PermissionDenied` are scan failures, surface into `errors[]` and stop attributing them to the port.

Then make the tool's summary distinguish "0 open ports across 254 reachable hosts" from "254 hosts unreachable". Regression test: a probe against an unroutable address must not report a scanned-and-clean host.

## Notes

Pre-existing on `main` (verified at `crates/platform/src/common.rs:48` on current `main`), not introduced by any open PR. Surfaced while reviewing #249, whose batch `subnet`/`hosts[]` support widens the blast radius to 1024 hosts per call, but the defect is independent of that PR.

jtomek-strike48:
## Summary

- `probe_port` classified every `connect_timeout` failure with `is_ok()`, so a refused port, a timeout, and a host that was never reachable all collapsed into `open: false`. A scan of an unroutable subnet returned `open_count: 0` with no errors, which the agent reads as "checked and clean" when no packet ever reached the target. Fixes #306.
- Inspect `io::ErrorKind` and classify into a new `PortState` enum: `ConnectionRefused` -> `Closed`; `HostUnreachable`/`NetworkUnreachable`/`PermissionDenied` -> `Unreachable`; `TimedOut`/other -> `Filtered` (never reported as closed).
- `ScanResult` gains `unreachable_count` + `errors[]`; `ScannedPort` gains `state` (`open: bool` retained for back-compat). The tool JSON surfaces both new fields, so a scan whose failures carry an unreachable errno is distinguishable from a clean one.
- Scope note (verified empirically): a host is classified `Unreachable` only when the OS supplies the errno (`HostUnreachable`/`NetworkUnreachable`/`PermissionDenied` - e.g. IPv6 no-route, on-link ARP failure, sandbox denial). On a normally-routed network, connecting to a dead IPv4 host **times out** before the OS emits `HostUnreachable`, so it classifies as `Filtered`, not `Unreachable`. That is deliberate: an all-timeout result is genuinely ambiguous between "host down" and "host silently drops all packets" (policy firewall), and inferring "down" would false-positive on live-but-firewalled hosts. Inferring host-down from an all-timeout/zero-response scan is tracked as a follow-up in #337.
- `tcp_port_scan` now owns timing/tallies and returns `ScanResult`, collapsing the three identical per-platform wrappers (desktop/android/ios) into one assembly path so they cannot drift on how a failed probe is counted.

## Test plan

- [x] `cargo test -p pentest-platform --lib common::` - 8 new unit tests pass (classify + assemble seam, network-free)
- [x] `cargo test -p pentest-core --test tool_integration port_scan` - 4 existing integration tests pass
- [x] `cargo check --workspace --all-targets` - clean
- [x] `cargo clippy --workspace --all-targets -- -D warnings` - clean (only pre-existing `screenshots` future-incompat warning)
- [x] Mutation-verified (Lens 1): reverting `classify_connect` to collapse-all-to-closed turns the classifier tests red; zeroing the tally in `assemble_scan_result` turns the "scanned-and-clean" regression tests red
- [x] CI-scope (Lens 2): tests are unfeatured `#[cfg(test)]` in `pentest-platform` `--lib`, so they run in both the Linux (`--workspace`) and macOS (package-scoped `--lib`) required lanes

## Notes

Acceptance criteria from #306, honestly scoped: `ErrorKind` inspected and classified; refused stays closed; timeout reported separately (`Filtered`, never closed); host/network-unreachable + permission-denied surfaced into `errors[]` and `unreachable_count`; tool summary distinguishes an errno-unreachable scan from a clean one; regression test for an unroutable address. Partial: the "dead host on a normal network" case reaches the target via a timeout, so it reports as `Filtered` rather than `Unreachable` (see the scope note above) - full host-down inference from an all-timeout scan is a tracked follow-up (#337), not delivered here. `ScannedPort`/`ScanResult` are wire types (serialized outward, never deserialized), so the additive fields need no serde back-compat.

### DIFF
--- a/crates/platform/src/android/network.rs
+++ b/crates/platform/src/android/network.rs
@@ -2,24 +2,13 @@
 
 use crate::traits::*;
 use pentest_core::error::Result;
-use std::time::{Duration, Instant};
+use std::time::Duration;
 
 /// Perform a port scan
 pub async fn port_scan(config: ScanConfig) -> Result<ScanResult> {
-    let start = Instant::now();
     let timeout = Duration::from_millis(config.timeout_ms);
 
-    let ports = crate::common::tcp_port_scan(&config.host, &config.ports, timeout, 0).await;
-
-    let open_count = ports.iter().filter(|p| p.open).count();
-    let duration_ms = start.elapsed().as_millis() as u64;
-
-    Ok(ScanResult {
-        host: config.host,
-        ports,
-        duration_ms,
-        open_count,
-    })
+    Ok(crate::common::tcp_port_scan(&config.host, &config.ports, timeout, 0).await)
 }
 
 /// Get the ARP table with layered fallback (bd-23):
--- a/crates/platform/src/common.rs
+++ b/crates/platform/src/common.rs
@@ -3,10 +3,11 @@
 //! This module extracts duplicated port-scanning and ARP-parsing code that was
 //! previously copy-pasted across android, desktop, and iOS implementations.
 
-use crate::traits::{port_to_service, ArpEntry, ScannedPort};
+use crate::traits::{port_to_service, ArpEntry, PortState, ScanResult, ScannedPort};
+use std::io;
 use std::net::{SocketAddr, TcpStream, ToSocketAddrs};
 use std::sync::Arc;
-use std::time::Duration;
+use std::time::{Duration, Instant};
 use tokio::sync::Semaphore;
 
 /// The command that reports whether a binary is on the host's PATH.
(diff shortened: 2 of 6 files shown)


cchao-devo (approved):
Review: Pick PR #333 — fix(port_scan): distinguish unreachable hosts from closed ports (#306)

  Verdict: ✅ Approve — the core silent-failure is genuinely closed; a few residuals to flag (mostly pre-existing / tracked). The errno-unreachable arm of #306 is correctly fixed and
  verified by execution, the platform consolidation is a real anti-drift improvement, and gates are green. Two residual silent-failure paths exist that this PR doesn't close — one
  worth surfacing before merge.

  Lenses run: the real silent-failure-hunter (lead) + my own empirical runs of classify_connect/assemble_scan_result + source verification of every residual finding + gates. Verified
  in pick_08/pick.

  Gates (on pr-333)
  
  - cargo fmt --all --check clean · cargo clippy --workspace --all-targets -- -D warnings clean · common:: unit 8/0 · port_scan integration 4/0

  The core fix is correct (verified by running it)

  classify_connect maps (ran it): ConnectionRefused→Closed, HostUnreachable/NetworkUnreachable/PermissionDenied→Unreachable, TimedOut/other→Filtered (never Closed).
  assemble_scan_result tallies correctly — mixed 8-port vec → open=2, unreachable=3, Filtered counts as neither (no double-count/miss). The inner spawn_blocking join failure → 
  Unreachable (not silent open:false) is a subtle silent-failure path the author caught. port_scan.rs surfaces unreachable_count + errors[] to the agent JSON, and per-port state
  serializes too — so the errno-unreachable case is distinguishable end-to-end. The 3 platform wrappers now delegate to one tcp_port_scan→ScanResult (verified: only one ScanResult{}
  constructor remains), so they can't drift on failure counting. Wire types are serialize-out-only, additive fields need no back-compat. All solid.

  Residual silent-failure paths (verified at source) — the review value here

  1. 🟠 HIGH-ish, but PRE-EXISTING (not introduced by this PR): DNS-resolution failure silently probes 127.0.0.1. resolve_addr (unchanged by this diff — confirmed) falls back to
  ([127,0,0,1], port) when a hostname doesn't resolve. So an unresolvable-host scan probes localhost → reads Closed (clean) or even Open (misattributing localhost's ports to the
  target). This is squarely the #306 "failed scan reads as clean" class — arguably worse — but it predates this PR and isn't touched. Worth flagging: this fix improves the errno case
  while leaving the DNS case fully silent; it deserves either a fix (resolution failure → Unreachable + error) or an explicit follow-up issue rather than being left undocumented. Not a
  blocker for this diff (it doesn't regress it), but the author's "#306 fixed" framing should note DNS failure is still open.
  2. 🟡 MEDIUM — the common "dead host on a routed network" case is summary-invisible (honestly tracked to #337). Confirmed by running: an all-Filtered scan (dead host times out) and
  an all-Closed scan (genuinely clean) produce byte-identical summaries (open=0, unreachable=0, errors=[]). They differ only in per-port ports[].state (filtered vs closed). So the
  fix's headline promise ("tell 'checked and found nothing' from 'could not check'") holds for the errno case but NOT the more-common timeout case at the summary level. The author
  documents this and defers to #337 — acceptable, but the PR should state plainly that unreachable_count==0 && errors==[] does not yet mean "host was reached" so downstream agent logic
  doesn't treat it as such.
  3. 🟡 MEDIUM — outer-task join failure silently drops a port. The inner join failure is mapped to Unreachable (good), but the outer loop is if let Ok(result) = handle.await { push }
  with no else — an outer-task panic (e.g. the semaphore.acquire().unwrap() at line 172) drops that port entirely, silently shrinking total_scanned with no signal. The author applied
  the "join failure must not vanish" fix to the inner join but not the outer. Small fix: push an Unreachable/error ScannedPort in the else arm (or assert results.len() == ports.len()).
  Low likelihood (requires a panic), but it's the exact silent-failure class and mirrors a fix already made one layer in.
  4. 🟢 LOW note: ConnectionReset/ConnectionAborted/AddrNotAvailable/NetworkDown all bucket to Filtered via the catch-all. Safe direction (nothing false-collapses to Closed);
  NetworkDown is the one arguably worth promoting to Unreachable, overlaps #337's scope.

  On the honest-scope judgment call

  The author's decision to classify dead-host-on-normal-network as Filtered not Unreachable (timeout is genuinely ambiguous between "host down" and "silently firewalled") is correct —
  inferring "down" from all-timeout would false-positive on live-but-firewalled hosts. Deferring host-down inference to #337 is the right call. My only ask is the summary-invisibility
  (finding 2) be stated explicitly so agents don't misread.

  init-dev relevance

  Low — port_scan is a scan tool; this improves result fidelity (agent can now distinguish unreachable from clean for the errno case). No connector/cluster-contract surface.

PR #331, opened 2026-07-27
feat(security): sanitize tool output before it re-enters an LLM agent (#320)

jtomek-strike48:
## Summary

- Raw tool stdout/stderr reached the Red Team, Validator, and Report agents verbatim - an indirect prompt-injection surface and a secret/PII leak path. Closes #320 (P0).
- Adds a single sanitization pass at the **production** agent-facing choke point - `PickConnector::execute_with_context`'s return (`crates/ui/src/liveview_connector/pick_connector.rs:396`), the connector the SDK runner actually drives - so all ~40 output-producing tools, present and future, are covered without each author opting in. The same pass is also applied at `PentestConnector::execute` (`crates/core/src/connector.rs:250`) so that connector's path (used by integration tests and any other consumer) is covered too.
- New `crates/core/src/sanitize.rs`: secret-scrubs `data`/`error` string leaves by reusing `provenance::redact` (policies can't drift); neutralizes a conservative, high-precision set of injected-instruction markers (override phrases, new-instructions preambles, fake `system`/`assistant` role tags) into inert text - neutralized, not dropped; flags suspected injection with a `_sanitization` marker that travels with the payload plus an audit log line.

## Design notes

- **Choke point over per-tool:** the issue's example put the sanitizer in `crates/tools`, but the threat model spans ~40 tools. The connector return is the one place every `ToolResult` is serialized to the agent - one pass there can't be forgotten on tool #41. The production runner drives `PickConnector::execute_with_context` (`PickConnector::execute` is a stub that errors), so that return is the choke point that matters for #320.
- **`provenance.raw_response_excerpt` is sanitized:** it holds attacker-controlled target-response bytes and is serde-included in the agent-facing value. Emit-time `truncate_excerpt` only redacts secrets, not injection markers, so an instruction in a banner reached the model via provenance even with `data`/`error` clean. The sanitizer now neutralizes markers in that one field; the rest of `provenance` (tool name, version, `probe_commands`, timestamp) stays byte-for-byte audit metadata (`leaves_provenance_audit_metadata_untouched`).
[...]

PR #334, opened 2026-07-27
fix(ssrf): route zap target through the shared SSRF guard (#311)

### LINKED ISSUE #311: SSRF: zap tool has its own validate_url with no IP check, bypassing the shared guard

jtomek-strike48:
## Summary

- `zap` had its own `validate_url` with no IP-address check, so it bypassed the shared SSRF guard. A target-controlled hostname resolving to cloud-metadata or an RFC-1918 address reached ZAP unchecked. Fixes #311.
- Route the zap target through the shared guard via `validate_target(raw, mode)`, matching the sibling web tools that already call it (ffuf/nikto/dirb/gobuster). Note: `nuclei` still has no SSRF guard - same unfixed bug class, tracked separately, out of scope here.
- Behavior change: zap now blocks RFC-1918 targets in Production (release default); an operator sets `PENTEST_ALLOW_PRIVATE_IPS=true` to scan private targets. The old test asserting a private IP was accepted is inverted accordingly.
- Updates two stale doc-comments referencing the renamed `zap::validate_url`.

## Test plan

- [x] `cargo test -p pentest-tools zap` - 11 pass (rebased onto merged #296)
- [x] `cargo check -p pentest-core` - clean on top of merged #296

## Notes

Previously stacked on #296 (now merged to main); rebased so this is an independent PR. #296's shared-guard refactor (`target_validation_mode`) is the dependency this consumes.

### DIFF
--- a/crates/tools/src/external/ad/common.rs
+++ b/crates/tools/src/external/ad/common.rs
@@ -1,13 +1,13 @@
 //! Shared validation helpers for the Active Directory attack tools.
 //!
 //! These keep the four AD wrappers DRY: a single metacharacter guard (copied
-//! from `zap::validate_url`'s check) and a conservative path/flag allowlist used
+//! from `zap::sanitize_target_url`'s check) and a conservative path/flag allowlist used
 //! for free-form positional values such as wordlist paths and `extra_args`.
 
 use pentest_core::error::Result;
 
 /// Reject shell metacharacters defensively even though args are passed as an
-/// argv vector (never a shell string). Mirrors `zap::validate_url` and
+/// argv vector (never a shell string). Mirrors `zap::sanitize_target_url` and
 /// `metasploit::reject_metacharacters`.
 ///
 /// Passwords may legitimately contain symbols, so callers validating a password
--- a/crates/tools/src/external/metasploit.rs
+++ b/crates/tools/src/external/metasploit.rs
@@ -331,7 +331,7 @@ fn validate_lport(lport: u64) -> Result<()> {
 }
 
 /// Reject shell metacharacters defensively even though args are passed as an
-/// argv vector (never a shell string). Mirrors `zap::validate_url`.
+/// argv vector (never a shell string). Mirrors `zap::sanitize_target_url`.
 fn reject_metacharacters(value: &str, field: &str) -> Result<()> {
     if value.chars().any(|c| {
         matches!(
--- a/crates/tools/src/external/zap.rs
+++ b/crates/tools/src/external/zap.rs
@@ -17,6 +17,7 @@ use pentest_core::tools::{
     execute_timed_with_provenance, ExternalDependency, ParamType, PentestTool, Platform,
     ToolCategory, ToolContext, ToolParam, ToolResult, ToolSchema,
 };
+use pentest_core::url_validation::{target_validation_mode, validate_url, ValidationMode};
 use pentest_platform::{get_platform, CommandExec};
 use serde_json::{json, Value};
 use std::time::Duration;
@@ -83,7 +84,7 @@ impl PentestTool for ZapTool {
             let platform = get_platform();
 
             let target = param_str_or(&params, "target", "");
-            let target = validate_url(&target)?;
+            let target = validate_target(&target, target_validation_mode())?;
             let scan_type = param_str_or(&params, "scan_type", "baseline");
             let timeout_secs = param_u64(&params, "timeout", 300);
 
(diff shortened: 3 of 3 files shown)


cchao-devo (approved):
Review: Pick PR #334 — fix(ssrf): route zap target through the shared SSRF guard (#311)
  
  Verdict: ✅ Approve. Closes the exact zap.rs-bypass gap I flagged during #296. Zap now runs the shared SSRF guard; verified by execution that all metadata encodings are blocked and
  PENTEST_ALLOW_PRIVATE_IPS is honored. Clean, correct, well-tested. No blockers.

  Lenses run: manual source trace + empirical runs of zap's validate_target against metadata (incl. #296's encoded forms) + private/public, both modes + no-residual-bypass check +
  gates. Verified in pick_08/pick.

  Gates (on pr-334)
  
  - cargo fmt --all --check clean · cargo clippy --workspace --all-targets -- -D warnings clean · zap tests 11/0 (incl. the 2 new SSRF guard tests)

  The fix is correct and closes #311 (verified by running it)

  Ran zap's composed validate_target on the full battery:

  ┌─────────────────────────────────┬────────────┬─────────────────────┐
  │             Target              │ Production │   PrivateNetwork    │
  ├─────────────────────────────────┼────────────┼─────────────────────┤
  │ bare metadata 169.254.169.254   │ blocked ✅ │ blocked ✅          │
  ├─────────────────────────────────┼────────────┼─────────────────────┤
  │ mapped [::ffff:169.254.169.254] │ blocked ✅ │ blocked ✅          │
  ├─────────────────────────────────┼────────────┼─────────────────────┤
  │ compatible [::169.254.169.254]  │ blocked ✅ │ blocked ✅          │
  ├─────────────────────────────────┼────────────┼─────────────────────┤
  │ NAT64 [64:ff9b::a9fe:a9fe]      │ blocked ✅ │ blocked ✅          │
  ├─────────────────────────────────┼────────────┼─────────────────────┤
  │ RFC1918 10.0.0.5                │ blocked ✅ │ allowed ✅ (opt-in) │
  ├─────────────────────────────────┼────────────┼─────────────────────┤
  │ public example.com              │ allowed ✅ │ allowed ✅          │
  └─────────────────────────────────┴────────────┴─────────────────────┘

  So zap now inherits #296's hardened classifier (all four encoded-metadata forms blocked) and honors PENTEST_ALLOW_PRIVATE_IPS via target_validation_mode() — identical posture to the
  sibling web tools (nikto/ffuf/dirb/gobuster). Before this, zap would scan 169.254.169.254 unconditionally.

  Design is clean

  - Two-layer: sanitize_target_url (local scheme + shell-metachar pre-check, kept as defense-in-depth — survives even if the shared call is refactored) → shared
  url_validation::validate_url(pre, mode, None) (the SSRF IP classifier). Right separation.
  - No residual bypass (verified): execute (line 87) is the sole call site; the old IP-check-less validate_url is renamed to sanitize_target_url and fully replaced — no path skips
  layer 2.
  - Production-path confirmed (the #331 lesson): execute calls validate_target(&target, target_validation_mode()); target_validation_mode() returns Production in release. The
  doc-comment's "execute only gets Development in debug" note is just explaining why the helper is mode-parameterized (so unit tests can force Production regardless of build profile) —
  not a production gap.
  - Tests exercise the composed validate_target (not the shared guard directly), so a revert of layer 2 turns them red — the right altitude, with a mutation-guard note.
  - ad/common.rs + metasploit.rs changes are pure doc-comment ref updates (zap::validate_url→sanitize_target_url), no logic.

  Honest scope

  The PR correctly notes nuclei still has no SSRF guard (same bug class, tracked separately, out of scope). That's the right call — one tool per PR, and #311 was specifically the zap
  gap I'd flagged. The behavior change (zap now blocks RFC-1918 in Production) is disclosed and the old "accepts private IP" test is correctly inverted.

  init-dev relevance

  This is the production tool path init-dev's headless connector runs. The RFC-1918-blocked-in-Production behavior change matters for init-dev: engagement scans of private .svc/cluster
  targets via zap now require PENTEST_ALLOW_PRIVATE_IPS=true — but that's already how init-dev runs the other web tools (nikto/ffuf/etc.), so zap just joins the consistent posture. No
  new wiring needed; consistent with the #231 PrivateNetwork work.

PR #336, opened 2026-07-27
security(pii): load banned-name list from outside the repo

jtomek-strike48:
## Summary

- The PII scanner (`scripts/check-pii.sh`) hardcoded the real customer/tenant names in its source. Because this repo is public, it published the exact list it exists to keep out of public artifacts.
- Removes the inline list and loads it at runtime from outside the repo: `$PII_NAMES` (CI, from a repository secret) or a gitignored `.pii-names.local` (local dev, seeded from a placeholder-only example).
- Fails loud (exit 2) when no name list is available, so the scanner can never silently pass with an empty list.
- Stops echoing the name list in the "PII detected" error message, since that output lands in public CI logs.

## Follow-up required (not in this PR)

This removes the names from `HEAD` only. They remain in git history from the scanner's introduction and are readable via `git log`. Scrubbing history needs a separate, coordinated force-push to `main` (owner decision). Because the repo is public, third-party clones/caches may already exist.

## Operator action after merge

Set the `PII_NAMES` repository secret (Settings -> Secrets and variables -> Actions), newline- or comma-separated, or the PII Check job fails by design.

## Test plan

- [x] `shellcheck -s bash scripts/check-pii.sh` clean
- [x] Detects a banned name via `.pii-names.local`, `$PII_NAMES`, and `$PII_NAMES_FILE`
- [x] Clean input passes (exit 0); word boundaries hold (`maidart` alone does not match)
- [x] No name list anywhere -> exit 2 (fail loud)
- [x] `--text`, stdin, and file modes all work; CI/hook `git diff | grep '^+[^+]'` pipeline still catches added lines
- [x] No customer names remain in any tracked file (script, workflow, example)
- [x] Repo's own pre-commit + commit-msg PII hooks pass on this change

### DIFF
--- a/.github/workflows/pii-check.yml
+++ b/.github/workflows/pii-check.yml
@@ -13,6 +13,11 @@ jobs:
   check-pii:
     name: Scan for customer names
     runs-on: ubuntu-latest
+    # The banned-name list is injected from a repository secret so the customer
+    # names never live in this public repo. Configure it under Settings ->
+    # Secrets and variables -> Actions -> PII_NAMES (newline- or comma-separated).
+    env:
+      PII_NAMES: ${{ secrets.PII_NAMES }}
     steps:
       - uses: actions/checkout@v4
         with:
@@ -20,6 +25,14 @@ jobs:
 
       - name: Ensure scanner is executable
         run: chmod +x scripts/check-pii.sh
+
+      - name: Verify PII name list is configured
+        run: |
+          if [[ -z "${PII_NAMES:-}" ]]; then
+            echo "::error::PII_NAMES secret is not set - the scanner cannot run."
+            echo "Add it under Settings -> Secrets and variables -> Actions."
+            exit 1
+          fi
 
       - name: Check PR title
         env:
--- a/.github/workflows/pii-check.yml
+++ b/.github/workflows/pii-check.yml
@@ -26,6 +26,13 @@ jobs:
       - name: Ensure scanner is executable
         run: chmod +x scripts/check-pii.sh
 
+      # Runs regardless of the PII_NAMES secret (it supplies its own names), so
+      # the metacharacter-name regression is guarded on every PR, forks included.
+      - name: Scanner self-test (metacharacter names)
+        run: |
+          chmod +x scripts/check-pii.test.sh
+          scripts/check-pii.test.sh
+
       # GitHub does NOT pass repository secrets to workflows triggered by a
       # pull_request from a fork, so `PII_NAMES` is empty on every fork PR. A
       # hard failure there would permanently red-x external contributions for a
@@ -78,6 +85,6 @@ jobs:
           echo "Scanning diff for added lines between $BASE_SHA..$HEAD_SHA"
           # Only scan added lines (lines starting with '+' but not '+++') to avoid
           # tripping on unchanged historical content or on removal of bad content.
-          git diff "$BASE_SHA..$HEAD_SHA" -- . ':!scripts/check-pii.sh' ':!.github/workflows/pii-check.yml' \
+          git diff "$BASE_SHA..$HEAD_SHA" -- . ':!scripts/check-pii.sh' ':!scripts/check-pii.test.sh' ':!.github/workflows/pii-check.yml' \
             | grep -E '^\+[^+]' \
             | scripts/check-pii.sh
--- a/.github/workflows/pii-check.yml
+++ b/.github/workflows/pii-check.yml
@@ -26,15 +26,26 @@ jobs:
       - name: Ensure scanner is executable
         run: chmod +x scripts/check-pii.sh
 
-      - name: Verify PII name list is configured
+      # GitHub does NOT pass repository secrets to workflows triggered by a
+      # pull_request from a fork, so `PII_NAMES` is empty on every fork PR. A
+      # hard failure there would permanently red-x external contributions for a
+      # reason they cannot fix. Instead: when the list is available (same-repo
+      # PRs), the scan below is MANDATORY and hard-fails on a match; when it is
+      # absent (fork PRs), we skip the scan with a loud warning and rely on the
+      # same-repo re-run + maintainer review as the backstop. The skip is never
+      # silent.
+      - name: Detect whether PII name list is available
+        id: pii_config
         run: |
           if [[ -z "${PII_NAMES:-}" ]]; then
-            echo "::error::PII_NAMES secret is not set - the scanner cannot run."
-            echo "Add it under Settings -> Secrets and variables -> Actions."
-            exit 1
+            echo "::warning::PII_NAMES secret unavailable - customer-name scan SKIPPED for this run. Expected for fork PRs (GitHub does not pass secrets to them). Coverage backstop: the scan runs on same-repo branches (set the PII_NAMES secret under Settings -> Secrets and variables -> Actions) and on maintainer review before merge."
+            echo "configured=false" >> "$GITHUB_OUTPUT"
+          else
+            echo "configured=true" >> "$GITHUB_OUTPUT"
           fi
 
       - name: Check PR title
+        if: steps.pii_config.outputs.configured == 'true'
         env:
           PR_TITLE: ${{ github.event.pull_request.title }}
         run: |
@@ -42,6 +53,7 @@ jobs:
           printf '%s\n' "$PR_TITLE" | scripts/check-pii.sh
 
       - name: Check PR body
+        if: steps.pii_config.outputs.configured == 'true'
         env:
           PR_BODY: ${{ github.event.pull_request.body }}
         run: |
@@ -49,6 +61,7 @@ jobs:
           printf '%s\n' "${PR_BODY:-<empty>}" | scripts/check-pii.sh
 
       - name: Check commit messages
+        if: steps.pii_config.outputs.configured == 'true'
         env:
           BASE_SHA: ${{ github.event.pull_request.base.sha }}
           HEAD_SHA: ${{ github.event.pull_request.head.sha }}
@@ -57,6 +70,7 @@ jobs:
           git log --format='%H%n%s%n%b%n---' "$BASE_SHA..$HEAD_SHA" | scripts/check-pii.sh
 
       - name: Check changed file contents
+        if: steps.pii_config.outputs.configured == 'true'
         env:
           BASE_SHA: ${{ github.event.pull_request.base.sha }}
           HEAD_SHA: ${{ github.event.pull_request.head.sha }}
--- a/.gitignore
+++ b/.gitignore
@@ -61,3 +61,7 @@ android-jniLibs
 .mise.local.toml
 .direnv/
 __pycache__/
+
+# PII scanner: the real banned-name list lives here, never committed.
+# Seed it from .pii-names.local.example. See scripts/check-pii.sh.
+.pii-names.local
--- a/.pii-names.local.example
+++ b/.pii-names.local.example
@@ -1,0 +1,14 @@
+# PII name list for scripts/check-pii.sh (LOCAL - do not commit the real one).
+#
+# Copy this file to `.pii-names.local` and replace the placeholders with the
+# actual customer/tenant names that must never appear in public artifacts.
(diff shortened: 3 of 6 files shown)


jtomek-strike48:
Pushed `10bc704` to fix the red `Scan for customer names` check. Two things were compounding:

1. This PR *introduces* `pii-check.yml`, and its `Verify PII name list is configured` step hard-failed because the `PII_NAMES` secret isn't set yet.
2. More fundamentally, this is a **fork PR** (`head: jtomek-strike48`), and GitHub does not pass repository secrets to `pull_request` runs from forks. So a hard gate on `PII_NAMES` would red-x every external contribution permanently, not just this one -- a design flaw, not a missing value.

**Fix:** the `Verify` step is now a `Detect` step that sets a `configured` output. When the secret is present (same-repo PRs / pushes to `main`) the scan is **mandatory and hard-fails on a match, unchanged**. When it's absent (fork PRs) the four scan steps skip with a loud `::warning::` naming the fork-secret limitation and the backstops. The skip is never silent.

Verified on this run: `Detect` emitted the warning, all four scan steps show `skipped`, job is green. Locally confirmed `check-pii.sh` still fails loud (exit 1 on a banned name, 2 with no list) -- the scanner is unchanged; the workflow just guards the no-list path instead of tripping over it.

**Admin follow-up (not blocking this PR):** to get real coverage on same-repo branches, set the `PII_NAMES` repo secret (Settings -> Secrets and variables -> Actions), newline- or comma-separated. I have `maintain`, not `admin`, so I can't set it -- and it holds actual customer names, which shouldn't pass through me anyway.

ngdevo:
HIGH

  H1. Names are not regex-escaped when built into the pattern.
  scripts/check-pii.sh:build_pattern: names are concatenated verbatim into \b(name1|name2|...)\b. The old hardcoded list was all [a-z]+ and safe. The new external-source design accepts anything an
  operator drops into $PII_NAMES or .pii-names.local. If a name legitimately contains regex metacharacters — foo.bar, acme-corp (llc), tenant? — the pattern is either malformed (grep prints "Invalid
  extended regular expression") or matches too broadly (foo.bar matches fooxbar, fooybar, ...). This is a scanner-that-doesn't-scan risk exactly where the risk is highest. Fix: run each name
  through a regex-escape (sed 's/[][\\.*^$()+?{}|]/\\&/g') before joining.

  H2. Fork-PR "skip with warning" is exploitable in a public repo.
  .github/workflows/pii-check.yml:32-37: on fork PRs the scan is skipped with a ::warning::. In a public repo, anyone can fork, put a customer name in the PR title/body/commit message, and the PII
  check will not fail. The "same-repo re-run + maintainer review as the backstop" only kicks in if a maintainer notices, and by that point the names are already visible on github.com in the fork's
  commit history. The right mitigation is either (a) run the scan against the committed content of the base ref + the diff patch only (no secret needed to grep the patch text for names, if names are
  known via a separate source), or (b) require the check to pass via a maintainer-triggered rerun in the base repo before merge (the current wording implies this but there is no branch-protection
  rule tying it in).

  MEDIUM

  M1. Empty stdin passes silently (pre-existing, but now more consequential).
  main(): if [[ -n "$input" ]]; then scan "stdin" ... fi. Previously an empty stdin was benign because the hardcoded list guaranteed at least attempted scans on other inputs. Now that load_names
  runs unconditionally at startup and fails-loud on missing config, the empty-stdin branch means: config is loaded, no scan runs, exit 0. Any hook that expects stdin | check-pii.sh to always run a
  real check silently doesn't. Consider requiring at least one input source.

  M2. The scan is not applied to the diff on fork PRs OR the git history.
  The PR body explicitly defers history scrub to a coordinated force-push. Fine as scope. But the names are already in origin/main's history for Strike48-public/pick, publicly readable via git log
  and by every existing clone and by GitHub's mirror/archive. This PR only prevents new commits from adding them. That deserves a follow-up ticket flagged as a security remediation (not just
  "operator action") because the horse has already left the barn on the public repo.

  M3. Applicability to matrix and sibling repos.
  Answering the user's inline question: I grep'd the workspace and there are 136+ hits of these four names across the matrix workspace outside pick/. Notable clusters: infrastructure-config/ (test
  fixtures, terragrunt scripts, docs — some as constants like "mssp-talion-euw1", some as prose comments in regression tests), artemis/tests/matrix/chat/test_chat_smoke.py:118 (regression pin
  comment naming a customer), infra_api/lib/**/*.ex (multiple deepseas NP flicker regression notes). Matrix is visibility: INTERNAL, so the risk is lower than pick's PUBLIC — but "internal" still
  means accessible to any org member, contractors, and via source archives. Yes, apply the same scanner + secret pattern to matrix, and add it to any other repo that shows up in the grep
  (infrastructure-config especially). Note: some of these hits are legitimate references to region/tenant codenames used as k8s cluster names — those are technically identifiers in production infra,
  not customer names in the PII sense, and scrubbing them may be neither possible nor useful. A scan pattern for matrix probably needs a more nuanced allow-list.

  LOW

  L1. install-hooks.sh copies the example to .pii-names.local on install, seeding the scanner with the placeholder example-customer-one. That's a valid list for the scanner but it "protects
  nothing." A comment about "edit before this is meaningful" would be worth surfacing at install time.

  L2. readonly REPO_ROOT="$(cd ... && pwd)" resolves symlinks; if the repo is checked out under a symlinked path, the reported errors will show the resolved path (minor UX).

  L3. PII_NAMES and PII_NAMES_FILE are both consulted, but if the env var is empty-string vs. unset the code treats them as unset. Fine, just worth knowing.

cchao-devo (changes requested):
Review of pick PR #336 — security(pii): load banned-name list from outside the repo
  
  Verdict: REQUEST CHANGES — one blocker (regex escaping); solid otherwise.

  The core design is right and worth landing: externalizing the list, fail-loud exit 2 on an empty list, and dropping the name echo from public CI logs are all genuine improvements. I ran the
  scanner to verify behavior rather than reading it by eye, and found one blocking issue that defeats the control, plus two notes.

  Blocker — operator-supplied names are not regex-escaped (confirmed by running it)
  
  build_pattern concatenates names verbatim into grep -E "\b(n1|n2|...)\b". With the old hardcoded list this was safe (all names were plain [a-z]), but the whole point of this PR is that the
  list is now operator-supplied via $PII_NAMES / .pii-names.local — so a name containing an ERE metacharacter breaks it. Demonstrated on this branch:

  - Name foo(bar, input contains foo(bar here → grep: parentheses not balanced → script exits 0 (PASS) despite the banned name being present. Because the bad name corrupts the whole alternation,
  one malformed entry silently disables the entire scan — and since the failure is a nonzero grep handled as "no match," CI goes green while scanning nothing.
  - Name a.c, input the axc token → matches (exit 1). Unescaped . (and *, +, [, \, …) also cause false positives.

  This is exactly the HIGH already raised by @ngdevo; the runnable repro above confirms it's a silent no-op of a security control, not just a matching quirk. Fix: escape each name before
  joining, e.g.
  
  for name in "${NAMES[@]}"; do
    esc=$(printf '%s' "$name" | sed 's/[.[\*^$()+?{|\\]/\\&/g')
    pattern+="${sep}${esc}"; sep="|"
  done

  or sidestep regex entirely with fixed-string word matching (grep -Fiw -e "$name" per name). Please add a test with a metacharacter name so this can't regress.

  Note 1 — local hooks are toothless by default (Medium)

  install-hooks.sh seeds .pii-names.local from the placeholder example, so after install-hooks.sh the local pre-commit/commit-msg hooks scan only for example-customer-one/two — not the real
  names — yet report "hooks installed." Since the real list is secret, a dev can't populate it without out-of-band distribution, so local protection is effectively opt-in and CI (same-repo, with
  the PII_NAMES secret) is the only real gate. That's a defensible architecture, but consider making the placeholder-only state more obvious (e.g. the hook printing "scanning against
  placeholder names — set real names in .pii-names.local").

  Note 2 — fork-PR skip is reasonable and correctly loud

  Skipping the scan when PII_NAMES is empty (fork PRs, where GitHub withholds secrets) with a ::warning:: and relying on the same-repo re-run + maintainer review is the right call given the
  secret model — and it avoids the pull_request_target footgun. Accepted residual risk; no change needed. Worth confirming the same-repo/base branch run does fire before merge so the backstop is
  real.

  Verified

  - shellcheck -s bash scripts/check-pii.sh clean (exit 0).
  - .pii-names.local is gitignored (git check-ignore confirms).
  - Normal detection (exit 1), clean input, and no-list fail-loud (exit 2) all behave as described.
  - History-scrub follow-up is correctly called out as out-of-scope (names remain in git log).

jtomek-strike48:
Blocker fixed in `c854499`, plus Note 1. Verified empirically, not by reasoning.

**Blocker — operator names not regex-escaped**
Switched from building an ERE alternation to fixed-string whole-word matching: `grep -Fiw -e "$name"` per name. A name can no longer be interpreted as a pattern, so the control holds by construction rather than by escaping every metacharacter. Reproduced the old failure and confirmed the fix on this branch:
- Old path, name `foo(bar`: `grep -E '\b(foo(bar)\b'` errors ("mismatched ( )") -> nonzero -> the `if grep ...` caller treats it as "no match" -> silent PASS while scanning nothing. New path detects it (exit 1).
- Old path, name `a.c` vs input `axc`: `.` matched any char -> false positive (exit 1). New path: literal, no match (exit 0).

**Regression test + CI wiring**
Added `scripts/check-pii.test.sh` with metacharacter-name cases (unbalanced paren detected literally; one bad name does not disable the rest of the list; `.` does not false-positive; word-boundary and fail-loud preserved). Wired it into the PII Check workflow as a step that runs on every PR regardless of the `PII_NAMES` secret (forks included, since it supplies its own names), so this cannot silently regress. The test file is excluded from the added-lines diff scan since it intentionally contains name-like tokens.

**Note 1 — placeholder-only local hooks**
`install-hooks.sh` now warns when `.pii-names.local` is still identical to the committed example, making the placeholder-only state (and the fact that CI with the `PII_NAMES` secret is the authoritative gate) explicit.

**Note 2 — fork-PR skip**: accepted as-is; the same-repo run is the backstop.

shellcheck clean on all three scripts; the regression suite passes. Re-requesting review.

cchao-devo (approved):
Review of pick PR #336 (re-review @ c854499) — security(pii): load banned-name list from outside the repo

  Verdict: APPROVE — blocker resolved, ready to merge.

  Thanks for the fix. The regex-escaping blocker from my earlier review is fully addressed, and I re-verified it by running the scanner rather than reading the change. Both prior notes are
  handled too.

  Blocker — resolved and confirmed by running it

  Switching from an ERE alternation to fixed-string whole-word matching (grep -Fiw -e "$name" per name) is the right fix. Re-ran the exact cases that previously failed, on c854499:
  - Name foo(bar, input contains foo(bar here → now detected (exit 1), where it previously errored on the broken regex and silently passed.
  - Name a.c → no longer over-matches axc (exit 0), and matches the literal a.c (exit 1).
  - Word boundary preserved: name maidar does not match maidart (exit 0) — -w cleanly replaces the old \b(...)\b.
  - Baseline detection of a plain name still works (exit 1).

  Regression guard — strong, and runs everywhere

  scripts/check-pii.test.sh is hermetic (each case supplies its own $PII_NAMES) and wired as an un-gated workflow step, so it guards the fix on every PR including forks (where the secret is
  absent). It passes locally (exit 0), and I especially like the foo(bar\nacme case — it pins the sharpest failure mode (one malformed name must not disable matching of the others), which the
  old alternation got wrong. Excluding check-pii.test.sh (and the scanner) from the diff-content scan so its metacharacter fixtures don't trip the scan is the correct call.

  Notes — both handled

  - Note 1 (local placeholder assurance): install-hooks.sh now warns loudly when .pii-names.local is still identical to the example, making the placeholder-only (no real protection) state
  obvious. Good.
  - Note 2 (fork-PR skip): unchanged and still the right trade-off given the secret model.

  Verified

  - scripts/check-pii.test.sh passes (exit 0).
  - shellcheck -s bash clean on check-pii.sh, check-pii.test.sh, and install-hooks.sh.
  - The self-test step has no if: guard, so it executes on fork PRs too.

PR #343, opened 2026-07-29
feat(demo): Pick + DVWA docker-compose demo unit (matrix#3207 SK-101)

cchao-devo:
> **Status: Pick-repo half of a multi-repo effort (matrix#3207).** This adds the
> self-contained docker-compose demo unit; the init-dev orchestration
> (`task demo:pick:up`/`attach`, realm provisioning, tenant-UUID wiring) lands in a
> separate init-dev PR and *invokes* the recipes added here. Contents here are now
> settled and fix-reviewed (see "Review fixes applied" below).

## What this adds

A self-contained docker-compose unit that runs the **Pick** connector against **DVWA**
on an isolated network, for the StrikeKit sales/SE demo (matrix#3207).

- **`docker-compose.dvwa.yml`** - `pick` + `dvwa` on an isolated `scan-net` (internal
  bridge) so Pick reaches DVWA by service DNS (`http://dvwa`); `pick` also joins
  `backend-net` for egress to the Strike48 backend, and maps the backend hostnames
  (`connectors-studio`/`studio`/`auth.strike48.test`) to `host-gateway` so it can reach a
  local/newdev cluster. Pick is outbound-only (no inbound port); DVWA has no `ports:` and
  is not exposed to the host or beyond `scan-net`.
- **`just dvwa-up` / `dvwa-down` / `dvwa-check`** - bring-up (env-guarded), clean
  teardown, and a no-daemon config + isolation validation.
- **`.env.dvwa.example`** - env template (token-less OTT pending-approval flow; the
  security-relaxation flags default off and are opted-in only for newdev).
- **`.gitignore`** - ignores the generated `.env.dvwa` (carries the resolved tenant UUID
  + insecure-TLS flags; must never be committed).
- **`Dockerfile`** - adds `setcap cap_net_raw,cap_net_admin+ep` (+`libcap2-bin`) to the
  slim image's non-root binary. See divergence #1 - this is orthogonal to the demo.

## Auth / isolation model

- Token-less **OTT pending-approval**: Pick registers PENDING; an operator approves it in
  Studio -> Gateways. Credentials persist to a named volume so the approval survives
  restarts.
- The external demo Pick is intended to register into a dedicated **demo realm** (created
  by the init-dev side); it does not touch any existing connector.
- **Isolation invariant** (guard-tested by `dvwa-check`): `scan-net` is `internal: true`
  (no egress), DVWA has no host ports and is only on `scan-net`. These are hardcoded
  literals, not env-overridable, so no `.env` can silently re-expose the deliberately
  vulnerable target.

## Two intentional divergences

1. **`Dockerfile.scratch` (Kali), not the slim `Dockerfile`.** The compose builds
   `dockerfile: Dockerfile.scratch` - the same Kali image prod/k8s ships, which bundles the
   pentest toolchain (nmap, nikto, nuclei, ffuf, sqlmap, ...). The slim `Dockerfile` builds
   only the pentest-agent binary with **no external tools**, so scans would fail "tool not
   found". `Dockerfile.scratch` runs as **root** (HOME=/root, creds mounted at `/root`), so
   `NET_RAW`/`NET_ADMIN` from `cap_add` are effective without setcap. The `setcap` change to
   the slim `Dockerfile` is therefore **not used by this demo** (nor by CI/k8s, which also
   build `Dockerfile.scratch`) - it is included only to make the slim image's in-process
   libpcap path work under plain Docker, and is a no-op here. `DISABLE_SANDBOX=true` because
   the Kali image doesn't ship the proot sandbox rootfs; the demo's boundary is the
   container + internal `scan-net`, not the in-container sandbox.
2. **Naming / image, vs the matrix#3207 text.** Recipes are `dvwa-up`/`dvwa-down` (flat),
   not the issue's literal `pick:dvwa-up` - a colon in a `just` recipe name denotes a
   prerequisite dependency, not a namespace, so `pick:dvwa-up` is not a valid flat recipe.
   `dvwa-up`/`dvwa-down` satisfy the AC's intent (one-command up/down in the Pick repo), and
   the init-dev consumer invokes exactly these names. The DVWA image is
   `docker.io/vulnerables/web-dvwa` rather than the issue's `ghcr.io/digininja/dvwa`, which
   is **not anonymously pullable** (`pull denied`); `web-dvwa` is amd64-only (runs under
   emulation on arm64, fine for a demo target), and its healthcheck uses a PHP probe because
   the image ships no `curl`.

## Review fixes applied

A pre-push review panel (code-review / silent-failure / test-coverage / AC-traceability)
surfaced three issues, all fixed in `a73edd7` and verified:

- **`dvwa-down` could orphan the vulnerable DVWA.** When `.env.dvwa` was absent, the
  compose model's mandatory `${VAR:?}` interpolations failed and `down` aborted at
  parse-time, leaving DVWA (`restart: unless-stopped`) running. Fixed: `down` now defaults
  the required vars to stubs (it never uses their values), so teardown always succeeds even
  with no env file. Verified: `docker compose config` resolves with stubs and no env
  present, and fails without them (bug + fix both proven).
- **`dvwa-up` could report success on a failed registration.** The SDK retries on
  registration failure, so the container stays `running` while never appearing in Studio.
  The recipe no longer asserts "registers PENDING"; it tells the operator to confirm via
  `docker compose logs pick` and points at `STRIKE48_TENANT` / backend reachability if the
  connector never appears.
- **`.env.dvwa` was not gitignored** (it carries the tenant UUID + insecure-TLS flags).
  Fixed by adding it to `.gitignore`.

## Verification (run live on colima)

- Image builds; `docker compose config` topology correct (DVWA no host ports, `scan-net`
  `internal: true`, DVWA off `backend-net`); DVWA healthcheck goes healthy (PHP probe).
- Service DNS `pick -> dvwa` resolves on `scan-net`; the scan-net subnet + per-container
  IPs are documented for engagement scoping on the init-dev side.
- Full end-to-end run (with the init-dev side): Pick registers OTT pending-approval, is
  approved in Studio -> Gateways, and runs a real nmap scan against DVWA.
- `dvwa-down` tears down cleanly (incl. with no env file); `dvwa-check` guard-tested (fails
  if a future edit re-exposes DVWA via a host port or attaches it to `backend-net`).
- Local gates green: PII check, gitleaks. No Rust touched (cargo gates N/A).

## Known follow-up (optional, not blocking)

`dvwa-check` is a good isolation tripwire but is not wired into CI (no compose/just job),
and it asserts DVWA's isolation but not that `pick` stays outbound-only (no host ports). A
small `pull_request` job running `just dvwa-check` plus a `pick`-has-no-ports assertion
would turn it into an enforced gate.

Part of matrix#3207. The init-dev orchestration PR follows.


jtomek-strike48 (approved):
Reviewed empirically at head `a73edd7` in an isolated worktree, base `origin/main` `84d0b8e` (branch is 0 commits behind main). Infra-only diff, no Rust, so cargo gates are correctly N/A.

## Verified sound (ran it, not just read it)

- **Isolation guard is genuine (mutation-tested).** `just dvwa-check` passes clean. I broke each invariant and confirmed it goes RED, then restored GREEN:
  - Added `ports: ["8080:80"]` to DVWA -> `FAILED: dvwa publishes host ports` (exit 1)
  - Flipped `scan-net internal: true -> false` -> `FAILED: scan-net is not internal:true` (exit 1)
  - Attached DVWA to `backend-net` -> `FAILED: dvwa attached to backend-net` (exit 1)
- **The `dvwa-down` never-orphans fix reproduces and resolves.** Naive `docker compose down` with no env vars aborts at parse time (`required variable STRIKE48_HOST is missing`) and would orphan DVWA; `just dvwa-down` with no `.env.dvwa` and no exported vars tears down cleanly (exit 0).
- **`dvwa-up` env guard** errors cleanly with no env file (exit 1). **`docker compose config`** renders the correct topology (DVWA on scan-net only, no host ports, `scan-net internal:true`).
- **`.env.dvwa` is git-ignored, `.env.dvwa.example` is not** (`git check-ignore` confirmed both). No secrets or PII in the example (sentinels + `.test` hosts only).
- **Claim audit, all backed:** only `docker-multiarch.yml` builds an image and it uses `Dockerfile.scratch`, so the slim-`Dockerfile` setcap change is a genuine no-op for CI/k8s. `Dockerfile.scratch` runs as root and bundles the toolchain; `HOME=/root` confirmed via `docker run`. The setcap layer builds on `debian:bookworm-slim`, and `getcap` shows `cap_net_admin,cap_net_raw=ep` with `CapEff=0` for uid 999 without `cap_add` and effective with it, exactly matching the comment.
- **CI-scope is honest:** Pick has no compose/just/shellcheck CI gate, and the PR discloses `dvwa-check` has no green signal in CI as optional follow-up. shellcheck on the recipes shows only `{{...}}` just-template noise (SC1083) and one intentional single-quoted heredoc (SC2016), no real defects.

## Optional nits (LOW, non-blocking)

1. **`dvwa-check` (justfile:134-136):** `set -e` does not catch a failing `docker compose config` in the prefix-assignment form (`RENDERED=$(...) python3 ...`). In the realistic failure (malformed compose -> empty stdout -> `yaml.safe_load` None -> `TypeError` -> exit 1) the guard fails-closed anyway, so this is not a live bug. The only latent hole is the adversarial case where `config` exits nonzero but still prints valid isolated YAML to stdout. Real `docker compose` errors to stderr with empty stdout, so this is robustness polish. If you want it airtight, split the assignment from the invocation:
   ```bash
   RENDERED=$(STRIKE48_HOST=... docker compose -f "{{dvwa_compose}}" config)
   python3 -c '...' <<< "$RENDERED"
   ```
   Verified this propagates the failure (exit 9) before Python runs.

2. **`dvwa-up` (justfile:165-170):** the `sleep 2` + `ps --status running` check only catches crash-on-boot, not a crash at ~5-10s (e.g. late backend DNS/gRPC failure). This is already disclosed in the comment ("running is NOT proof it registered") and directs the operator to check logs + Studio, so it is a documented trade-off for a demo helper, not a masked failure. A short poll loop (~10s) would tighten it if desired.

Not reproduced locally: the PR body's full end-to-end run (Pick registers, approved in Studio, runs a real nmap scan) needs the init-dev backend, so that is author-verified-only, consistent with the disclosed Pick-repo/init-dev split.

Isolation invariant is mutation-verified, both headline bug fixes reproduce-and-resolve as claimed, every code and comment claim traces to code, no secrets, cleanly scoped as half of matrix#3207. Approving.

Issue #347, opened 2026-07-29
Agent-facing: inject live subnets + fact-first prompt rule + scope-aware errors


jtomek-strike48:
Child of #344 (PR 3 - Agent-facing behavior). Depends on the foundation child; independent of the target-pipeline child.

## Summary

Close the root cause directly: give the Red Team Agent its host network facts up front and tell it to use them, and make failed/empty scans self-explanatory so the model course-corrects even when it does guess.

## Evidence (current code)

- No host network context is injected. `RED_TEAM_SYSTEM_PROMPT` (`crates/ui/src/components/chat_panel/constants.rs:99`) never mentions interfaces/subnets; the agent-registration path (`crates/ui/src/liveview_connector/*`) injects tool schemas + instance metadata but no live interface/subnet data. The model must call a tool to learn its own subnets.
- Zero-host results are currently opaque - an empty ARP/nmap sweep returns no hint that the *target* was wrong vs. the network being empty.

## Why it matters

This is the piece that actually prevents the `192.168.1.0/24`-on-a-`10.0.x` failure. Prompt rules alone are soft (the model already "should" gather facts); pairing the rule with pre-injected facts removes the need to guess. Scope-aware errors turn a silent 0-host result into a teachable signal.

## Proposed approach

- Inject the active subnets (from `network_context()`, foundation child) into the agent's context at registration / session start.
- Add a fact-first rule to `RED_TEAM_SYSTEM_PROMPT` (`constants.rs:99`): before any CIDR scan, use the injected subnets or `target="auto"`/`"current"`; never invent a range.
- Enrich zero-host / degraded scan results with "you're on X, you scanned Y" - and, when raw-socket capability was unavailable in-sandbox, say so explicitly (see #251) rather than implying an empty network. Coordinate with #337 (host-down inference) and #309 (false-success on probes that never ran).

## Verification

- Deterministic parts (the enrichment string, context-injection payload) get unit tests, guard-checked.
- The behavioral claim ("agent gathers facts first") is verified **empirically**: run `./run-pentest.sh headless dev` and confirm the agent reads injected subnets instead of guessing a home range. This cannot be meaningfully unit-tested; the test plan will say what was run.

## Related

- Parent #344. Depends on foundation child.
- #251 (sandbox degradation wording), #337 (host-down inference), #309 (probe-never-ran false success), #325 (injected context feeds mode dispatch).

## Acceptance criteria

- [ ] Active subnets injected into agent context at session start.
[...]

Issue #346, opened 2026-07-29
Scan target pipeline: auto/current/all sentinels + validate the 5 bypass tools


jtomek-strike48:
Child of #344 (PR 2 - Target pipeline). Depends on the foundation child (network_context / prefix_len).

## Summary

Teach the discovery scanners to accept `target="auto"`/`"current"`/`"all"`, resolving to the host's real subnets before validation - and, in the same per-tool edit, close the validation gap where 5 tools shell the raw target with no checks. Both changes touch the same `execute()` region, so they ship together to avoid a self-inflicted merge conflict and an ordering bug.

## Evidence (current code)

- No CIDR-target sentinel exists anywhere. The only synonym-folding precedent is **ports-only**: `resolve_port_args` folds `"all"`/`"-"`/`"1-65535"` -> `-p-` at `crates/tools/src/external/nmap.rs:885`, before `validate_port_spec`. Mirror this shape for targets.
- Primary scanners already validate: nmap (`external/nmap.rs:139`), masscan (`external/masscan.rs:96`), rustscan (`external/rustscan.rs:95`), port_scan (`port_scan.rs:67`) all call `validate_target` (`crates/core/src/validation.rs:308`).
- **Full bypass (no validation at all)** - raw string shelled via `CommandBuilder.positional`/`.arg`: `arp_scan` (`external/network/arp_scan.rs:57`), `netdiscover` (`external/network/netdiscover.rs:71`), `masscan_fast` (`external/network/masscan_fast.rs:63`), `nmap_vuln` (`external/network/nmap_vuln.rs:57`), `responder`. Argv (not shell-string) execution limits injection risk, but there is no target-format or scope validation.
- Tool schema has **no `enum` slot** (`crates/core/src/tools.rs:436` `to_json_schema`) - accepted values are advertised only via the free-text param `description`.

## Why it matters

`auto`/`current` makes the correct target the default instead of relying on the model to remember to gather facts. Folding it in ahead of validation, and adding validation to the bypass tools in the same pass, keeps the pipeline order correct (**resolve -> validate -> warn**) and the files touched once.

## Proposed approach

- Add a shared `resolve_scan_target(raw) -> Vec<Subnet>` in `crates/core` (next to `validate_target` / `network_context` from the foundation child). `current` = primary/default-route subnet; `auto`/`all` = every active non-loopback subnet; anything else passes through unchanged.
[...]

Issue #344, opened 2026-07-29
Epic: Fact-first network discovery for the Red Team Agent


jtomek-strike48:
## Summary

The Red Team Agent can fall back to a hallucinated network range (e.g. `192.168.1.0/24`) instead of gathering host network facts before a CIDR scan. On a real multi-homed host (`10.0.8.0/22` + `10.0.40.0/24`) this produces a scan of the wrong network and zero results, with no signal that the target was fabricated.

Make "gather network facts first" the **default path**, not a discipline the model must remember.

## Root cause

The correct active subnet is **already computed** from the real interface netmask (`crates/tools/src/safety_check/network_map.rs:243`, `subnet_cidr_v4`, shipped in PR #182) - but it is thrown away:

- It is local to `discover_network()` and never returned in `NetworkMap` (`network_map.rs:105`).
- It is not reusable as a scan target (no `auto`/`current` sentinel on any scanner).
- It is not injected into agent context - the model must call a tool to learn its own subnets, and nothing pre-feeds them (`RED_TEAM_SYSTEM_PROMPT` at `crates/ui/src/components/chat_panel/constants.rs:99` never mentions interfaces/subnets).
- A parallel code path still hardcodes `/24` because the platform `NetworkInterface` struct has no netmask field (`crates/tools/src/autopwn/network_plan.rs:111`; struct at `crates/platform/src/traits.rs:279`).

So the fix is not "more rules for the model" - it is **exposing the source of truth that already exists** and making the safe target the default.

## Decomposition

Split by merge-conflict seam + verification method (not by conceptual tier), because the sentinel-folding and target-validation changes edit the same `execute()` region of the same tools and must not land in conflicting branches.

- [ ] #345 - **Foundation** (plumbing, no agent-visible change): `prefix_len` on `NetworkInterface`; cheap `network_context()` accessor (no scan); thread derived CIDR into `SafetyCheckResult`; remove the `network_plan.rs:111` `/24` hardcode.
- [ ] #346 - **Target pipeline** (one coherent per-tool pass): shared `resolve_scan_target()`; wire discovery scanners in fixed order **resolve sentinel -> validate_target -> non-blocking out-of-subnet warning**. Delivers `auto`/`current`/`all` AND closes the 5-tool validation gap in one edit.
[...]

PR #348, opened 2026-07-29
feat(network): expose scan-free host subnets as reusable context (#345)

### LINKED ISSUE #345: Network foundation: expose netmask-derived subnets as reusable context
Child of #344 (PR 1 - Foundation). Pure plumbing; no agent-visible behavior change.

## Summary

Expose the active-subnet source of truth that already exists internally so it can be reused as a scan target (#344 PR 2) and injected into agent context (#344 PR 3). Add the missing netmask field to the platform interface struct and remove the hardcoded `/24` it forced.

## Evidence (current code)

- `subnet_cidr_v4()` already derives the correct subnet from the real netmask - `crates/tools/src/safety_check/network_map.rs:243` (shipped in PR #182) - but it runs only inside `discover_network()` (which fires an nmap sweep) and is not returned in `NetworkMap` (`network_map.rs:105`).
- `NetworkInterface` has `name`/`ip_addresses`/`mac_address`/`is_up`/`is_loopback` but **no netmask/prefix** - `crates/platform/src/traits.rs:279`.
- Because of that gap, a second path hardcodes `/24`: `crates/tools/src/autopwn/network_plan.rs:111` (comment: "default to /24 since netmask not available").
- Desktop enumeration: `crates/platform/src/desktop/system.rs:42` (+ `get_network_interfaces_fallback()` `ip addr` parser at `:77`). Android delegates to it.

## Why it matters

Every downstream fix in #344 needs a cheap, scan-free way to answer "what subnets is this host actually on?". Today that answer requires running a network sweep (`discover_network`) or is fabricated (`/24`). This issue makes it a plain accessor.

## Proposed approach

- Model prefix per-address (netmask is per-address in network-interface v2.0.5, and one interface can hold multiple IPs): replace `ip_addresses: Vec<String>` with `addresses: Vec<InterfaceAddr { ip, prefix_len }>` on `NetworkInterface`; populate it in the desktop `network-interface` path, the `ip addr` fallback parser, and Android. Fail loud / `None` where genuinely unavailable (align with #202).
- Add a `network_context()` accessor (in `crates/core`, next to target validation) that enumerates interfaces and derives each subnet from its netmask **without running a scan** - factor the pure math out of `subnet_cidr_v4` so it is shared, not duplicated.
- Thread the derived CIDR(s) into `SafetyCheckResult` / `NetworkMap` so callers can read them.
- Remove the `/24` hardcode at `network_plan.rs:111`; use the real prefix.

## Related

- Parent #344. Blocks #344 PR 2 and PR 3.
- #202 - fail loudly when platform-capability fallbacks are compiled out (network interfaces).

## Acceptance criteria

- [ ] `NetworkInterface.addresses[].prefix_len` populated on desktop (both paths) and Android; `None` only when truly unavailable.
- [ ] `network_context()` returns active subnets without running a scan; subnet math is shared with `subnet_cidr_v4` (no duplication).
- [ ] Derived CIDR(s) exposed on `SafetyCheckResult`/`NetworkMap`.
- [ ] `network_plan.rs:111` `/24` hardcode removed.
- [ ] Unit tests for prefix parsing and subnet derivation; **guard-checked** (neuter the derivation -> test goes red).

---

**Design update (2026-07-29, verified during PR 1):** network-interface v2.0.5 exposes netmask *per-address* (`Addr::netmask() -> Option<IpAddr>`), and `NetworkInterface.ip_addresses` is a `Vec` (multi-homed / dual-stack). A single interface-level `prefix_len` scalar is therefore lossy for exactly the multi-homed case this epic targets. Chosen model: structured per-address `addresses: Vec<InterfaceAddr { ip: String, prefix_len: Option<u8> }>`. Struct is never serialized across a wire boundary (verified: 3 in-process read sites, 4 construction sites), so the shape change is safe.

jtomek-strike48:
Part of #344 (fact-first network discovery). This is **PR 1 - Foundation**: pure plumbing, no intended agent-visible behavior change. Closes #345.

## Summary

- Makes the host's active subnets a **reusable, scan-free source of truth** (`network_context()` in `crates/tools/src/network_context.rs`) so later PRs can resolve `auto`/`current` scan targets (#346) and seed the agent with real subnets (#347) instead of relying on the model to guess a range.
- Models interface addresses **per-address**: `NetworkInterface.ip_addresses: Vec<String>` becomes `addresses: Vec<InterfaceAddr { ip, prefix_len }>`. `network-interface` v2.0.5 exposes netmask per-address, and one interface can be multi-homed / dual-stack, so a per-interface prefix scalar would be lossy for exactly the case this epic targets. (Deviates from #345's original AC, which is updated with the rationale.)
- Populates the real prefix in all three enumeration paths (desktop `network-interface` primary, desktop `ip addr` fallback, Android).
- Relocates the pure `subnet_cidr_v4` netmask math into `network_context` as the single source of truth; `safety_check/network_map` now imports it (removes a duplicate).
- Removes the hardcoded `/24` in `autopwn/network_plan.rs:111` (derives from the real prefix; the fallback now also emits a valid CIDR base rather than a malformed `x.y.z/24`).
- Exposes `active_subnets` on `SafetyCheckResult`, derived scan-free and independent of whether the discovery sweep succeeds.

## Verification

- `cargo check --workspace` clean; `cargo clippy -p pentest-tools -p pentest-platform --all-targets -- -D warnings` clean; `cargo fmt --all` applied.
- `cargo test -p pentest-platform -p pentest-tools --lib` -> 58 + 343 pass, 0 fail. 15 new tests.
- **Mutation-verified** the new guards go red when the code is broken: (A) `subnet_cidr_v4` with masking removed -> 6 tests red; (B) `prefix_len_from_ipv4_netmask` trusting popcount (dropping the non-contiguous-mask rejection) -> red; (C) `interface_addr_from_token` dropping the prefix -> red.
- Blast radius: `NetworkInterface` shape change touches 4 construction + 3 read sites (all updated); verified the struct is never serialized across a wire boundary or persisted (no json/bincode/toml/cache/fixture consumers), so the field rename is safe.

## Reviewer note (CI-scope, deliberate)

A pick-reviewer pass flagged that the macOS CI lane (`ci.yml:90`, `--package pentest-platform --package pentest-core --lib`) does not run `pentest-tools`, so the new `network_context` tests do not execute on macOS. This is **intentional and low-risk**: those tests are platform-invariant pure math and **do** run in the required Linux `--workspace` lane (`ci.yml:74`); the platform-specific helper (`prefix_len_from_ipv4_netmask`) lives in `pentest-platform` and runs on both lanes. Not adding `--package pentest-tools` to macOS (would bloat the lane for zero macOS-specific coverage). If PR 2's shared resolver naturally lands in `pentest-core`, the pure math can migrate then.

## Test plan

- [x] `cargo test -p pentest-platform -p pentest-tools --lib` green
- [x] `cargo clippy --all-targets -- -D warnings` clean
- [x] Mutation-verified the new guards (3 mutants confirmed red)
- [ ] CI green on the required Linux lane

### DIFF
--- a/crates/platform/src/android/system.rs
+++ b/crates/platform/src/android/system.rs
@@ -202,11 +202,11 @@ pub async fn get_network_interfaces() -> Result<Vec<NetworkInterface>> {
                 let is_loopback = name == "lo";
 
                 // Try to get IP address using ip command
-                let ip_addresses = get_interface_ips(&name).await;
+                let addresses = get_interface_addrs(&name).await;
 
                 interfaces.push(NetworkInterface {
                     name,
-                    ip_addresses,
+                    addresses,
                     mac_address: None, // Would need to read from /sys/class/net/*/address
                     is_up: true,
                     is_loopback,
@@ -218,8 +218,8 @@ pub async fn get_network_interfaces() -> Result<Vec<NetworkInterface>> {
     Ok(interfaces)
 }
 
-async fn get_interface_ips(interface: &str) -> Vec<String> {
-    let mut ips = Vec::new();
+async fn get_interface_addrs(interface: &str) -> Vec<InterfaceAddr> {
+    let mut addrs = Vec::new();
 
     if let Ok(output) = tokio::process::Command::new("ip")
         .args(["addr", "show", interface])
@@ -229,17 +229,16 @@ async fn get_interface_ips(interface: &str) -> Vec<String> {
         let stdout = String::from_utf8_lossy(&output.stdout);
         for line in stdout.lines() {
             let line = line.trim();
-            if line.starts_with("inet ") {
-                if let Some(addr) = line.split_whitespace().nth(1) {
-                    // Remove CIDR notation
-                    let ip = addr.split('/').next().unwrap_or(addr);
-                    ips.push(ip.to_string());
+            if line.starts_with("inet ") || line.starts_with("inet6 ") {
+                if let Some(token) = line.split_whitespace().nth(1) {
+                    // Keep the CIDR prefix (e.g. "/22") instead of discarding it.
+                    addrs.push(interface_addr_from_token(token));
                 }
             }
         }
     }
 
-    ips
+    addrs
 }
 
 #[cfg(test)]
--- a/crates/platform/src/desktop/system.rs
+++ b/crates/platform/src/desktop/system.rs
@@ -49,15 +49,24 @@ pub async fn get_network_interfaces() -> Result<Vec<NetworkInterface>> {
         Ok(interfaces
             .into_iter()
             .map(|iface| {
-                let ip_addresses: Vec<String> = iface
+                let addresses: Vec<InterfaceAddr> = iface
                     .addr
                     .into_iter()
-                    .map(|addr| addr.ip().to_string())
+                    .map(|addr| {
+                        // netmask is per-address; convert an IPv4 mask to a CIDR
+                        // prefix. IPv6 masks (and a missing/non-contiguous v4
+                        // mask) yield None rather than a fabricated prefix.
+                        let prefix_len = match addr.netmask() {
+                            Some(std::net::IpAddr::V4(mask)) => prefix_len_from_ipv4_netmask(mask),
+                            _ => None,
+                        };
+                        InterfaceAddr::new(addr.ip().to_string(), prefix_len)
+                    })
                     .collect();
 
                 NetworkInterface {
                     name: iface.name,
-                    ip_addresses,
+                    addresses,
                     mac_address: iface.mac_addr,
                     is_up: true,        // network-interface doesn't provide this
                     is_loopback: false, // would need to check IP
@@ -104,7 +113,7 @@ async fn get_network_interfaces_fallback() -> Result<Vec<NetworkInterface>> {
 
                     current_iface = Some(NetworkInterface {
                         name,
-                        ip_addresses: Vec::new(),
+                        addresses: Vec::new(),
                         mac_address: None,
                         is_up,
                         is_loopback,
@@ -112,18 +121,13 @@ async fn get_network_interfaces_fallback() -> Result<Vec<NetworkInterface>> {
                 }
             } else if let Some(ref mut iface) = current_iface {
                 let trimmed = line.trim();
-                if trimmed.starts_with("inet ") {
+                if trimmed.starts_with("inet ") || trimmed.starts_with("inet6 ") {
                     let parts: Vec<&str> = trimmed.split_whitespace().collect();
                     if parts.len() >= 2 {
-                        // Extract IP without subnet mask
-                        let ip = parts[1].split('/').next().unwrap_or(parts[1]);
-                        iface.ip_addresses.push(ip.to_string());
-                    }
-                } else if trimmed.starts_with("inet6 ") {
-                    let parts: Vec<&str> = trimmed.split_whitespace().collect();
-                    if parts.len() >= 2 {
-                        let ip = parts[1].split('/').next().unwrap_or(parts[1]);
-                        iface.ip_addresses.push(ip.to_string());
+                        // `ip addr` prints the address with its prefix, e.g.
+                        // "inet 10.0.8.42/22" - keep the prefix instead of
+                        // discarding it (the /24 assumption this PR removes).
+                        iface.addresses.push(interface_addr_from_token(parts[1]));
                     }
                 } else if trimmed.starts_with("link/ether ") {
                     let parts: Vec<&str> = trimmed.split_whitespace().collect();
@@ -478,7 +482,7 @@ pub async fn check_wifi_connection_status(
     // Find active WiFi interfaces (up + has IP)
     let active_wifi: Vec<_> = interfaces
         .iter()
-        .filter(|i| i.is_up && !i.ip_addresses.is_empty() && is_wifi_name(&i.name))
+        .filter(|i| i.is_up && i.has_addresses() && is_wifi_name(&i.name))
         .collect();
 
     // Find all WiFi interfaces (even if down)
--- a/crates/platform/src/traits.rs
+++ b/crates/platform/src/traits.rs
@@ -4,6 +4,7 @@ use async_trait::async_trait;
 use pentest_core::error::Result;
 use serde::{Deserialize, Serialize};
 use std::collections::HashMap;
+use std::net::IpAddr;
 use std::time::Duration;
 
 /// Combined platform provider trait
@@ -307,14 +308,100 @@ pub struct DeviceInfo {
     pub platform_specific: PlatformDetails,
 }
 
+/// A single address bound to a network interface, with its prefix length.
+///
+/// Modeled per-address (not per-interface) because a single interface can carry
+/// several addresses on different prefixes - a dual-stack NIC (IPv4 + IPv6) or a
+/// host with a secondary v4 on another subnet. A per-interface scalar prefix
+/// would be ambiguous for exactly the multi-homed case network discovery needs
+/// to get right, so the prefix travels with the address it describes.
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+pub struct InterfaceAddr {
+    /// The IP address, without any CIDR suffix (e.g. "10.0.8.42").
+    pub ip: String,
+    /// CIDR prefix length for this address (e.g. `Some(22)` for a /22), or
+    /// `None` when the backend could not determine it.
+    pub prefix_len: Option<u8>,
+}
+
+impl InterfaceAddr {
+    /// Construct from an IP string and an optional prefix.
+    pub fn new(ip: impl Into<String>, prefix_len: Option<u8>) -> Self {
+        Self {
+            ip: ip.into(),
+            prefix_len,
+        }
+    }
+
+    /// Parse the address into an [`IpAddr`], dropping any accidental CIDR
+    /// suffix. Returns `None` for a malformed address.
+    pub fn parse_ip(&self) -> Option<IpAddr> {
+        let bare = self.ip.split('/').next().unwrap_or(&self.ip);
+        bare.parse().ok()
+    }
+}
+
 /// Network interface information
 #[derive(Debug, Clone, Serialize, Deserialize)]
 pub struct NetworkInterface {
     pub name: String,
-    pub ip_addresses: Vec<String>,
+    /// Addresses bound to this interface, each carrying its own prefix length.
+    pub addresses: Vec<InterfaceAddr>,
     pub mac_address: Option<String>,
     pub is_up: bool,
     pub is_loopback: bool,
+}
+
+impl NetworkInterface {
+    /// The interface's IP addresses as bare strings (no prefix), preserving the
+    /// pre-`InterfaceAddr` ergonomics for the many call sites that only need the
+    /// address, not its prefix.
+    pub fn ip_strings(&self) -> Vec<String> {
+        self.addresses.iter().map(|a| a.ip.clone()).collect()
+    }
+
+    /// True if the interface has at least one bound address.
+    pub fn has_addresses(&self) -> bool {
+        !self.addresses.is_empty()
+    }
+}
+
+/// Parse an `ip addr`-style address token (e.g. `"10.0.8.42/22"` or a bare
+/// `"10.0.8.42"`) into an [`InterfaceAddr`].
+///
+/// The prefix is taken verbatim from the `/N` suffix when present and parseable;
+/// a missing or malformed suffix yields `prefix_len == None` rather than a
+/// guessed default. Shared by the Linux `ip addr` fallback parsers (desktop and
+/// Android) so the token handling stays consistent and unit-testable.
+pub fn interface_addr_from_token(token: &str) -> InterfaceAddr {
+    let mut parts = token.splitn(2, '/');
+    let ip = parts.next().unwrap_or(token).to_string();
+    let prefix_len = parts.next().and_then(|p| p.parse::<u8>().ok());
+    InterfaceAddr { ip, prefix_len }
+}
+
+/// Convert an IPv4 netmask (e.g. `255.255.252.0`) to a CIDR prefix length
+/// (e.g. `22`).
+///
+/// Returns `None` for a non-contiguous mask (bits not left-packed, e.g.
+/// `255.0.255.0`), which is not a valid CIDR netmask and must not be silently
+/// coerced into a plausible-looking prefix. Kept free of I/O so the conversion
+/// is unit-testable.
+pub fn prefix_len_from_ipv4_netmask(mask: std::net::Ipv4Addr) -> Option<u8> {
+    let bits = u32::from(mask);
+    let ones = bits.count_ones();
+    // A valid netmask is `ones` contiguous 1s followed by zeros. Reconstruct that
+    // canonical mask and compare; a mismatch means the mask was non-contiguous.
+    let canonical = if ones == 0 {
+        0
+    } else {
+        u32::MAX << (32 - ones)
+    };
+    if bits == canonical {
+        Some(ones as u8)
+    } else {
+        None
+    }
 }
 
 /// WiFi network information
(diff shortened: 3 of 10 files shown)


jtomek-strike48:
## Self-review (pre-merge) - PR #348 (network-context foundation, closes #345)

Cannot formally `Approve` my own PR, so recording the review as a comment. Another maintainer's approval is still required to clear the BLOCKED merge state.

**Method:** verified empirically, not by reasoning about the diff. Ran the three claimed mutants against a warm target dir on the PR head (`751f434`), backing up the two target files outside git and restoring them byte-identical afterward (both confirmed clean; the uncommitted follow-up work in the tree was never touched).

### Verdict: Sound. No blocking defects.

### Lens 1 - Mutation testing CONFIRMED (the claim not taken on trust)

The PR body claims "mutation-verified, 3 mutants go red." Reproduced all three; each matches exactly:

| Mutant | Change | Result |
|---|---|---|
| A | `subnet_cidr_v4`: `u32::from(addr) & mask` -> drop the mask | **6 tests red** (matches claim) |
| B | `prefix_len_from_ipv4_netmask`: skip the contiguity check, trust popcount | **red** - `netmask_to_prefix_rejects_noncontiguous_mask` (`255.0.255.0` -> `Some(16)` instead of `None`) |
| C | `interface_addr_from_token`: drop the parsed prefix | **red** - `interface_addr_from_token_keeps_prefix` (`10.0.8.42/22` -> `None` instead of `Some(22)`) |

The new tests genuinely guard the fix - not "green but non-guarding."

### Verified sound (read + empirical)
- **Bit math is correct.** `prefix_len_from_ipv4_netmask` handles the `/0` special case (avoids shift-by-32), rejects non-contiguous masks via canonical reconstruction, `count_ones() as u8` cast is safe. `subnet_cidr_v4` clamps wide prefixes to /16 and >32 to 32. Boundary shapes (`/0`, `/32`, IPv6, empty token, malformed suffix) handled gracefully (Lens 7).
- **Blast radius as claimed (Lens 3/5).** `ip_addresses` -> `addresses: Vec<InterfaceAddr>` touches exactly the construction + read sites listed, all updated. Types derive `Serialize/Deserialize` but no serialization call site consumes them (in-process only) - field rename safe. `SafetyCheckResult` gains `active_subnets` additively with `#[serde(default)]` (backward-compatible).
- **CI-scope note is honest (Lens 2).** network_context tests are in `pentest-tools` (Linux `--workspace` only, 9 tests via `--list`); the platform helpers are in `pentest-platform` (both lanes). macOS gap disclosed and justified. Not a finding.
- **All 5 ACs met**, all 21 CI checks green (macOS included), no PII/secrets/agent-notes in the diff.

### SHOULD-FIX (one, non-blocking)
- **`autopwn/network_plan.rs:119` - the `/24` fallback is now silent.** Happy path derives the CIDR from the real prefix, but when `prefix_len` is `None` or the address isn't IPv4 it still falls back to a guessed `/24` with **no log** - the exact behavior this epic exists to eliminate, now reachable silently (e.g. Android where the prefix may be unavailable). Add a `tracing::warn!` on that branch. Minor claim nuance (Lens 3): #345 AC#4 says the `/24` was "removed"; it is actually happy-path-only.

### Notes (non-blocking, not this PR)
- Android `get_interface_addrs` swallow (`if let Ok(output) = Command::new("ip")...` -> empty Vec on failure) is **pre-existing** (identical on `origin/main`). Follow-up under #202, not introduced here.
- `network_context()` returning empty on enumeration failure is documented best-effort and logs a warning. PR 2/PR 3 (#346/#347) will want to distinguish "no interfaces" from "enumeration failed" when resolving `auto`/`current` targets - flagging forward.
- Branch is current with `origin/main`; no rebase needed.

cchao-devo (approved):
Review: Pick PR #348 — expose scan-free host subnets as reusable context (#345)
  
  Verdict: ✅ Approve with one SHOULD-FIX (the author already flagged a weaker version of it; the panel + my verification found it's sharper). Clean foundation PR — the type design
  and the security-relevant subnet math are solid and well-tested. The one real gap is a silent, now-more-reachable malformed-CIDR fallback in the un-refactored sibling
  (network_plan.rs). Not a merge-blocker for a "Foundation / no agent-visible behavior change" PR, but worth fixing since it's in exactly the path the epic exists to eliminate.

  Lenses run (per the review-panel rule): pr-review-toolkit plugin agents — type-design-analyzer (struct-shape change + relocated shared helper, both required triggers),
  silent-failure-hunter, pr-test-analyzer — spawned in the background; plus my own mechanical gates and independent source/empirical verification of every finding. Verified in
  pick_08/pick.

  Mechanical gates (mine, all green)
  
  - cargo fmt --all --check clean · cargo clippy --workspace --all-targets -D warnings clean · pentest-tools 343/0, pentest-platform 56/0 (6 ignored). The one evidence_producer
  failure under parallelism is the known pre-existing global-singleton flake (per my memory) — this PR touches no evidence code, and it passes in isolation. Not a regression.

  Type design — clean (verified: cargo check --workspace compiles + I confirmed the caller sweep)

  - NetworkInterface.ip_addresses: Vec<String> → addresses: Vec<InterfaceAddr {ip, prefix_len: Option<u8>}> — every construction (3 prod + 3 test) and read site (5 prod) migrated
  correctly; sites needing only the IP use the parse_ip()/has_addresses() shim, sites needing the prefix (network_context, network_plan) read the full addr. Per-address modeling is
  right for dual-stack/multi-homed.
  - subnet_cidr_v4 relocation is byte-for-byte identical (the #182/#299 /16-clamp scan-scope math I reviewed before) — no duplicate remains, both callers use the relocated one,
  only the log wording differs. Good de-duplication.
  - active_subnets additive + #[serde(default)]; InterfaceAddr/NetworkInterface confirmed never serialized across a wire/persistence boundary, so the field rename can't break a
  stored consumer.
  - Prefix is Option<u8> and never guessed — missing/malformed/non-contiguous-mask → None, not a fabricated default. prefix_len_from_ipv4_netmask correctly rejects non-contiguous
  masks (the 255.0.255.0-shares-/16-popcount trap). This is the epic's core discipline, done right in the new module.

  🟠 SHOULD-FIX (not blocking) — silent /24 fallback in network_plan.rs can emit a malformed CIDR, and this diff raised its reachability

  The autopwn planner's CIDR derivation (network_plan.rs:116-124): happy path uses subnet_cidr_v4; the _ => arm falls back to format!("{}.0/24", …rfind('.')…) with no 
  tracing::warn!. Verified by running it:
  - v4 + no prefix → "10.0.8.0/24" (silent guessed /24 — the exact thing #344 exists to kill)
  - IPv6 primary → "fe80::1.0/24" / "2001:db8::5.0/24" — a garbage CIDR fed into the attack plan → nmap. determine_subnet in network_map.rs rejects IPv6; this sibling fabricates.

  This PR increased reachability (confirmed at source): Android enumeration changed from inet -only to inet -or-inet6 , and network_plan takes .addresses.first() (any family) — so
  an IPv6-first address list now reaches the malformed-CIDR arm where it couldn't before. The author's self-review flagged the silent-/24 half; the panel + my run found the
  malformed-CIDR + reachability half is worse.

  Fix (small): in network_plan.rs, (1) tracing::warn! on the fallback naming the interface + why the prefix was unavailable (matching the module's own clamp-logging observability),
  and (2) select the first IPv4 address (as subnets_from_interfaces correctly does) or return Error::InvalidParams rather than string-slicing an arbitrary-family address. That
  eliminates the garbage-CIDR case entirely.

  Severity framing: the guessed-/24 shape is pre-existing (old code produced its own malformed 192.168.1/24), the PR is explicitly "Foundation, no intended agent-visible change,"
  This PR increased reachability (confirmed at source): Android enumeration changed from inet -only to inet -or-inet6 , and network_plan takes .addresses.first() (any family) —so
  an IPv6-first address list now reaches the malformed-CIDR arm where it couldn't before. The author's self-review flagged the silent-/24 half; the panel + my run found the
  malformed-CIDR + reachability half is worse.

  Fix (small): in network_plan.rs, (1) tracing::warn! on the fallback naming the interface + why the prefix was unavailable (matching the module's own clamp-logging
  observability), and (2) select the first IPv4 address (as subnets_from_interfaces correctly does) or return Error::InvalidParams rather than string-slicing an arbitrary-family
  address. That eliminates the garbage-CIDR case entirely.

  Severity framing: the guessed-/24 shape is pre-existing (old code produced its own malformed 192.168.1/24), the PR is explicitly "Foundation, no intended agent-visible change,
  This PR increased reachability (confirmed at source): Android enumeration changed from inet -only to inet -or-inet6 , and network_plan takes .addresses.first() (any family) —
  so an IPv6-first address list now reaches the malformed-CIDR arm where it couldn't before. The author's self-review flagged the silent-/24 half; the panel + my run found the
  malformed-CIDR + reachability half is worse.

  Fix (small): in network_plan.rs, (1) tracing::warn! on the fallback naming the interface + why the prefix was unavailable (matching the module's own clamp-logging
  malformed-CIDR + reachability half is worse.

  Fix (small): in network_plan.rs, (1) tracing::warn! on the fallback naming the interface + why the prefix was unavailable (matching the module's own clamp-logging
  This PR increased reachability (confirmed at source): Android enumeration changed from inet -only to inet -or-inet6 , and network_plan takes .addresses.first() (any family)
  — so an IPv6-first address list now reaches the malformed-CIDR arm where it couldn't before. The author's self-review flagged the silent-/24 half; the panel + my run found
  the malformed-CIDR + reachability half is worse.

  This PR increased reachability (confirmed at source): Android enumeration changed from inet -only to inet -or-inet6 , and network_plan takes .addresses.first() (any family)
  an IPv6-first address list now reaches the malformed-CIDR arm where it couldn't before. The author's self-review flagged the silent-/24 half; the panel + my run found the
  malformed-CIDR + reachability half is worse.

  Fix (small): in network_plan.rs, (1) tracing::warn! on the fallback naming the interface + why the prefix was unavailable (matching the module's own clamp-logging observability),
  and (2) select the first IPv4 address (as subnets_from_interfaces correctly does) or return Error::InvalidParams rather than string-slicing an arbitrary-family address. That
  eliminates the garbage-CIDR case entirely.

  Severity framing: the guessed-/24 shape is pre-existing (old code produced its own malformed 192.168.1/24), the PR is explicitly "Foundation, no intended agent-visible change,"
  and network_plan is the un-refactored sibling (the epic's later PRs #346/#347 do the target resolution). So SHOULD-FIX, not block — but it's the one behavioral hole in the exact
  path the epic targets.

  🟡 Notes (non-blocking, forward-looking)

  - active_subnets computed but not surfaced to the agent (type-design finding, verified): run_safety_check sets + logs it, but safety_check_tool.rs's structured_data JSON doesn't
  include it — so the AI consumer doesn't see it in this PR. That's consistent with the stated scope (#347 is the "seed the agent with real subnets" PR), but the active_subnets
  doc-comment shouldn't imply it's agent-visible now. Just verify the wording.
  - active_subnets: Vec<String> conflates "enumeration failed" vs "genuinely no subnets" (silent-failure finding): logging distinguishes them, but the stored empty Vec doesn't, and
  no consumer reads it yet. The PR itself anticipates PR2/PR3 needing the distinction — worth changing to Option<Vec<String>> (or a small enum) before a consumer lands, so a
  failed enumeration doesn't read as "host on no networks." Latent, file against #346/#347.
  - Test gaps (pr-test-analyzer): the pure math is excellently covered (DAMP names, non-contiguous-mask guard, /16 clamp, mutation-verified). Uncovered: the network_plan.rs
  fallback branch (GAP A — no test pins the /24 or the IPv6-garbage case; adding one would catch the SHOULD-FIX), the netmask()→prefix_len→InterfaceAddr wiring in desktop
  enumeration, and the dual-stack-single-interface subnets_from_interfaces path. None block, but a test on the network_plan fallback should ride with the fix.
  - CI-scope note is honest — the new network_context tests are pentest-tools (Linux --workspace lane only, not macOS), platform-invariant pure math, and the platform helper runs
  on both lanes. Reasonable.

ngdevo:
HIGH

  H1. NetworkInterface is Serialize/Deserialize — the "no persistence" claim is unverified in this diff.
  crates/platform/src/traits.rs:314: NetworkInterface derives both. The PR body says "verified the struct is never serialized across a wire boundary or persisted (no json/bincode/toml/cache/fixture
  consumers)." I can't independently verify this from the diff — a grep NetworkInterface | grep -iE 'to_(json|string)|serde_json' on the full source would settle it. If the PR author has run that
  grep, this is fine. If not, the ip_addresses → addresses rename is a hard-break for any consumer holding a cached copy or a legacy fixture. Worth confirming before merge (grep -rn "ip_addresses"
  crates/ apps/).

  MEDIUM

  M1. MIN_SUBNET_PREFIX = 16 silently narrows real /12 and /8 subnets.
  crates/tools/src/network_context.rs:59: the cap is documented in a comment, and the info-level tracing::info! says "clamping." But when an operator legitimately runs on a /12 network (RFC 1918 is
  a /8 as declared, though carved into smaller subnets in practice) the derived scan target silently becomes a /16. Two failure modes: (a) the actual hosts on the operator's /12 outside the clamped
  /16 are silently excluded from auto/current; (b) the info-level log is easy to miss vs. warn-level. Consider: return the cap in the tool result (visible to the LLM) or warn instead of info.

  M2. autopwn/network_plan.rs IPv6 fallback produces malformed CIDR.
  autopwn/network_plan.rs:111-117 fallback: format!("{}.0/24", &local_ip[..local_ip.rfind('.').unwrap_or(local_ip.len())]). When the address is IPv6 (no .), rfind('.') is None → slice becomes the
  full v6 string → result is <v6>.0/24, e.g. fe80::1.0/24. That's meaningless. Cleaner: reject or error on non-v4 in the fallback branch. Low-severity because the fallback is only hit when
  parse_ip() fails or prefix_len is None, which itself only happens when we couldn't derive anything usable.

  LOW

  L1. prefix_len_from_ipv4_netmask correctly rejects non-contiguous masks — this is a genuinely good defensive detail and the test explicitly names the popcount-collision case. Good.

  L2. subnets_from_interfaces dedup: dedup by CIDR string preserves the first interface's name in Subnet.interface even if the primary interface is seen later. Diagnostic-only, but the interface
  field on the result could show the wrong interface for a subnet claimed by two interfaces. The primary flag is preserved correctly.

  L3. Subnet doesn't derive Serialize/Deserialize — but SafetyCheckResult.active_subnets is Vec<String> (the CIDRs are flattened at the wire boundary). Consistent, but a Subnet result would preserve
  the primary/interface metadata; that gets discarded here.

  L4. network_context() calls default_net::get_default_interface() synchronously inside an async function. default_net is CPU-bound but does read /proc/net/route on Linux, which is a fast
  synchronous file read. Not a concern in practice.

  L5. Test file subnet_cidr_v4 in network_map.rs was moved to network_context.rs as the single source of truth. Good deduplication. The vestigial comment in network_map.rs:337-341 explains this.
  Fine.

  L6. macOS CI scope: The PR text openly acknowledges pentest-tools isn't in the macOS lane, which means these tests only run on Linux. Author's rationale (platform-invariant pure math) is sound.
  Ack.

jtomek-strike48:
Thanks @ngdevo — verified each point against the merged head (`b17074f` → squashed as `10195c7` on `main`):

**H1 (serialization/persistence):** Confirmed safe. Ran the grep you asked for:
- `grep -rn "\.ip_addresses" crates/ apps/` → **zero** remaining consumers; the rename fully propagated.
- No `to_json` / `serde_json::to_*` / `bincode` / `toml::to_*` / `to_writer` is ever called on `NetworkInterface` or `InterfaceAddr` anywhere. Every one of the ~9 sites constructs the struct fresh from a live `get_network_interfaces()` enumeration or reads it in-process — nothing deserializes a cached/persisted copy or a legacy fixture. The only value that crosses a wire boundary is `SafetyCheckResult.active_subnets`, which is `Vec<String>` (flattened CIDRs), not the struct. So the `ip_addresses → addresses` rename can't break a stored consumer.

**M2 (IPv6 fallback → malformed `fe80::1.0/24`):** Already fixed on the merged head. Your comment cited the pre-fix `network_plan.rs:111-117` string-slice — that's the same finding @cchao-devo raised as a SHOULD-FIX, and it was closed in `b17074f`: the fallback is now `derive_scan_cidr()`, which selects the first usable **IPv4** address (never any-family), derives via `subnet_cidr_v4` real-netmask math for both known- and missing-prefix cases, `warn!`s on the assumed /24, and returns `Error::InvalidParams` when the interface has no IPv4 rather than fabricating a CIDR from a v6 string. Four guard-checked unit tests cover all four shapes.

**M1 (`MIN_SUBNET_PREFIX = 16` silently narrows /12,/8; info-level log):** Good catch — real behavior, not a defect (it's the intentional #182/#299 scan-scope safety cap), so I've filed it as a follow-up rather than expand this PR: surface the clamp in the tool result (LLM-visible) and/or bump info → warn. See #351.

L1–L6 acked. Appreciate the thorough pass.

PR #349, opened 2026-07-29
feat(network): resolve auto/current/all scan targets + validate bypass tools (#346)

### LINKED ISSUE #346: Scan target pipeline: auto/current/all sentinels + validate the 5 bypass tools

jtomek-strike48:
Part of #344 (fact-first network discovery). This is **PR 2 - Target pipeline**. Closes #346.

> **Stacked on #348** (PR 1, `feat/network-context-foundation`). Base is that branch, not `main`, because this consumes `network_context()`. **Merge #348 first**, then rebase this onto `main` (or it retargets automatically once #348 merges).

## Summary

- **`auto`/`current`/`all` become first-class scan targets.** `current` -> the host's primary (default-route) subnet; `auto`/`all` -> every active local subnet, resolved scan-free from `network_context()` (#345). This makes the safe target the default instead of relying on the model to remember to gather facts.
- **One shared `prepare_scan_targets` pipeline** enforces a fixed order: **resolve sentinel -> `validate_target` (each) -> warn**. Routing the four previously-unvalidated discovery tools through it *closes the bypass gap* where they shelled a raw target string with no format/injection check.
- **Non-blocking out-of-subnet advisory:** a literal target outside every known local subnet (the `192.168.1.0/24`-on-`10.0.x` case) is flagged in the result but still scanned - routed/pivot scans are legitimate and never blocked.
- **Wiring per CLI capability** (verified empirically): `arp_scan`, `masscan_fast`, `nmap_vuln` accept multiple positional targets -> fan out in one invocation; `netdiscover`'s `-r` takes a single range -> loops per subnet with a split timeout budget and non-fatal per-subnet failures.
- Each tool result now carries `resolved_targets`/`resolved_ranges` + `out_of_subnet_warning`. Param descriptions document the sentinels (the schema has no `enum` slot).

## Scope correction

`responder` was dropped from the bypass-tool list: it targets an **interface**, not an IP/CIDR (only `interface` + bool params), so there's nothing to resolve or range-check. #346 updated. Real scope is the 4 target-taking tools.

## Verification

- `cargo check --workspace` clean; `cargo clippy -p pentest-tools --all-targets -- -D warnings` clean; fmt applied.
- `cargo test -p pentest-tools -p pentest-platform --lib` -> 354 + 58 pass, 0 fail (11 new `scan_target` tests).
- **Mutation-verified** 3 guards go red when broken: (A) `current` -> last-instead-of-primary; (B) dropping the auto/all empty-subnet guard; (C) inverting `ipv4_in_cidr` (4 tests red).
- **Multi-target CLI verified empirically:** `nmap -sL 10.0.8.0/30 192.168.99.0/30` scans both; arp-scan usage is `[hosts...]`; netdiscover `-r` is single-range (hence the loop); masscan doc-verified (not installed locally).

## Review dispositions (pick-reviewer pass)

- **BLOCKING "validate_target rejects masscan dash-range" - RETRACTED.** Verified empirically: `validate_target("10.0.0.1-10.0.0.20")` **accepts** it (passes as a hostname-shaped string). No regression. The finding was reasoned, not run.
- **netdiscover per-subnet timeout + fail-fast (SHOULD-FIX/LOW) - FIXED.** Timeout is now split across subnets (total ~= `timeout`, floored 1s/subnet); a per-subnet failure logs and continues (recorded in `failed_ranges`); only a wholesale failure errors out.
- **macOS CI scope (SHOULD-FIX) - deliberate, as in #348.** `scan_target` is platform-invariant pure logic and runs in the required Linux `--workspace` lane; not bloating the macOS lane.
- **Empty-subnet warning suppression (LOW):** by design - with no known subnets there's no basis to judge out-of-subnet, so no false warning is emitted.

## Test plan

- [x] `cargo test -p pentest-tools -p pentest-platform --lib` green (354 + 58)
- [x] `cargo clippy --all-targets -- -D warnings` clean
- [x] Mutation-verified resolver + containment guards (3 mutants red)
- [x] Multi-target CLI acceptance verified empirically
- [ ] CI green on the required Linux lane
- [ ] Rebased onto `main` after #348 merges

### DIFF
--- a/crates/tools/src/external/network/arp_scan.rs
+++ b/crates/tools/src/external/network/arp_scan.rs
@@ -34,7 +34,10 @@ impl PentestTool for ArpScanTool {
             .param(ToolParam::required(
                 "target",
                 ParamType::String,
-                "Target network (CIDR)",
+                "Target network (CIDR), or a sentinel: 'current' (this host's \
+                 primary subnet), 'auto'/'all' (every active local subnet). \
+                 Prefer a sentinel over guessing a range - it uses the host's \
+                 real interface subnets.",
             ))
             .param(ToolParam::optional(
                 "timeout",
@@ -63,7 +66,13 @@ impl PentestTool for ArpScanTool {
 
             let timeout_secs = crate::util::param_u64(&params, "timeout", 60);
 
-            let builder = CommandBuilder::new().positional(&target);
+            // Resolve auto/current/all -> real subnets, validate, and surface an
+            // out-of-subnet advisory. arp-scan accepts multiple positional
+            // targets ("arp-scan [options] [hosts...]"), so a multi-subnet
+            // resolution fans out in one invocation.
+            let prepared = crate::scan_target::prepare_scan_targets(&target).await?;
+
+            let builder = CommandBuilder::new().extend(&prepared.targets);
 
             let args = builder.build();
             let args_refs: Vec<&str> = args.iter().map(|s| s.as_str()).collect();
@@ -80,6 +89,8 @@ impl PentestTool for ArpScanTool {
 
             Ok(json!({
                 "target": target,
+                "resolved_targets": prepared.targets,
+                "out_of_subnet_warning": prepared.warning,
                 "hosts": hosts,
                 "count": hosts.len(),
             }))
--- a/crates/tools/src/external/network/masscan_fast.rs
+++ b/crates/tools/src/external/network/masscan_fast.rs
@@ -34,7 +34,10 @@ impl PentestTool for MasscanFastTool {
             .param(ToolParam::required(
                 "target",
                 ParamType::String,
-                "Target IP or CIDR range",
+                "Target IP or CIDR range, or a sentinel: 'current' (this host's \
+                 primary subnet), 'auto'/'all' (every active local subnet). \
+                 Prefer a sentinel over guessing a range - it uses the host's \
+                 real interface subnets.",
             ))
             .param(ToolParam::optional(
                 "rate",
@@ -70,8 +73,13 @@ impl PentestTool for MasscanFastTool {
             let rate = params.get("rate").and_then(|v| v.as_u64()).unwrap_or(10000);
             let timeout_secs = crate::util::param_u64(&params, "timeout", 120);
 
+            // Resolve auto/current/all -> real subnets, validate, warn if
+            // out-of-subnet. masscan accepts multiple positional targets, so a
+            // multi-subnet resolution fans out in one invocation.
+            let prepared = crate::scan_target::prepare_scan_targets(&target).await?;
+
             let builder = CommandBuilder::new()
-                .positional(&target)
+                .extend(&prepared.targets)
                 .arg("-p", "21,22,23,25,80,443,445,3306,3389,8080")
                 .arg("--rate", &rate.to_string());
 
@@ -90,6 +98,8 @@ impl PentestTool for MasscanFastTool {
 
             Ok(json!({
                 "target": target,
+                "resolved_targets": prepared.targets,
+                "out_of_subnet_warning": prepared.warning,
                 "open_ports": open_ports,
                 "count": open_ports.len(),
             }))
--- a/crates/tools/src/external/network/netdiscover.rs
+++ b/crates/tools/src/external/network/netdiscover.rs
@@ -35,7 +35,10 @@ impl PentestTool for NetdiscoverTool {
             .param(ToolParam::optional(
                 "range",
                 ParamType::String,
-                "IP range (e.g., 192.168.1.0/24)",
+                "IP range (e.g., 192.168.1.0/24), or a sentinel: 'current' (this \
+                 host's primary subnet), 'auto'/'all' (every active local subnet, \
+                 scanned one per invocation). Empty = netdiscover auto-scan. \
+                 Prefer a sentinel over guessing a range.",
                 json!(""),
             ))
             .param(ToolParam::optional(
@@ -53,7 +56,8 @@ impl PentestTool for NetdiscoverTool {
             .param(ToolParam::optional(
                 "timeout",
                 ParamType::Integer,
-                "Timeout",
+                "Total timeout in seconds, split across subnets when 'auto'/'all' \
+                 expands to several (each gets total/N, floored at 1s).",
                 json!(60),
             ))
             .platforms(vec![Platform::Desktop, Platform::Tui])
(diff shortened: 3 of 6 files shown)


ngdevo:
HIGH

  H1. resolve_scan_target unconditionally enumerates interfaces — docstring is wrong.
  crates/tools/src/scan_target.rs:206-213. The docstring says "Only performs the interface enumeration when the target is actually a sentinel or a concrete address worth range-checking — a hostname
  passes straight through without touching the network." But the implementation is:
  let subnets: Vec<Subnet> = network_context().await.unwrap_or_else(|e| { ... Vec::new() });
  resolve_against_subnets(raw, &subnets)
  network_context() is called every time. Not a security issue, but the docstring lies and the per-hostname enumeration cost is real (each network scanner tool call triggers a
  platform.get_network_interfaces().await). Either short-circuit for !is_sentinel(raw) && target_ipv4(raw).is_none(), or fix the doc.

  MEDIUM

  M1. Netdiscover result loses per-subnet host attribution.
  crates/tools/src/external/network/netdiscover.rs:127-149. When auto/all fans out to multiple subnets, all hosts get flattened into a single hosts array. The result carries resolved_ranges and
  failed_ranges but no hosts_by_range mapping. An agent that finds a suspicious host at 10.0.8.42 cannot tell which subnet's discovery surfaced it (or whether it was seen on more than one). For a
  multi-homed target this is genuinely useful triage information that's now discarded.

  M2. Netdiscover per-range timeout can starve when subnet count grows.
  per_range_timeout = Duration::from_secs((timeout_secs / ranges.len().max(1) as u64).max(1)). Integer division floors: timeout=60, ranges=4 → 15s each, but timeout=60, ranges=61 → 0 → floor to 1s
  each. On a system with lots of active subnets (VPN + docker + wifi + wired), a 60s total means each subnet gets 1s and none complete. Consider: a floor on total (timeout_secs * ranges.len() for
  auto/all) or a per-subnet minimum with a documented cap on N.

  M3. prepare_scan_targets emits both tracing::warn! and returns the warning in the result.
  Line 302. Fine, but duplicated across log sink and tool output isn't documented. Not a bug, worth noting.

  LOW

  L1. resolved_targets field is added to arp_scan, masscan_fast, nmap_vuln results; resolved_ranges (plural, different name) to netdiscover. Naming inconsistency across tools that do the same
  conceptual thing.

  L2. ipv4_in_cidr handles a malformed CIDR by returning false, silently. The safe-by-default choice, but the caller has no visibility into "your active-subnet list contains a malformed CIDR" —
  which shouldn't be possible given subnet_cidr_v4 produces canonical strings, but defensive parsing that silently drops errors is a class of bug worth surfacing.

  L3. is_sentinel is exported (pub fn) but only used inside this module. pub(crate) would be enough.

  L4. out_of_subnet_warning message includes every known subnet CIDR. On a host with 15 active subnets that's a lot of text going to the LLM context per warning. Fine, just note.

  L5. The stack (PR #349 based on #348 base ref) means the review dance is: land #348, then rebase #349 on main. The PR body handles this. Good discipline.

  L6. BLOCKING-retracted "validate_target rejects masscan dash-range" review: verified — validate_target("10.0.0.1-10.0.0.20") splits on . into labels ["10", "0", "0", "1-10", "0", "0", "20"], each
  label passes the alphanumeric-hyphen check in validate_hostname. The retraction is correct. (Sibling risk: 10.0.0.1,10.0.0.20 fails validation because comma is not in the alnum-hyphen set — worth
  knowing since some tools accept comma-lists.)

jtomek-strike48:
Thanks @ngdevo — all verified against the current head and addressed in `75e1443`:

**H1 (docstring lies — unconditional enumeration):** Fixed. Confirmed the body called `network_context()` on every path regardless of target. Gated it behind a pure `needs_subnet_context(raw)` = `is_sentinel(raw) || target_ipv4(raw).is_some()`, so a hostname (or any non-IPv4 literal) now skips enumeration entirely — the documented "passes straight through without touching the network" contract is actually true. Unit-tested (`needs_subnet_context_only_for_sentinels_and_ipv4`), guard-checked (forcing it to `true` turns the test red).

**M2 (per-range timeout starvation):** Fixed. Extracted `per_range_timeout_secs(total, n)` with a `MIN_PER_RANGE_TIMEOUT_SECS = 10s` floor. When the floor binds (your `60s / 61 subnets → 1s` case), total time exceeds the requested budget rather than running sweeps too short to complete. Also fixed the stale "floored at 1s" schema doc. Unit-tested at the boundary (`per_range_timeout_floors_so_many_subnets_dont_starve`), guard-checked (removing the floor turns it red).

**L3 (`is_sentinel` visibility):** `pub` → `pub(crate)` — confirmed no external users.

**M1 (netdiscover per-subnet host attribution):** Agreed this is genuinely useful triage info, but it's an additive result-shape change (a `hosts_by_range` map) rather than a correctness fix, so I've filed it as a follow-up to keep this PR scoped: #352.

Other L's:
- **L1** (naming: `resolved_targets` vs `resolved_ranges`): noted; a naming-consistency pass across the discovery tools is worth its own small cleanup — can fold into #352 or a separate chore.
- **L2** (`ipv4_in_cidr` silently returns false on malformed CIDR): defensive-by-default, and `subnet_cidr_v4` only ever produces canonical strings, so it's unreachable in practice — leaving as-is.
- **L4** (out-of-subnet warning lists every CIDR): acked, low text volume in practice.
- **L6**: thanks for re-verifying the dash-range retraction, and the comma-list sibling-risk note is a good catch — worth a validation follow-up if any tool accepts comma-lists.

ngdevo (approved):
LGTM

PR #354, opened 2026-08-03
fix(port_scan): infer host-level reachability from an all-timeout scan (#337)

### LINKED ISSUE #337: port_scan: infer host-down from an all-timeout / zero-response scan (follow-up to #306/#333)
## Problem

`port_scan` (via `classify_connect` in `crates/platform/src/common.rs`) now distinguishes an unreachable host from a closed port **when the OS supplies an errno** (`HostUnreachable`/`NetworkUnreachable`/`PermissionDenied`). That covers IPv6 no-route, on-link ARP failure, and sandbox `PermissionDenied`.

It does **not** cover the most common real case: a dead IPv4 host on a normally-routed network. There, `TcpStream::connect_timeout` **times out** (default 2000ms) before the OS emits `HostUnreachable`, so every port classifies as `Filtered`. An all-`Filtered`, zero-response scan of a down host is therefore still indistinguishable from a scan of a live-but-silently-firewalled host - which is part of what #306 set out to fix.

This was left out of #333 (PR that fixed the errno cases) deliberately, because inferring "host down" from an all-timeout result is a judgment call with a real false-positive risk (see below), not a mechanical classification.

## Proposed direction (needs design)

When a scan of a single host yields **zero responses of any kind** (no open, no `ConnectionRefused`, no errno-unreachable - every probe timed out), consider surfacing a host-level `likely_down` / `no_response` signal distinct from per-port `Filtered`.

Open questions for the implementer:
- **False positives:** a host behind a default-drop firewall is *live* but produces the identical all-timeout signature. Reporting it as "down" would mislead the operator. Is a separate `no_response` state (neither "down" nor "filtered-but-up") the honest framing?
- **Signal strength:** does an all-timeout result on a *single* port mean anything, or only across a representative port set? A cheap liveness probe (ICMP echo / a common-port TCP knock) before/alongside the scan would disambiguate but adds capability + consent surface.
- **Bounded probing / no-DoS:** any added liveness probe must stay within the engagement's non-destructive, bounded-probing posture.

## Acceptance criteria (to refine at design time)

- [ ] An all-timeout, zero-response single-host scan is distinguishable in the tool JSON from a scan that got at least one response, without falsely asserting "down" for a silently-firewalled live host.
- [ ] Behavior is documented (what each host-level state means for the operator/agent).
- [ ] Regression test for the all-timeout-on-routed-network case.

## References

- Follow-up to #306 / #333 (PR #333 handles the errno-supplied unreachable cases; this is the timeout case it explicitly scoped out).
- `crates/platform/src/common.rs` `classify_connect` (the `TimedOut => Filtered` arm) and `tcp_port_scan` assembly.

jtomek-strike48:
Closes #337. Follow-up to #306/#333.

## Summary

- #306/#333 made an **errno-supplied** unreachable (`HostUnreachable`/`NetworkUnreachable`/`PermissionDenied`) distinct from a closed port, but explicitly scoped out the most common real case: a **dead IPv4 host on a routed network**, where `connect_timeout` times out before the OS emits an errno — so every port classifies `Filtered` and an all-timeout scan of a down host is indistinguishable from a scan of a live-but-silently-firewalled host.
- Adds a host-level `HostReachability { Reachable, Unreachable, NoResponse }`, computed in `assemble_scan_result` and surfaced in the `port_scan` tool JSON as `reachability` + a plain-language `host_state_note`:
  - **Reachable** — ≥1 `Open` or `Closed` (a refusal is still a response → host is up).
  - **Unreachable** — every probe was errno-`Unreachable` (the #306/#333 case).
  - **NoResponse** — ports scanned, zero responses, ≥1 bare timeout.
- **Honesty constraint (the crux of #337):** `NoResponse` is deliberately **not** "host down". A live host behind a default-drop firewall produces the identical zero-response signature, so both the enum docs and `host_state_note` name both possibilities and steer to a separate liveness probe rather than asserting the host is offline.

## Blast radius

`ScanResult` gains one field, constructed in exactly one place (`assemble_scan_result`); its only consumer is the `port_scan` tool JSON. `#[serde(default)]` keeps older/persisted results deserializable.

## Test plan

- [x] `cargo clippy --workspace --features pentest-platform/desktop-pcap -- -D warnings` clean
- [x] `cargo test --workspace --features pentest-platform/desktop-pcap` — platform 64/0, tools 349/0
- [x] Regression test for the all-timeout-on-routed-network case (`all_timeout_scan_is_no_response_not_down_or_unreachable`)
- [x] Adversarial port-state mixes covered: all-Filtered, Filtered+Unreachable, single Filtered, empty, Open+Unreachable, Closed-only
- [x] Guard-checked by mutation: neutering `infer_reachability` reddens the reachability tests; injecting an "offline" verdict into `host_state_note` reddens the honesty test
- [ ] Live headless run confirming the JSON `reachability`/`host_state_note` on a real down host (not run — needs a live target)

## Notes for reviewers

- CI scope: `common.rs` reachability tests run on both the Linux (`--workspace`) and macOS (package-scoped `pentest-platform`) lanes; the `port_scan.rs` `host_state_note` tests run on the Linux lane (tools crate not in the macOS lane) — standard for this repo.
- This closes the bug *class* the follow-up targeted; no sibling scan path has the same all-timeout blindness.

### DIFF
--- a/crates/platform/src/common.rs
+++ b/crates/platform/src/common.rs
@@ -3,7 +3,9 @@
 //! This module extracts duplicated port-scanning and ARP-parsing code that was
 //! previously copy-pasted across android, desktop, and iOS implementations.
 
-use crate::traits::{port_to_service, ArpEntry, PortState, ScanResult, ScannedPort};
+use crate::traits::{
+    port_to_service, ArpEntry, HostReachability, PortState, ScanResult, ScannedPort,
+};
 use std::io;
 use std::net::{SocketAddr, TcpStream, ToSocketAddrs};
 use std::sync::Arc;
@@ -102,6 +104,8 @@ fn assemble_scan_result(host: String, ports: Vec<ScannedPort>, duration_ms: u64)
         .filter(|p| p.state == PortState::Unreachable)
         .count();
 
+    let reachability = infer_reachability(&ports);
+
     let errors: Vec<String> = ports
         .iter()
         .filter(|p| p.state == PortState::Unreachable)
@@ -125,6 +129,40 @@ fn assemble_scan_result(host: String, ports: Vec<ScannedPort>, duration_ms: u64)
         open_count,
         unreachable_count,
         errors,
+        reachability,
+    }
+}
+
+/// Infer host-level reachability from a single-host scan's port states (#337).
+///
+/// The honest three-way split:
+/// * any `Open` or `Closed` -> [`HostReachability::Reachable`]: the host
+///   answered at least once (a refusal is still an answer), so it is up.
+/// * ports scanned, none reachable-responding, and **every** probe was
+///   `Unreachable` -> [`HostReachability::Unreachable`]: an errno said no packet
+///   reached the target (the #306/#333 case).
+/// * ports scanned, no response of any kind, at least one `Filtered` (timeout)
+///   and no proof of reach -> [`HostReachability::NoResponse`]: down OR silently
+///   firewalled. Deliberately not called "down" (#337 false-positive guard).
+///
+/// An empty port list is [`HostReachability::Reachable`] (nothing was probed, so
+/// there is no basis to claim a reachability problem).
+fn infer_reachability(ports: &[ScannedPort]) -> HostReachability {
+    // A response of any kind (listening OR actively refused) proves the host is up.
+    let got_response = ports
+        .iter()
+        .any(|p| matches!(p.state, PortState::Open | PortState::Closed));
+    if got_response || ports.is_empty() {
+        return HostReachability::Reachable;
+    }
+
+    // No response. If every probe carried an errno-unreachable, the host was
+    // never reached; otherwise (at least one bare timeout) it is the ambiguous
+    // down-or-firewalled case.
+    if ports.iter().all(|p| p.state == PortState::Unreachable) {
+        HostReachability::Unreachable
+    } else {
+        HostReachability::NoResponse
     }
 }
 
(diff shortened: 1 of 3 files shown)


cchao-devo (approved):
Review of pick PR #354 — fix(port_scan): infer host-level reachability from an all-timeout scan (#337)

  Verdict: APPROVE — merge as is.

  Well-engineered follow-up that closes the #337 gap without over-claiming. The classification is correct and exhaustive, the honesty constraint is enforced in both the type docs and the
  operator-facing note, and the tests are thorough and actually run. I verified the logic against the full type space and executed the new tests rather than reading them by eye.

  What I verified (beyond the diff)

  - infer_reachability is exhaustive. PortState has exactly four variants (Open, Closed, Filtered, Unreachable), so the three-way split is total: Open|Closed ⇒ Reachable; else all Unreachable ⇒ 
  Unreachable; else (≥1 Filtered, no response) ⇒ NoResponse. Each branch matches its doc.
  - empty ⇒ Reachable can't misfire in production. The sole caller (tcp_port_scan) spawns one task per requested port and, on a join failure, records a fallback Unreachable ScannedPort (the #306
  fix) — so results is empty only when zero ports were requested, which is legitimately "no basis to flag." A failed scan of requested ports always records at least one port, so it can never be
  silently labeled Reachable.
  - Honesty constraint holds. NoResponse never asserts "down" in either the enum docs or host_state_note; the note names both the offline and default-drop-firewall possibilities and steers to a
  separate liveness probe. Guarded by no_response_note_never_asserts_host_is_down.
  - Tests pass and execute (not skipped): all six common.rs reachability tests run green on the pentest-platform lib lane (62 passed / 0 failed), and both port_scan.rs note tests run green on
  the tools lane (349 passed / 0 failed) — matching the PR's stated counts.

  Type design
  
  HostReachability is a tight Copy enum with #[serde(rename_all = "snake_case")]; embedding reachability + host_state_note in the tool JSON is clean. Using a named default_for_deser for
  #[serde(default = …)] instead of a blanket impl Default is a deliberate, correct call — it keeps persisted results deserializable without implying a general-purpose default that could be
  misused elsewhere.

  Minor (non-blocking)
  
  - The only unexercised item is the live-headless run of the JSON reachability/host_state_note against a real down host (checkbox left unchecked). Low risk — the value is unit-tested and
  serialization is derive-only — but worth a one-off manual confirmation when a live target is handy.

Issue #360, opened 2026-08-10
all_tools_registered compares all registered tools against host-supported capabilities; fails on macOS


cchao-devo:
`pentest-core`'s `all_tools_registered` fails on macOS. It is a **test bug, not a code
bug**: the test asserts an equivalence that the production code deliberately does not
hold.

```
thread 'all_tools_registered' panicked at crates/core/tests/tool_integration.rs:569:9:
tool 'wifi_scan_detailed' not in capabilities: [...]
```

## Root cause

The test (`crates/core/tests/tool_integration.rs:562-573`) compares two sets that are
intentionally different:

- `pentest_tools::tool_names()` (`crates/tools/src/lib.rs:275`) -> `registry.names()`
  (`crates/core/src/tools.rs:883`), which returns **every registered tool**.
- `connector.capabilities()` (`crates/ui/src/liveview_connector/pick_connector.rs:154`),
  which returns `supported_schemas()` -- i.e. **only tools supported on this host**, with
  the comment "Advertise only tools supported on this host (platform + desktop OS). See
  #183."

Tools are registered unconditionally and the OS gate is applied at advertise/execute time
(that design is documented at `crates/core/src/tools.rs:900-906`). So on any host where
some tool is not supported, `names()` is a strict superset of `capabilities()` and the
test cannot pass. It happens to pass on Linux because both Linux-only tools are supported
there.

Exactly two tools are Linux-only:

- `crates/tools/src/wifi_scan_detailed.rs:126` -> `vec![DesktopOs::Linux]` ("requires
  aircrack-ng; Linux only for now (#183)")
- `crates/tools/src/autopwn/crack.rs` -> same

## Why the reported tool name changes between runs

Earlier runs failed on `autopwn_crack`; it now fails on `wifi_scan_detailed`. This is not
instability in the tools -- `ToolRegistry.tools` is a `HashMap<String, BoxedTool>`
(`crates/core/src/tools.rs:821`), whose iteration order is randomized per process. The
test's `for name in tool_names()` loop asserts on the first unsupported tool it happens to
reach, so the *reported* name varies while the underlying mismatch set is the same two
tools every time.

This is worth fixing for its own sake: the drifting name makes the failure look
intermittent and sends readers chasing the wrong tool.

## Impact

- The CI `Test` job is red on macOS. Note this only became *visible* recently: before
  `--no-fail-fast` was added to the workflow, cargo aborted at the first failing target,
  which masked this (and prevented later crates' tests from being built at all).
[...]

Issue #359, opened 2026-08-10
flaky: evidence_producer::multiple_push_single_drain races siblings on the global PENDING_EVIDENCE buffer


cchao-devo:
Observed on a clean tree with no local diff: **2 failures in 10 consecutive runs** of the
`pentest-tools` lib target.

```
thread ..multiple_push_single_drain panicked at crates/tools/src/evidence_producer.rs:617:9:
assertion `left == right` failed: Should find our 10 nodes
  left: 6
 right: 10
```

## Root cause

`PENDING_EVIDENCE` (`crates/tools/src/evidence_producer.rs:47`) is a process-global
`LazyLock<RwLock<Vec<EvidenceNode>>>`, and `drain_pending_evidence()` (`:139`) is
`std::mem::take` — it removes **everything** in the buffer, not just the calling test's
nodes.

`multiple_push_single_drain` (`:593`) pushes 10 uniquely-named nodes, then drains and
filters the result to its own `test-multi-unique-` prefix. But
`evidence_flows_through_buffer` (`:527`) drains the same global buffer, and cargo runs
both concurrently in one process. If the sibling drains between this test's pushes and
its own drain, the sibling's `mem::take` carries off some of these 10 nodes and the
filtered count comes up short (observed 6 of 10).

The unique-ID scheme the test comments describe ("Uses unique IDs to avoid interference
from parallel tests") defends against *name collisions*, but not against another thread
draining the shared buffer — which is the actual shared-state hazard here.

Three sibling tests that touch the same static are already `#[ignore]`d with
`"Requires exclusive access to global buffer - run with --test-threads=1"` (`:634`,
`:684`, `:734`). This test has the same requirement but was left enabled.

## Impact

Intermittently reddens the CI `Test` job on unrelated PRs. Load-sensitive: it did not
reproduce in 3 of 3 runs on an idle machine, then failed 2 of 10 under load.

## Reproduce

```bash
for i in $(seq 1 10); do
  DISABLE_SANDBOX=true cargo test --package pentest-tools --lib --locked \
    --features pentest-platform/desktop-pcap 2>&1 | grep -E "^test result:" | tail -1
done
```

Passes 5 of 5 when the module is run alone (`... --lib evidence_producer`), consistent
with a cross-test race rather than a logic bug.

## Suggested fix

Prefer a real fix over `#[ignore]`, since ignoring loses the coverage:

1. Serialize the buffer-touching tests behind a shared `static Mutex<()>` in the test
[...]

PR #361, opened 2026-08-10
feat(plg): StrikeHub PLG Pick+DVWA compose stack bound via OTT, with pre-runner tenant resolution (matrix#3519)

cchao-devo:
## What and why

  Adds a PLG-mode sibling of `docker-compose.dvwa.yml` so an external Pick+DVWA can
  attach to StrikeHub PLG mode (matrix#3519), plus the connector-side fix that makes
  the One-Time-Token (OTT) binding actually work.

  The demo compose hardcodes a fixed tenant UUID in `STRIKE48_TENANT`. PLG has no fixed
  tenant: the personal tenant is minted at login (`personal-<hash>`), so its UUID is
  unknowable ahead of time. The binding is instead the StrikeHub-minted OTT, which bakes
  `tenant_id` into the connector's Keycloak JWT via `register-with-ott`.

  ## Changes
  
  ### 1. Pre-runner OTT tenant resolution (`crates/ui/src/liveview_connector/mod.rs`)

  New `resolve_ott_tenant()`, called from `connect_and_run` before the `ConnectorRunner`
  is constructed. This works around an upstream SDK defect (filed separately): the SDK's
  `initialize_auth()` writes back only `auth_token` and drops the `Credentials.tenant_id`
  returned by the OTT exchange, so registration then sends the wrong tenant and the server
  rejects it with `tenant_mismatch`, even though the exchange succeeded.

  Constraints the implementation holds:

  - **Scoped to the `STRIKE48_REGISTRATION_TOKEN` env carrier only.** Exchanging up-front
    consumes the single-use OTT, so the runner's own `initialize_auth` must not exchange
    it again (that would 401 and abort). The only way to suppress the second attempt is to
    unset the variable the SDK gates `has_ott()` on, which is possible for the env carrier
    and not for the FILE carrier / mounted k8s secret. A file-carried OTT is therefore left
    entirely to the SDK (those deployments pass an explicit tenant).
  - **A fresh OTT still wins over saved credentials**, mirroring SDK precedence, so an
    operator can rebind to a different tenant. Saved credentials are the fallback when the
    exchange fails, which also fixes the "a spent OTT bricks restarts" symptom.
  - **Cert-manager / direct-config (Priority 1) is left untouched** via a `has_direct_config()`
    guard that returns `None`.
  - Sets both `sdk_config.tenant_id` and `self.config.tenant_id` so the connector banner
    and the Matrix chat session point at the right tenant too.
  - Trims a padded carrier in place: the SDK does not trim the env carrier (only the FILE
    one), so a hand-edited `.env.plg` with a stray space would otherwise 401 and silently
    land in the saved-credentials fallback.

  ### 2. `docker-compose.plg.yml` (new)

  Clone of the DVWA stack with:

  - `STRIKE48_REGISTRATION_TOKEN` mandatory (the OTT is the binding); no `STRIKE48_TENANT`
  - `STRIKE48_API_URL` set (the `register-with-ott` origin the SDK reads and re-reads to
    enforce a same-origin allowlist on the server-supplied register URL)
  - `CONNECTOR_NAME=pentest-connector` (StrikeHub mints the OTT for the `pentest-connector`
    type; any other name is rejected `ott_type_mismatch`)
  - `name: pick-plg` compose project + distinct `container_name: pick-plg-*` / volume
    `pick-plg-creds`, so it never collides with a demo stack
  - preserves the DVWA scan-net isolation invariant (DVWA on `internal` scan-net, no host ports)

  ### 3. `justfile`

  `plg-up` / `plg-down` / `plg-check` recipes mirroring `dvwa-*`. `plg-check` asserts DVWA
  isolation and that `STRIKE48_REGISTRATION_TOKEN` is present / `STRIKE48_TENANT` absent.
  `plg-down` reaps containers that hold the `pick-plg-*` names but carry a stale compose
  project label from before `name: pick-plg` existed (named one by one, regex-anchored, so
  it never touches `pick-demo-*`).

  ### 4. CI (`.github/workflows/ci.yml`)

  Adds `--no-fail-fast` to the workspace test run. Without it, cargo stops at the first
  failing test target and never builds later crates, so a single red target hid every later
  crate's tests. See the pre-existing failure below for why this matters now.

  ### 5. `.env.plg.example` (new) + `.gitignore`

  Template with an `ott_...` sentinel and no tenant UUID; `.gitignore` excludes the
  generated `.env.plg` (it carries a live OTT).

  ## Verification

  - `cargo fmt --all --check`: clean
  - `cargo clippy --workspace --locked --features pentest-platform/desktop-pcap -- -D warnings`: clean
  - `cargo test --workspace --no-fail-fast`: all pass except one pre-existing failure (below)
  - `just plg-check`: compose valid, DVWA isolated, PLG bound via OTT with no `STRIKE48_TENANT`
  - End-to-end on a from-scratch newdev cluster: OTT exchange succeeds, tenant adopted,
    `Registered successfully` with zero reconnects, scan runs to a generated report.

  ## Pre-existing CI failures (not introduced by this change)

  The `Test` job will show one red target. It is pre-existing and reproduces identically on
  a clean `main` checkout with this branch's changes stashed:

  - **`pentest-core` `all_tools_registered`** (`crates/core/tests/tool_integration.rs`) —
    panics with `tool 'autopwn_crack' not in capabilities`. This is a tool-registry
    completeness assertion, unrelated to this PR; this branch touches nothing under
    `crates/core`. Tracked in #360.
  - **`pentest-tools` evidence-buffer race** — a global-singleton `PENDING_EVIDENCE` buffer
    makes two `webwright_integration` tests flaky under concurrent drain. Tracked in #359.

  Verified by running the failing target on clean `origin/main`: same panic, same tool.


jtomek-strike48 (commented):
## Review — verified empirically at head `0f9af74`

Checked out the PR head in an isolated worktree and ran the code rather than reasoning about the diff. **No blocking issues.** The mechanism is correct, the SDK rationale is accurate, the tests run in the required CI job and provably guard the fix, and the demo infra is well isolated. One SHOULD-FIX (a false CI claim) and one LOW (comment wording).

### Verified sound

- **SDK claims — confirmed against the resolved `strike48-connector 0.4.4` source** (`Cargo.lock` pins 0.4.4):
  - `initialize_auth` (connector.rs:2114-2128) writes back only `auth_token`, never `credentials.tenant_id` → registration sends the stale config tenant. This is exactly the defect being worked around. ✓
  - OTT (`register_with_ott(...).await?`, connector.rs:2116) runs **before** saved-creds (:2138) and `?` propagates → a spent OTT genuinely bricks `restart: unless-stopped`. ✓
  - env carrier passed raw (`parse_ott`, no trim); file carrier trimmed (ott_provider.rs:282) → the in-place trim is genuinely needed. ✓
  - `register_with_ott` persists creds (`save_credentials`, ott_provider.rs:366) → the "remove carrier → SDK falls through to saved creds" handoff is real. ✓
- **CI-scope (Lens 2):** ran the exact CI Linux command (`cargo test --workspace --locked --no-fail-fast --features pentest-platform/desktop-pcap`) — all **9 `ott_tenant_tests` compiled and passed**, whole run exited 0. Pulled in via `apps/headless` unifying `pentest-ui/connector`; correctly absent from the package-scoped macOS lane.
- **Guard (Lens 1):** neutered the success-path `remove_var` (mod.rs:780) — exactly `fresh_ott_wins_over_saved_credentials` and `exchange_returning_blank_tenant_adopts_nothing_and_clears_the_carrier` went **RED** (*"a CONSUMED OTT must be removed... got Ok(\"ott_freshlyminted\")"*), other 7 stayed green. Real guards, not tautological.
- **Consumer trace (Lens 5):** headless (`apps/headless/src/main.rs:189`, the PLG path) and desktop both route through `connect_and_run → resolve_ott_tenant`. Fresh-OTT-wins precedence, empty-tenant handling, and error-path carrier logic all check out.
- **Infra:** compose preserves DVWA isolation (`scan-net internal:true`, no host ports, distinct `pick-plg` project + creds volume); `.env.plg.example` uses a non-secret sentinel; `plg-check` asserts the real SDK contract. No findings.

---

### 🟠 SHOULD-FIX — the `--no-fail-fast` justification claims a CI failure that doesn't occur

The new `ci.yml` comment and the PR body assert `pentest-core`'s `all_tools_registered` **fails "in this job"** and that *"The Test job will show one red target."* Both are contradicted by evidence:

- The Test check on this PR concluded **`success`** (the `cargo test ... --no-fail-fast` step itself is green).
- Running the identical command locally shows **`test all_tools_registered ... ok`** and exits **0**.
- Root cause: `autopwn_crack` is only registered under the `desktop-pcap` feature set CI uses; the panic reproduces only in a *bare* `cargo test` **without** those features. So in CI that target passes and `pentest-ui`'s tests already run.
- Secondary: *"never builds the remaining crates"* — cargo compiles all test targets before running any; `--no-fail-fast` changes test **execution** (don't stop at the first failing binary), not compilation.

The flag itself is fine and useful. Suggested rewrite:

```yaml
      # --no-fail-fast: cargo compiles every test target, then stops RUNNING at
      # the first target with a failing test. Without this flag a single red
      # target hides every later crate's test *results* (they're compiled but not
      # run/reported), so you learn about failures one CI round at a time. With it,
      # the job reports ALL failures in one run; exit code is unchanged, so any
      # failure still fails the job.
      - run: cargo test --workspace --locked --no-fail-fast --features pentest-platform/desktop-pcap
```

And drop/adjust the "one red target" line in the PR body to match the green Test check.

### 🟡 LOW — env-mutation SAFETY comment justification is imprecise (code is safe)

`resolve_ott_tenant` `set_var`/`remove_var`s `STRIKE48_REGISTRATION_TOKEN` (mod.rs:732/780/833) while the desktop poller (`connector_app.rs:485` → `read_credentials_tenant_id` → `env::var("HOME")`, config.rs:443) reads env. The comment dismisses this with *"Under edition 2021 this is moot — remove_var is safe."*

That reasoning doesn't hold: `setenv`/`unsetenv` mutate the shared `environ` array and race a concurrent `getenv` on **any** variable — UB in all editions (that's *why* 2024 made them `unsafe`), not safe because the two variables differ.

The code is nonetheless **race-free**, but for the reason worth stating instead: `spawn` here is `dioxus::prelude::spawn` (single-threaded cooperative executor — the synchronous getenv/setenv never interleave or run on parallel threads), and the headless (`#[tokio::main]`) path spawns no concurrent env reader during `resolve_ott_tenant`. Optional: reword the comment to cite that; it only becomes a real race if these tasks are moved onto a multi-threaded executor alongside a concurrent env reader.

cchao-devo:
@jtomek-strike48 ,

  Thanks for the empirical review — and for catching the false CI claim. Addressed both in `4552506`:
  
  **SHOULD-FIX (the `--no-fail-fast` "one red target" claim):** you're right, and I verified it —
  the actual `Test` check concluded `success`, and `autopwn_crack` is OS-gated (registers on the
  Linux CI lane, absent on macOS/bare `cargo test`), so that target passes in CI. My earlier 
  "verification" ran the CI command on macOS and got a red that falsely confirmed the wrong story.
  Replaced the ci.yml comment with your suggested accurate wording (cargo compiles all targets,
  stops *running* at the first failure; reports all failures; exit code unchanged), and I'll drop
  the "one red target" line from the PR body.
  
  **LOW (env-mutation SAFETY comment):** agreed — the "edition 2021 makes it moot" reasoning was
  wrong; setenv/getenv racing getenv is UB in every edition. Reworded the comment to cite the real
  reason it's race-free here: the single-threaded Dioxus cooperative executor (desktop `spawn`) and
  the headless `#[tokio::main]` path spawning no concurrent env reader during `resolve_ott_tenant` —
  noting it only becomes a real race under a multi-threaded executor with a concurrent env reader. 
  
  Both are comment-only; the 9 `ott_tenant_tests` still pass. Rebased onto current `main` as well.

jtomek-strike48 (approved):
## Re-review (approve) — PLG OTT stack, matrix#3519

Both findings from my prior review at `0f9af74` are resolved by `4552506e`, verified against the diff:

- **SHOULD-FIX (ci.yml `--no-fail-fast` claim):** the comment no longer asserts `pentest-core`'s target "currently fails" in this job. It now correctly explains compile-all-targets, then stop RUNNING at the first failing target, with an unchanged exit code. Matches the green `Test` check.
- **LOW (env-mutation SAFETY):** rewritten to the accurate reason. `set_var`/`remove_var` race a concurrent `getenv` on ANY variable (UB in every edition, which is why edition 2024 made them `unsafe`), so race-freedom here comes from the single-threaded Dioxus executor plus the headless path spawning no concurrent env reader during `resolve_ott_tenant`, not from the two variables differing.

Everything I verified empirically last round still holds (the SDK 0.4.4 `initialize_auth` tenant drop, OTT-over-saved-creds precedence, the 9 `ott_tenant_tests` guarding via the required CI job, DVWA scan-net isolation). All CI green. Approving.

Residual LOW (non-blocking, body-only): the PR description still says "The `Test` job will show one red target" and "all pass except one pre-existing failure". That is the same overstatement as the old ci.yml comment. The `Test` check is green because CI runs with `--features pentest-platform/desktop-pcap`, under which the target passes; the panic only reproduces in a bare `cargo test`. Worth tidying the body to match, but it does not gate merge.

Issue #117, opened 2026-05-22
Add aggression level to SpecialistContext so spawned specialists can modulate their probing depth


jtomek-strike48:
## Problem

The four specialist sub-agent prompts shipped in PR #116 each describe how the specialist should modulate its own probing depth based on the engagement's `AggressionLevel` — Conservative/Balanced/Aggressive/Maximum. This is the intended design: aggression should affect both *spawn frequency* (Red Team's decision, already implemented in `aggression.rs::SpawnPolicy::to_guidelines()`) **and** *probing depth* (specialist's behavior once spawned, just-shipped via the prompts).

But the carrier signal between the two layers is missing: `SpecialistContext` (the struct serialized into the spawned specialist's runtime context) does not include `aggression`. The specialist receives `targets`, `initial_findings`, `concerns`, and `attack_surface`, but no aggression level.

`SpecialistSpawner` *knows* the level — it's a field on the spawner — but it never gets passed through to the specialist via `SpecialistContext`. The specialist has no way to read its level deterministically; it can only fall back to a default behavior (Balanced) regardless of what the engagement actually configured.

## Code references

- `crates/core/src/specialist_spawner.rs:67-81` — `SpecialistContext` struct (needs new `aggression` field)
- `crates/core/src/specialist_spawner.rs:110-126` — `SpecialistSpawner::new()` and `aggression` field (already has the value)
- `crates/core/src/specialist_spawner.rs:162-208` — `SpecialistSpawner::spawn()` (needs to populate the new field on the context before serializing)
- `skills/claude-red/specialists/*.md` — prompts already reference "the engagement's AggressionLevel, which the Red Team agent passed through when spawning you"

## Proposed change

```rust
#[derive(Debug, Clone, Serialize, Deserialize)]
pub struct SpecialistContext {
    pub targets: Vec<String>,
    pub initial_findings: Vec<String>,
    pub concerns: Vec<String>,
    pub attack_surface: AttackSurface,
    /// Engagement-level aggression. Determines how aggressively the
    /// specialist probes within its scope. See `AggressionLevel` and
    /// the per-specialist "Aggression policy hooks" sections in
    /// `skills/claude-red/specialists/*.md`.
    pub aggression: AggressionLevel,
}
```

[...]

PR #363, opened 2026-08-12
fix(deps): bump webbrowser to 1.2.4 (RUSTSEC-2026-0257)

jtomek-strike48:
## Summary
- Bumps `webbrowser` 1.2.0 -> **1.2.4** to resolve **RUSTSEC-2026-0257** ("Unix `BROWSER` handling allows browser argument injection"), fixed in >=1.2.2.
- **Pins 1.2.4 rather than the advisory floor 1.2.2.** 1.2.2 introduced a regression ([upstream #120](https://github.com/amodm/webbrowser-rs/issues/120)) that appends the URL twice when `$BROWSER` is set without a `%s` placeholder — the same Unix code path this fix hardens. 1.2.4 fixes that (plus #121 and the 1.2.3 Windows UNC path fix), and its dependency set is **identical to 1.2.2's**, so the lock change stays surgical. Its trigger (`BROWSER` without `%s`) is disjoint from the advisory's precondition (`BROWSER` with `%s`), so this is not a security regression — just avoids shipping a known UX regression on the same path we're touching.
- `webbrowser` is transitive via `dioxus-desktop 0.7.3` -> `dioxus 0.7.3`; nothing in the workspace depends on it directly.
- Minimal 3-line lockfile change: version, checksum, and webbrowser's own dependency swap (`core-foundation` -> `objc2-app-kit 0.3.2`, already present in the lock). Kept deliberately surgical: a plain `cargo update -p webbrowser` also **de-unifies** ~8 unrelated `windows-sys` entries (main's lock unifies them at 0.61.2) plus a `prost-build` `heck 0.5.0 -> 0.4.1` downgrade, but that churn is not entailed by this fix, so it is excluded.

## Why now
`cargo-audit` queries the live RustSec advisory DB, so once RUSTSEC-2026-0257 was published the **Dependency Audit** check began failing on `main` and on every open PR against an otherwise-unchanged `Cargo.lock`. This unblocks that check repo-wide (e.g. #357 hit it despite touching zero Rust/lock files).

## Test plan
- [x] `cargo metadata --locked` passes (lock is a valid fixpoint, no re-resolution needed; CI builds with `--locked`)
- [x] `cargo audit` exits 0 locally (RUSTSEC-2026-0257 cleared; only pre-existing non-failing unmaintained/unsound warnings remain)
- [x] `cargo check --workspace --locked --features pentest-platform/desktop-pcap` exits 0 (webbrowser 1.2.4 compiles; only the pre-existing `screenshots` future-incompat warning)
- [ ] CI green (Dependency Audit + Linux/macOS build lanes)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10156,15 +10156,15 @@ dependencies = [
 
 [[package]]
 name = "webbrowser"
-version = "1.2.0"
+version = "1.2.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "fe985f41e291eecef5e5c0770a18d28390addb03331c043964d9e916453d6f16"
+checksum = "62c35be770821a214dbc362fc26908c853e776c0004294d0b10b8a6bad582f94"
 dependencies = [
- "core-foundation 0.10.1",
  "jni 0.22.4",
  "log",
  "ndk-context",
  "objc2 0.6.4",
+ "objc2-app-kit 0.3.2",
  "objc2-foundation 0.3.2",
  "url",
  "web-sys",


ngdevo:
Findings by severity

  🟠 Medium — 1.2.2 carries a regression in the patched code path; use 1.2.4

  Upstream issue #120 (filed 2026-08-05): "Regression in 1.2.2: URL is appended twice to $BROWSER command when it contains no %s" — introduced by commit 31d1b92, i.e. the same commit the advisory cites as the fix. Present in
  1.2.2 and 1.2.3, fixed in 1.2.4 (PR #121), which also ships "Unix: Fix the sequence in which BROWSER fields are substituted."

  Why it matters here: dioxus-desktop 0.7.3 calls webbrowser::open in two places — src/webview.rs:379 (external navigation) and src/app.rs:275 (browser_open IPC). Any Linux user or dev with a plain BROWSER=firefox-style value
  (no %s) gets two tabs per clicked external link.

  Cost of moving to 1.2.4: zero. Its normal deps are byte-identical in shape to 1.2.2's (jni, log, ndk-context, objc2, objc2-app-kit, objc2-foundation, url, web-sys), same MSRV, same ^1.0 compatibility. The lock diff stays 3
  lines — only the checksum differs (62c35be7…2f94). You also pick up 1.2.3's Windows UNC-path fix for free. cargo update -p webbrowser --precise 1.2.4. No open upstream issues against 1.2.4.

  🟠 Medium — a macOS backend rewrite rides along, and nothing runtime-tests it

  1.2.2 also includes #117 "Move to objc2 for macos" — the macOS open path was re-implemented on NSWorkspace/NSRunningApplication instead of core-foundation. That is the real reason for the core-foundation → objc2-app-kit swap
  in the diff, and it is not part of the security fix. The PR describes the swap mechanically ("webbrowser's own dependency swap") without flagging that it represents a rewritten macOS implementation.

  The Check (macOS) lane is cargo check only — nothing executes an open on any platform. Ask for a manual smoke test: click an external link in the desktop app on macOS and on Linux. Note the app's own opener
  (crates/core/src/matrix/auth.rs:296 open_url_in_browser) goes through the open crate, so the Studio auth flow is not on this code path — only webview link handling is. Worth stating explicitly in the PR so the blast radius
  is understood.

  🟡 Low — hand-edited lockfile: correct, but not reviewer-reproducible

  The description makes clear the diff was curated by hand to avoid windows-sys churn. The hand-edit is genuinely correct (verified above), and --locked CI proves cargo accepts it. Still, cargo update -p webbrowser --precise
  <ver> produces exactly this diff without hand-editing — --precise does not re-resolve unrelated packages, so windows-sys is untouched. That turns "trust the author's surgery" into "run the command, git diff is empty." Prefer
  it for lockfile PRs.

  Secondary signal worth a follow-up issue (not this PR): if a plain cargo update -p webbrowser re-pins ~8 windows-sys entries, main's lock is already non-canonical. That latent churn will land on whoever next runs an
  unqualified update.

  🟡 Low — defense-in-depth not evaluated, and the IPC path takes an unfiltered href

  The advisory explicitly recommends the crate's hardened feature for apps that only need http(s). Looking at the two reachable call sites: webview.rs:379 filters to http:///https:///mailto: — and mailto: is precisely the
  "non-HTTP(S) URL whose parsed form retains spaces" shape the advisory describes. app.rs:275 handle_browser_open passes msg.params()["href"] straight to webbrowser::open with no scheme filter at all, reachable from any JS
  running in the webview. For a tool that renders target-derived and LLM-generated content (crates/ui/src/components/chat_panel/), that's the path worth reasoning about.

  Because webbrowser is transitive, enabling hardened means adding a direct workspace dep carrying that feature and relying on feature unification — and the trade-off is that hardened would break mailto: links. I'm flagging
  this as a decision to record (or a follow-up issue), not something to change in this PR.

  ⚪ Info — the audit lane is the actual root cause; consider decoupling it

  .github/workflows/ci.yml:112-119 installs cargo-audit unpinned and runs it against the live RustSec DB on every PR, so any newly published advisory reddens every open PR regardless of content (exactly what hit #357). This PR
  is the right response to this advisory — upgrade rather than suppress, consistent with the documented policy in .cargo/audit.toml. But the structural fix is to make the DB-dependent run a scheduled/nightly blocking gate and
  informational at PR time, or to pin the advisory DB revision per branch. Out of scope here; worth an issue.

  ⚪ Info — PR body nits

  Add the advisory link (GHSA-2ph8-5cr8-hr33) to the body. Also worth stating that no Cargo.toml pin or [patch] is needed: dioxus-desktop's ^1.0 requirement means any future lock regeneration resolves to ≥1.2.2 anyway, so
  there's no silent-revert risk and no .cargo/audit.toml ignore entry to add or remove.

--

 Do not as-is — I'd request the one-line change to 1.2.4 first, then approve.

  The reasoning is that the objection is cheap to resolve: cargo update -p webbrowser --precise 1.2.4 keeps the diff at 3 lines (only the checksum differs from what's there now), needs no other lock churn, and avoids merging a
  release with a confirmed regression in the very function being patched. Landing 1.2.2 now means a second PR later anyway, so there's no time saved.

  If unblocking the audit gate repo-wide is urgent enough that you'd rather not round-trip, approving as-is is defensible — it does clear RUSTSEC-2026-0257 and CI is fully green — but then the 1.2.4 bump should follow
  immediately, not go on a backlog.

  Either way, the two things I'd want the author to acknowledge before merge: the macOS open path was rewritten onto objc2/NSWorkspace (not just a dependency rename), and nothing in CI exercises link-opening at runtime on any
  platform — a manual click-an-external-link check on macOS and Linux is the only real verification.

cchao-devo (approved):
## Review of Strike48-public/pick PR #363 — fix(deps): bump webbrowser to 1.2.2 (RUSTSEC-2026-0257)

  **Verdict:** APPROVE — correct and safe to merge as-is. One non-blocking recommendation that I'd suggest folding into this PR rather than a follow-up: pin **1.2.4** instead of 1.2.2.

  ### Verified independently (not just by reading the diff)

  - **Checksum** `ef62a3d5…c40f` matches crates.io for 1.2.2 exactly, and the removed `fe985f41…` matches 1.2.0. Not yanked.
  - **Dependency swap is exactly right.** webbrowser 1.2.2's real manifest drops the macOS `core-foundation` dep and adds `objc2-app-kit ^0.3`. The edited list is the exact union of
  1.2.2's deps across all targets. `objc2-app-kit 0.3.2` already had a `[[package]]` entry (Cargo.lock:5807), and `core-foundation 0.10.1` keeps 8 other reverse-deps, so nothing is
  orphaned — which is precisely why the surgical edit holds together.
  - **The fix is genuinely in the crate**, not just a version string: `webbrowser-1.2.2/src/unix.rs` tokenizes the `BROWSER` template before substituting the URL, keeping it in one argv
  slot.
  - **`cargo metadata --locked`** → exit 0. **`cargo check --locked --all-targets`** → exit 0, 0 warnings, lock unmodified.
  - **Negative control on the gate:** `cargo audit` against `main`'s lock → **exit 1**, `error: 1 vulnerability found!` naming RUSTSEC-2026-0257 / webbrowser 1.2.0. Against this PR's
  lock → **exit 0**, advisory gone. So the gate genuinely bites and this genuinely clears it. `.cargo/audit.toml` was correctly left untouched — a real fix, not suppression.
  - **Advisory range** is `patched = [">= 1.2.2"]`, so 1.2.2 is sufficient. This is the only webbrowser advisory in the DB.
  - **MSRV** fine: 1.2.2 declares `rust-version = "1.85"` under the repo's 1.92 floor.

  Also confirmed the advisory was **reachable here, not theoretical**: `dioxus-desktop-0.7.3/src/app.rs:275` calls `webbrowser::open(href)` on the raw href from a webview link click with
  no scheme filter, and Pick renders untrusted markdown via `dangerous_inner_html` (`file_viewer.rs:99`) with no sanitizer. Worth fixing on merit, not just to green CI. Container images
  are unaffected — both Dockerfiles build `-p pentest-headless`, which never links webbrowser; exposure is the desktop and Android builds.

  ### Recommendation: prefer 1.2.4 over 1.2.2

  Pinning the exact advisory floor adopts a regression that the security fix itself introduced, on the same platform and code path this PR is hardening.

  Upstream **issue #120**, *"Regression in 1.2.2: URL is appended twice to `$BROWSER` command when it contains no `%s`"* (closed 2026-08-05, fixed in v1.2.4). I confirmed it in the 1.2.2
  source: `browser_env_cmd_arr` pushes the URL at `unix.rs:148` when no `%s` is present, then its caller re-appends it at `unix.rs:108`. Users get two identical browser tabs per
  `open()`. v1.2.4 also carries #121, a further fix to the `BROWSER` substitution sequence, plus the 1.2.3 Windows UNC path fix.

  To be clear on severity: this is **not** a security regression, and it does not block. Its trigger (`BROWSER` set *without* `%s`) is disjoint from the advisory's exploit precondition
  (`BROWSER` *with* `%s`), so no user is exposed to both.

  The reason to just do it now is that it's **free** — 1.2.2 and 1.2.4 have **identical dependency sets**, so it's the same 3-line edit with only the version and checksum differing, and
  1.2.4 is also where a plain `cargo update -p webbrowser` naturally lands, which removes the recurring churn below:

  version = "1.2.4"
  checksum = "62c35be770821a214dbc362fc26908c853e776c0004294d0b10b8a6bad582f94"
  (MSRV still 1.85, not yanked, dep list unchanged from what this PR already writes.)
  
  ### Minor notes
  
  1. **Commit body says "293 other deps unchanged"** — the lock holds 1061 packages (`cargo audit` agrees: "1061 crate dependencies"). Harmless, but it's a wrong number going into
  permanent history.
  2. **The `windows-sys` rationale is right in substance, slightly off in detail.** I reproduced it: `cargo update -p webbrowser --precise 1.2.2` on main's lock differs from this PR's in
  exactly **9** lines — the 8 `windows-sys` edges as stated, plus an unmentioned `prost-build`: `heck 0.5.0 → 0.4.1` **downgrade**. Also the direction is inverted: cargo **de-unifies**
  (main's `0.61.2` → per-dependent `0.48.0/0.52.0/0.59.0/0.60.2`) rather than "re-pins". Main's unification is legal — every consumer's range admits `0.61.2` — so "non-canonically" is
  fair. The exclusion is inert for this PR either way: that unification is pre-existing and byte-identical here, so Windows resolution changes by zero lines. 
  3. The hand-edited lock is a **valid fixpoint, not off-manifold** — the webbrowser block is byte-identical to what cargo generates for `--precise 1.2.2`. Worth recording that
  `--locked` verifies the lock needs no *changes*, not that it's what cargo would *generate*: both locks pass `--locked` unmodified yet differ in 9 lines.
  
  ### Suggested follow-ups (out of scope here)
  
  - **`.cargo/audit.toml`'s stated unblock condition is now materially false for 3 of its 4 paths.** It says removal waits on `screenshots`/`syntect`/`winit` releases, but the quick-xml
  fix landed in the *intermediate* crates and is semver-reachable today: `plist 1.10.0`, `wayland-scanner 0.31.11`, and `xcb 1.7.1` all move to `quick-xml ^0.41`. Only `screenshots
  0.8.10 → libwayshot 0.2.0 → wayland-client 0.30.2 → wayland-scanner 0.30.1` is genuinely blocked. The ignores must stay, but the comment misdirects future reviewers. 
  - **Markdown rendering has no sanitizer.** `render_markdown_raw` (`crates/core/src/rendering.rs`) calls `pulldown_cmark::html::push_html` with no HTML sanitization or href scheme
  allowlist (zero sanitizer crates in the lock), and its output goes to `dangerous_inner_html` on arbitrary viewed-file and chat content. That's the front half of the chain above and
  outlives this dependency bump.

jtomek-strike48:
Thanks for the empirical review — folded in the **1.2.4** recommendation.

Kept it surgical rather than a plain `cargo update`: reverted to the committed lock and hand-edited only webbrowser's version + checksum, so the diff stays the same 3 lines and the ~9-line `windows-sys`/`prost-build` de-unification churn stays excluded (confirmed it's still inert here). Verified 1.2.4 independently:

- crates.io: not yanked, checksum `62c35be7…2f94`, MSRV 1.85 (under the 1.92 floor).
- `cargo metadata --locked` → exit 0 (hand-edited lock is a valid fixpoint; the identical dep set is what lets `--locked` pass unmodified).
- `cargo audit` → exit 0, RUSTSEC-2026-0257 gone; only the pre-existing unmaintained/unsound/yanked warnings remain. `.cargo/audit.toml` untouched.
- `cargo check --workspace --locked --features pentest-platform/desktop-pcap` → exit 0 (webbrowser 1.2.4 actually compiles here; only the pre-existing `screenshots` future-incompat warning).

Minor notes: fixed the wrong "293 other deps" count in the commit body (lock holds 1061 packages) — reworded rather than re-numbered. Your `windows-sys` detail (9 lines incl. the `heck` downgrade, de-unify not re-pin) reproduces exactly; noted in the updated PR description.

Follow-ups (out of scope here, tracking separately): the stale `.cargo/audit.toml` unblock comment (quick-xml fix is semver-reachable for `plist`/`wayland-scanner`/`xcb` today), and the unsanitized `render_markdown_raw` -> `dangerous_inner_html` path — the front half of the injection chain that outlives this bump.

PR #367, opened 2026-08-12
fix(security): sanitize untrusted markdown before dangerous_inner_html (#365)

### LINKED ISSUE #365: security: sanitize markdown before dangerous_inner_html (file viewer + chat panel)
## Problem

`render_markdown_raw` (`crates/core/src/rendering.rs:162`) renders markdown to HTML via `pulldown_cmark::html::push_html` (`:209`) with **no HTML sanitizer and no href scheme allowlist** (zero sanitizer crates in the lock). The parser is built with only `ENABLE_TABLES | ENABLE_STRIKETHROUGH | ENABLE_TASKLISTS`, so raw inline/block HTML in the source (`<script>`, `<iframe>`, `onerror=`) passes straight through as `Event::Html`, and link hrefs (`javascript:`, `data:`, `file:`) are emitted unfiltered.

Its output is written to `dangerous_inner_html` on **arbitrary, attacker-influenceable content** at two sinks:

- `crates/ui/src/components/file_browser/file_viewer.rs:99` — any viewed markdown file.
- `crates/ui/src/components/chat_panel/render.rs:50,70` — chat/agent content (HTML from `push_html` at `render.rs:230`).

This is the front half of the injection chain @cchao-devo traced while reviewing #363: untrusted markdown → raw HTML → webview, where a rendered link click reaches `webbrowser::open` (the RUSTSEC-2026-0257 surface the bump only partially closes). It outlives the dependency bump.

## Impact

Rendered HTML can carry `<script>`, event-handler attributes, and `javascript:`/`data:`/`file:` hrefs, executed in the desktop webview context. Chat/agent content is the higher-risk vector: a scanned target's response reflected into chat markdown is attacker-controlled. Container/headless images don't render this UI, so exposure is **desktop + Android**.

## Proposed fix (needs a decision before implementing)

Sanitize at the single rendering choke point (`render_markdown_raw`), consistent with the connector's choke-point-over-per-site pattern:

- **Option A — add a maintained sanitizer crate (`ammonia`)**: robust, well-tested allowlist covering tags, attributes, and URL schemes. Adds one dependency (needs approval).
- **Option B — hand-rolled**: drop raw-HTML `Event::Html`/`Event::InlineHtml` events and apply an href scheme allowlist (`http`/`https`/`mailto`). No new dependency, narrower coverage, more maintenance risk.

## Acceptance criteria

- [ ] Untrusted markdown cannot inject executable HTML/JS or non-allowlisted URI schemes into the webview.
- [ ] Both sinks (file viewer + chat panel) are covered by the single choke-point fix — verify the consumer trace.
- [ ] Regression tests feed adversarial markdown (`<script>`, `onerror=` attr, `javascript:` href, `<iframe>`) and assert neutralization; guard-verified (tests go red if the sanitizer is removed).
- [ ] Security review before merge.

jtomek-strike48:
## Summary
- Fixes **#365**: two markdown renderers wrote `pulldown_cmark` output to Dioxus `dangerous_inner_html` with no sanitizer — a stored-XSS sink reachable from viewed files and (highest-risk) chat/agent content that reflects tool + scanned-target output.
- Adds one shared choke point `pentest_core::rendering::sanitize_markdown_event` and routes **both** renderers through it (security-by-construction, **no new dependency**): the core file renderer and the chat panel's own `render_markdown` (which was previously missed — a separate, unprotected renderer).
- Filter: drop raw/inline HTML events; scheme-allowlist link/image destinations (default-deny — `http`/`https`/`mailto`/relative pass; `javascript:`/`data:`/`file:`/protocol-relative `//` rejected), trimming leading control chars so `\tjavascript:` can't smuggle a scheme.
- Syntect code highlighting and ` ```mermaid ` blocks are preserved (trusted highlight HTML is injected separately and bypasses the filter by construction).

## Why Option B (event filter) not Option A (`ammonia`)
Discussed on the issue: B closes the full `push_html` emit surface with no dependency; a post-hoc sanitizer would guard an already-closed door. `ammonia` is recorded as a future defense-in-depth option if we ever need to *render* trusted HTML rather than strip it. See the decision note on #365.

## Reviews (pick-reviewer + security-reviewer)
- **pick-reviewer: APPROVE**, mutation-verified guards, CI-scope confirmed, no additional sinks, highlighting/mermaid preserved.
- **security-reviewer** findings triaged empirically:
  - Its CRITICAL (HTML file-viewer iframe, `file_viewer.rs:105`) is a **false positive** — that iframe uses `sandbox="allow-same-origin"` with **no `allow-scripts`**, so scripts don't execute, and `"` is escaped so there's no attribute breakout. Also a pre-existing, non-markdown path (out of scope). pick-reviewer independently reached the same conclusion.
  - NBSP / percent-encoding "bypass" concerns were **run through the renderer** and are blocked by default-deny.
  - Protocol-relative `//host` was confirmed real and is now rejected in this PR.

## Residual risk (out of scope, noted for follow-up)
`https://` images/links are allowed by design, so a remote image can still act as a tracking/exfil beacon on view. Blocking all remote resource loads is a separate product/design decision, not part of closing the XSS class.

## Test plan
- [x] `cargo test -p pentest-core --lib rendering` — 14 tests (each vector + safe-content preservation).
- [x] Chat sink guarded in `pentest-ui` (`chat_panel::render::tests`); runs on the Linux CI lane (`--workspace --features pentest-platform/desktop-pcap`). macOS lane is core-scoped, where the shared `sanitize_markdown_event` logic is fully tested.
- [x] Guard-verified: neutering each defense (raw-HTML drop, scheme allowlist, protocol-relative) turns the relevant tests red; restoring makes them green.
- [x] `cargo clippy --workspace --features pentest-platform/desktop-pcap --all-targets -- -D warnings` clean; `cargo fmt --all --check` clean.
- [ ] CI green.

Addresses #365. (The XSS *class* is fully closed once the chart-processor
sink lands in the companion PR; that PR carries `Closes #365`, so this issue
stays open until both merge.)

### DIFF
--- a/crates/core/src/rendering.rs
+++ b/crates/core/src/rendering.rs
@@ -156,9 +156,108 @@ pub fn is_markdown(path: &str) -> bool {
 // Markdown rendering
 // ---------------------------------------------------------------------------
 
+/// URL schemes permitted in rendered-markdown link and image destinations.
+///
+/// This is an allowlist (default-deny), not a blocklist: any explicit scheme
+/// not listed here — `javascript:`, `data:`, `file:`, `vbscript:`, … — is
+/// rejected. Relative (scheme-less) URLs are always allowed.
+const ALLOWED_URL_SCHEMES: [&str; 3] = ["http", "https", "mailto"];
+
+/// Return `true` if `url` is safe to use as a link or image destination.
+///
+/// Relative URLs (no scheme) are allowed. A URL with an explicit scheme is
+/// allowed only if that scheme is in [`ALLOWED_URL_SCHEMES`]. Leading ASCII
+/// whitespace and control characters are stripped first, the way browsers do
+/// when parsing a scheme, so `"\tjavascript:…"` cannot smuggle a rejected
+/// scheme past the check. Non-ASCII/other-scheme prefixes (e.g. a leading
+/// U+00A0 before `javascript:`) fall through to the allowlist and are rejected
+/// by default-deny.
+///
+/// Protocol-relative URLs (`"//host/…"`) are rejected: they resolve to a remote
+/// origin, so they are treated as unsafe rather than as a scheme-less relative
+/// path (#365 review — avoids silent remote fetches/exfiltration beacons).
+fn is_safe_url(url: &str) -> bool {
+    let trimmed = url.trim_start_matches(|c: char| c.is_ascii_whitespace() || c.is_control());
+    if trimmed.starts_with("//") {
+        return false;
+    }
+    // The scheme is the text before the first ':' — but only when that ':'
+    // appears before any '/', '?', or '#', which would instead mark a
+    // relative/path reference (e.g. "/a?b#c" has no scheme).
+    match trimmed.find([':', '/', '?', '#']) {
+        Some(idx) if trimmed.as_bytes()[idx] == b':' => {
+            let scheme = &trimmed[..idx];
+            ALLOWED_URL_SCHEMES
+                .iter()
+                .any(|s| scheme.eq_ignore_ascii_case(s))
+        }
+        _ => true,
+    }
+}
+
+/// Make a single markdown event safe to feed to `dangerous_inner_html`.
+///
+/// Returns `None` for raw/inline HTML events (dropped so untrusted markdown
+/// cannot inject `<script>`, event-handler attributes, `<iframe>`, …). Link and
+/// image destinations are scheme-allowlisted via [`is_safe_url`]; a rejected
+/// destination becomes an inert empty string. Every other event passes through
+/// unchanged.
+///
+/// This is the shared choke point for all markdown rendering in the app (#365):
+/// [`render_markdown_raw`] and the chat panel's renderer both route through it,
+/// so the sanitization cannot drift between the two. Renderers that also build
+/// their own trusted HTML (e.g. syntect code highlighting) inject it separately
+/// and must not pass it through this function.
+pub fn sanitize_markdown_event(
+    event: pulldown_cmark::Event<'_>,
+) -> Option<pulldown_cmark::Event<'_>> {
+    use pulldown_cmark::{CowStr, Event, Tag};
+    match event {
+        Event::Html(_) | Event::InlineHtml(_) => None,
+        Event::Start(Tag::Link {
+            link_type,
+            dest_url,
+            title,
+            id,
+        }) => {
+            let dest_url = if is_safe_url(&dest_url) {
+                dest_url
+            } else {
+                CowStr::Borrowed("")
+            };
+            Some(Event::Start(Tag::Link {
+                link_type,
+                dest_url,
+                title,
+                id,
+            }))
+        }
+        Event::Start(Tag::Image {
+            link_type,
+            dest_url,
+            title,
+            id,
+        }) => {
+            let dest_url = if is_safe_url(&dest_url) {
+                dest_url
+            } else {
+                CowStr::Borrowed("")
+            };
+            Some(Event::Start(Tag::Image {
+                link_type,
+                dest_url,
+                title,
+                id,
+            }))
+        }
+        other => Some(other),
+    }
+}
+
 /// Render markdown to raw HTML (no wrapper div).
 ///
-/// Fenced code blocks are syntax-highlighted via syntect.
+/// Fenced code blocks are syntax-highlighted via syntect. Untrusted HTML and
+/// unsafe link/image schemes are stripped via [`sanitize_markdown_event`].
 pub fn render_markdown_raw(content: &str) -> String {
     use pulldown_cmark::{CodeBlockKind, CowStr, Event, Options, Parser, Tag, TagEnd};
 
@@ -179,7 +278,6 @@ pub fn render_markdown_raw(content: &str) -> String {
                 in_code_block = true;
                 code_lang = lang.to_string();
                 code_buf.clear();
-                continue;
             }
             Event::End(TagEnd::CodeBlock) if in_code_block => {
                 in_code_block = false;
@@ -193,16 +291,21 @@ pub fn render_markdown_raw(content: &str) -> String {
                     lang = html_escape(&code_lang),
                     code = highlighted,
                 );
+                // Trusted, locally-built HTML — intentionally bypasses the
+                // markdown sanitizer that all other events flow through.
                 events.push(Event::Html(CowStr::from(html)));
-                continue;
             }
             Event::Text(ref text) if in_code_block => {
                 code_buf.push_str(text);
-                continue;
             }
-            _ => {}
+            // Security (#365): every other event goes through the shared filter,
+            // which drops raw/inline HTML and scheme-allowlists link/image URLs.
+            other => {
+                if let Some(safe) = sanitize_markdown_event(other) {
+                    events.push(safe);
+                }
+            }
         }
-        events.push(event);
     }
 
     let mut html_output = String::new();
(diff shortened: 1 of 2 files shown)


ngdevo:
Verdict: hold for one small commit, then merge. Strong net improvement, well tested. The one in-diff defect is a 3-line fix + 2 tests inside the function this PR introduces. It's not a regression, so merging as-is would still beat main if you need CI unblocked.

Verified correct: all three markdown call sites are covered (file_viewer.rs:99, core/file_browser/render.rs:385, chat_panel/render.rs:40,66); the continue → fall-through refactor is behaviour-preserving (fence arms bind by ref); the CI-scope claim holds; and the file_viewer.rs:105 iframe CRITICAL really is a false positive — sandbox="allow-same-origin" without allow-scripts blocks execution, and &quot; can't close the srcdoc attribute since the tokenizer scans for a raw ". Worth noting the severity is higher than the PR states: core/file_browser/render.rs emits via AppPageResponse::html, i.e. into a Matrix-served https origin, not just a desktop webview.

High

1. Backslash defeats the protocol-relative guard. Per WHATWG URL, \ is treated as / for special schemes, so both of these pass is_safe_url as "relative" and resolve remote:
    - [x](/\evil.com/log?c=1) → /\evil.com/… → https://evil.com/log?c=1
    - [x](\\evil.com/log) → no :// found at all → hits _ => true

Fix: normalize \→/ on trimmed before the starts_with("//") check. Add both to is_safe_url_rejects_protocol_relative.

2. A second unsanitized sink sits downstream in the chat path — crates/ui/src/assets/chart_processor.js (injected at chat_panel/mod.rs:256) post-processes exactly the code blocks this PR preserves. Lines 46/49/82 do div.innerHTML = 'Mermaid error: ' + err.message, and mermaid/jison errors embed a snippet of the offending input, so a malformed ```mermaid fence containing  reaches innerHTML. Line 66/73 does chart.setOption(JSON.parse(code)), and ECharts renders tooltip.formatter string templates as HTML → {"tooltip":{"formatter":""}} is XSS on hover. Pre-existing, out of diff — but #365's stated class isn't closed until this is.

Medium

3. Rejected destinations become href="" / src="". <a href=""> navigates to the current document on click (re-submits the current query in the served browser);  still triggers a document re-fetch in some engines. Prefer dropping the element and keeping the child text. Also no user-visible signal a link was blocked.
4. No test for entity/percent-encoded schemes. [x](&#106;avascript:alert(1)) should already be caught since CommonMark decodes entities in destinations, but nothing proves pulldown-cmark 0.12.2 decodes before the filter sees it. Add that plus [x](%6Aavascript:alert(1)). The security-reviewer raised percent-encoding; a test is the cheap answer.
5. Adjacent supply chain: chart_processor.js pulls mermaid@11 / echarts@5 from jsdelivr, floating majors, no SRI, into the app origin.

Low — document the option coupling on sanitize_markdown_event (it only inspects Tag::Link/Tag::Image; enabling ENABLE_HEADING_ATTRIBUTES or math would need a re-review); the syntect trusted-HTML bypass is enforced only by a  comment.

jtomek-strike48:
Thanks for the thorough pass — all five findings are addressed. Split across two PRs so the markdown sanitizer stays tight and the chart-processor XSS lands as its own reviewable unit.

### This PR (#367) — markdown sanitizer

- **#1 (backslash protocol-relative):** fixed — `is_safe_url` now folds `\`→`/` before the protocol-relative + scheme checks, so `/\host` and `\\host` are rejected. Both added to `is_safe_url_rejects_protocol_relative`, plus an end-to-end render test.
  - One empirical correction on severity: I checked what actually reaches the DOM. pulldown-cmark's `push_html` percent-encodes the backslash to `%5C`, and browsers do **not** fold `%5C`→`/` (confirmed against the WHATWG URL algorithm). So `[x](/\evil.com)` rendered `href="/%5Cevil.com…"` (same-origin) and `[x](\\evil.com)` collapsed to `%5Cevil.com` — neither reached a *remote* origin through this renderer. The `is_safe_url` gap was real and is now closed by construction (rather than leaning on that downstream encoding), but it wasn't a live exfil path — flagging so the fix isn't oversold.
- **#3 (inert `href=""`):** fixed — replaced the stateless per-event filter with a small stateful `MarkdownSanitizer` that *unwraps* a rejected link/image (drops the element + its matching close tag, keeps the child text) instead of emitting `href=""`/`src=""`. Shared by both renderers so they can't drift.
- **#4 (encoded schemes):** added `neutralizes_entity_and_percent_encoded_schemes`. Verified against pulldown-cmark 0.12.2: HTML entities in a destination *are* decoded (`&#106;avascript:`→`javascript:`→rejected); percent-encoding is not (`%6Aavascript`→unknown scheme→rejected). Both collapse to an inert empty href.
- **Low (option coupling):** documented on `MarkdownSanitizer` — only `Tag::Link`/`Tag::Image` are inspected, so enabling `ENABLE_HEADING_ATTRIBUTES`/math would need a re-review.

Verified (Rust 1.92): core `rendering` 17/17, chat sink on the CI lane (`--workspace --features pentest-platform/desktop-pcap`) 8/8, `clippy --workspace` clean, `fmt` clean. Guard-checked by neutering each defense — the relevant tests go red.

### #2 + #5 (chart post-processor) → #371

Different subsystem, so it's in #371: mermaid/echarts errors now render via `textContent` (not `innerHTML`); ECharts options are walked to drop every `formatter` and force `renderMode:'richText'`; mermaid `securityLevel:'strict'` is explicit; and mermaid/echarts are pinned (`11.16.1`/`5.6.0`) with SRI + `crossorigin`.
- While hardening the ECharts option walk I checked the `graphic type:'html'` idea against the `echarts@5.6.0` bundle — there's no such graphic element type (every `"html"` literal is the tooltip `renderMode`; graphic text renders on canvas), so it isn't a sink. Formatter-as-array *is* covered (all formatters dropped regardless of type).

I moved `Closes #365` to **#371** so the issue stays open until the chart sink lands too — you're right the class isn't fully closed until then. `title.link`/remote-image beacons are noted there as out-of-scope follow-ups.

cchao-devo (approved):
## Review of Strike48-public/pick PR #367 — fix(security): sanitize untrusted markdown before dangerous_inner_html (#365)
  **Verdict: APPROVE — merge-ready at `1d6dc92`.** Both in-diff defects from the prior pass are genuinely fixed, no XSS bypass survived ~95 adversarial vectors, and all gates are green. One
  1-line change is strongly recommended (items A+B share a single fix) but I would not block on it; the rest are test-durability and doc accuracy.
  
  ### Gates — run with CI's exact invocations on `1d6dc92`
  
  - `cargo fmt --all -- --check` → 0
  - `cargo clippy --workspace --locked --features pentest-platform/desktop-pcap --all-targets -- -D warnings` → 0, zero warnings
  - `cargo test -p pentest-core --lib rendering` → **17/17**
  - `cargo test --workspace --locked --features pentest-platform/desktop-pcap` → green except one **pre-existing, unrelated** failure: `all_tools_registered`
  (`crates/core/tests/tool_integration.rs`, `autopwn_crack` platform-gated on macOS; outside this 2-file diff).
  - **"Do the new tests actually run in CI?"** — yes. Under CI's own workspace command, all three `components::chat_panel::render::tests::chat_markdown_*` execute in the `pentest_ui` lib
  target. Note `cargo test -p pentest-ui` alone does **not** compile (8 `E0425`/`E0433`, `liveview`-gated items arrive only via workspace feature unification), so local verification must
  use the `--workspace` form. 
  - Suggest adding `--no-fail-fast` to the test job: running CI's command verbatim, the `tool_integration` failure aborted before `pentest_ui` was reached, so the chat guards silently did
  not execute.
  
  ### Follow-up commits — verified, all five prior findings resolved
  
  | Prior finding | Status |
  |---|---|
  | Backslash protocol-relative | **Fixed correctly.** `\\host` and `/\host` now denied. I checked the normalization doesn't over-reach: single `\evil.com/x` stays allowed, which is right —
  WHATWG folds it to same-origin `/evil.com/x`, not `//evil.com`. `docs\readme.md` still resolves as a relative path. |
  | Inert `href=""` / `src=""` | **Fixed** by unwrapping. Verified tag-balanced across 40 vectors (nested image-in-link, link-in-alt, ref/collapsed/shortcut/autolink, table cells,
  blockquote+list, 3-deep image nesting) — zero imbalance, no orphan `</a>`. |
  | Encoded schemes | Test added; behavior verified (`&#106;avascript:` decoded-then-denied, `%6A` denied by default-deny). |
  | Parser-option coupling | **Documented, not enforced** — see C. |
  | `chart_processor.js` XSS | Correctly split to **#371**. |
  
  For the record on the reload/state-loss rationale in the new doc comment: it was real only for the **web/liveview** shell. On desktop, `dioxus-desktop-0.7.3/src/webview.rs:364-383` blocks
  all post-load navigation, so an empty `href` was inert there. The unwrap is still the better outcome (no misleading link affordance), so this doesn't change the fix — just scoping the
  "why".
  
  ### Remaining findings
  
  **A. [Important] Dropping an HTML block silently swallows adjacent plain text.** Measured:
  `render_markdown_raw("<div>a</div>\nvisible text after\n")` → `""`. The sibling text line is gone, not just the HTML. `<table>…</table>` → `""`; `<details><summary>Click me</summary>`
  loses the summary. In a pentest tool whose file viewer and chat pane carry target-derived evidence, that is unlogged evidence loss.
  
  **B. [Important] Dropping inline HTML fuses neighbouring text into wrong values.** Measured:
  `10.0.0.1<br>22 open` → `10.0.0.122 open`; `Contact:<br>alice@example.com<br>+1-555-0100` → `Contact:alice@example.com+1-555-0100`. This is corruption, not just loss — the operator reads
  a fabricated address. A `<p align="center"></p>` README header vanishes silently.
  
  Both fix with one line, and it is equally safe (`push_html` escapes `Event::Text`):
  ```rust
  Event::Html(h) | Event::InlineHtml(h) => Action::Text(h),  // keep as escaped text, don't delete
  I verified the escaped-text variant blocks every vector the current drop blocks.
  
  C. [Minor] The two renderers still drift on parser options. rendering.rs uses TABLES | STRIKETHROUGH | TASKLISTS; chat_panel/render.rs uses STRIKETHROUGH | TABLES — no TASKLISTS.
  Measured: - [x] done renders a checkbox in the file viewer and literal [x] done in chat. The comment "the two cannot drift apart" is true of sanitization but not of rendering. This
  matters more now that the new doc comment correctly says enabling attribute-emitting options requires re-reviewing the sanitizer: I confirmed that with ENABLE_HEADING_ATTRIBUTES on, # H 
  {#a onmouseover="alert(1)"} passes straight through the Pass catch-all as <h1 id="a" onmouseover=…>. Hoisting pub fn markdown_options() -> Options next to the sanitizer makes the coupling
  structural instead of a comment.
  
  D. [Minor] dropped_scopes.pop().unwrap_or(false) fails open. An unmatched close tag emits </a>. Not reachable today — and I checked the one seam that could cause it: render_markdown_raw
  hides CodeBlock/Text events from the sanitizer, so the stream it sees has holes, technically violating the new "call on the full event stream in order" contract. Safe because code blocks
  are block-level and can never split an inline link scope; verified across 10 fence/link interleavings (fence between rejected links, unterminated fence, fence in list/blockquote, nested
  fences, indented code) — all balanced. Worth a debug_assert! and one sentence in the doc noting why the hidden events are safe.
  
  E. [Nit] Statefulness raises the misuse bar. MarkdownSanitizer::process is pub; a caller must now build one per document and feed the whole stream. Both in-tree callers do this correctly.
  An iterator-adapter wrapper (pub fn sanitize<'a>(p: impl Iterator<Item=Event<'a>>) -> impl Iterator<Item=Event<'a>>) would make the correct usage the only usage. 
  
  F. [Test gap] The trusted-HTML carve-out is untested. rendering.rs hand-builds <pre><code class="language-{lang}"> and injects it as Event::Html, explicitly bypassing the sanitizer; the
  fence info string is attacker-controlled and lands inside a quoted attribute. It is escaped today (verified: class="language-rust&quot; onmouseover=&quot;alert(1)"), but its only defense
  is the general-purpose html_escape. Deleting .replace('"', "&quot;") from that helper leaves all 17 tests green while emitting a working onmouseover= — i.e. CI cannot see a regression in
  the one path this PR deliberately carves out. One assert closes it.
  
  G. [Test gap] Nothing exercises the control-flow restructure's actual risk surface — no test renders an indented (non-fenced) code block or an unterminated fence through
  render_markdown_raw. I verified both are correct (the refactor is behavior-preserving; independently confirmed byte-for-byte over a 57-input corpus with an identity sanitizer, 0 diffs),
  but a future regression there would be silent. Two asserts.
  
  H. [Doc] The //host rationale is still overstated. "avoids silent remote fetches/exfiltration beacons" — https://evil.com/p.png?c=leak is allowlisted, so beacons are permitted by design.
  Note also crates/ui/src/theme.rs:686-694 runs the opposite policy for custom CSS (blocks http(s)://, allows data:image/). Suggest rewording to "protocol-relative URLs are not relative
  paths" and flagging the deliberate divergence.

  I. [Body accuracy] The sink trace is incomplete. The PR says "two sinks (file viewer + chat panel)". There is a third consumer — crates/core/src/file_browser/render.rs:385 →
  render_file_view → AppPageResponse::html, i.e. an HTTP-served response body, not a webview dangerous_inner_html. The fix covers it transitively; the write-up never mentions it, and it
  makes the original severity higher than stated, not lower. 
  
  Follow-up, out of this diff's scope

  crates/ui/src/components/file_browser/file_viewer.rs:105 — the iframe srcdoc preview. Scripts are genuinely blocked (sandbox="allow-same-origin" with no allow-scripts, confirmed), and
  "-only escaping does contain the attribute. But sandbox has no network-blocking directive, so remote subresources in a viewed HTML artifact still fetch — an OPSEC leak when opening a
  captured target artifact. And the entire XSS-safety rests on the absence of one word in a string literal, with no test. Suggest sandbox="" + a CSP + a test asserting allow-scripts never
  appears. Happy to file this.

cchao-devo (approved):
LGTM

PR #369, opened 2026-08-13
ci: harden gitleaks binary download (pin version + retry)

jtomek-strike48:
## Summary
Fixes the flaky **Gitleaks** check without changing the scanner or adding any secret.

The job resolved gitleaks from `releases/latest` at run time and piped `curl … | tar`, which failed intermittently on a transient network error (`curl: (56) Connection died, tried 5 times`) — blocking PRs on a step that never reached scanning.

- **Pin `GITLEAKS_VERSION=8.30.1`** — reproducible, and removes the extra `releases/latest` API call that was a second network dependency.
- **Retry-hardened `curl`** — `--retry 5 --retry-all-errors --retry-delay 5 --connect-timeout 30`, downloading to a file instead of piping to `tar`, so a partial transfer is retried rather than corrupting the extract.

## Why not the official `gitleaks-action`
The official action **requires a `GITLEAKS_LICENSE` key for organization-owned repos** (free, but must be requested at gitleaks.io and stored as a secret). We don't have one, and without it the action hard-fails before scanning — so the license-free CLI approach stays. Supersedes #368 (which switched to the action).

## Behavior preserved
- Scan rules unchanged — `.gitleaks.toml` (the test-JWT allowlist) is still auto-loaded by `gitleaks detect --source .`.
- Same full-tree/history scan on every push and PR; `fetch-depth: 0` retained.

## Optional further hardening (not in this PR)
Could add SHA-256 checksum verification of the downloaded tarball (gitleaks publishes `*_checksums.txt`) for supply-chain integrity — happy to add if wanted.

## Test plan
- [x] `python3` YAML parse — workflow valid; `gitleaks` job = `checkout@v5` → install (pinned + retry) → `gitleaks detect`.
- [ ] CI green (this PR's own Gitleaks run exercises the new install path).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -97,15 +97,20 @@ jobs:
         with:
           fetch-depth: 0
       - name: Install gitleaks
-        env:
-          GH_TOKEN: ${{ github.token }}
         run: |
-          # `gh api` gives us a proper JSON parse via --jq so we don't have to
-          # regex our way through GitHub's compact JSON, which used to catch
-          # the whole `"tag_name":"v8.30.1","draft":...` blob and produce a
-          # malformed download URL.
-          GITLEAKS_VERSION=$(gh api repos/gitleaks/gitleaks/releases/latest --jq '.tag_name' | sed 's/^v//')
-          curl -sSfL "https://github.com/gitleaks/gitleaks/releases/download/v${GITLEAKS_VERSION}/gitleaks_${GITLEAKS_VERSION}_linux_x64.tar.gz" | tar xz
+          # Pinned instead of resolved from releases/latest at run time: that
+          # extra API call was a failure point, and an unpinned tool version is
+          # non-reproducible. Bump deliberately.
+          GITLEAKS_VERSION=8.30.1
+          # --retry with backoff + --retry-all-errors survives the transient
+          # "curl: (56) Connection died" that flaked this step before. Download
+          # to a file (not | tar) so a partial transfer fails the retry rather
+          # than corrupting the extract.
+          curl --fail --show-error --location \
+            --retry 5 --retry-all-errors --retry-delay 5 --connect-timeout 30 \
+            -o gitleaks.tar.gz \
+            "https://github.com/gitleaks/gitleaks/releases/download/v${GITLEAKS_VERSION}/gitleaks_${GITLEAKS_VERSION}_linux_x64.tar.gz"
+          tar -xzf gitleaks.tar.gz gitleaks
           sudo mv gitleaks /usr/local/bin/
       - run: gitleaks detect --source . --verbose
 


ngdevo:
Verdict: immediately mergeable. Strict improvement, nothing is a regression, and it unblocks PRs. v8.30.1 is the current latest (2026-03-21) and gitleaks_8.30.1_linux_x64.tar.gz is a real asset, so the pin and URL are good.

Dropping GH_TOKEN is right, and killing the pipe removes a real hazard — without pipefail, curl's exit status was masked by tar's.

Medium

1. No integrity verification — sudo mv of an unverified binary into /usr/local/bin. gitleaks publishes gitleaks_8.30.1_checksums.txt; with the version fixed, a hardcoded sha256 is stable and free. Pinning the version without the bytes is half the hardening. The author offered this — take them up on it.
2. The pin freezes gitleaks' built-in rule set, so detection coverage silently stops improving. Needs a bump mechanism (Renovate regex manager) or "bump deliberately" becomes "pinned forever".

Low — tarball downloads into $GITHUB_WORKSPACE (the scan target) and is never removed; harmless under git-mode detect but breaks if anyone adds --no-git (use $RUNNER_TEMP). --retry-all-errors + --fail retries 4xx, so a renamed asset burns 25s and buries the 404 (add --retry-max-time 120). sudo is avoidable via $GITHUB_PATH. Add explicit shell: bash + set -euo pipefail.

High — adjacent, not in this diff: .gitleaks.toml has no [extend] block, and gitleaks only inherits its built-in rules when the config sets [extend] useDefault = true (confirmed against the upstream annotated config) — a config with [[rules]] and no [extend] replaces the defaults. This repo's config defines one rule, jwt-test-tokens, which matches two literal test JWTs and allowlists three paths. So the Gitleaks job is very likely scanning for those two strings and nothing else, which would explain why it's always green. The intent is also inverted — that belongs in a top-level [allowlist]. Verify with gitleaks detect --source . --log-level debug and check the loaded rule count. Don't block #369 on it, but it's more consequential than anything in the diff.

ngdevo (approved):
See review comments.

cchao-devo (approved):
## Review of pick PR #369 — ci: harden gitleaks binary download (pin version + retry)
  **Verdict:** APPROVE — merge as is.

  Strict improvement over `main`, no regressions, and it unblocks PRs flaked by the transient `curl: (56)` failure. I verified the one real correctness risk empirically (details below). One
  optional supply-chain follow-up.

  ### What I verified (empirically, not by eye)
  - **Pinned asset is real:** `gitleaks_8.30.1_linux_x64.tar.gz` → HTTP 200, ~8.2 MB.
  - **Named-member extraction is correct:** the tarball's top-level members are exactly `LICENSE`, `README.md`, `gitleaks`; `tar -xzf gitleaks.tar.gz gitleaks` extracts the x86-64 ELF
  binary. The new form is actually *more* precise than the old `tar xz` (which unpacked LICENSE/README too).
  - **Error handling is sound:** GitHub's default `run:` shell is `bash --noprofile --norc -eo pipefail`. `curl --fail` exits non-zero on HTTP errors, and `tar` / `sudo mv` failures
  propagate — every failure now fails the step. Removing the `curl | tar` pipe means a partial transfer is retried rather than corrupting the extract, exactly as the description claims.
  - **Sibling paths:** gitleaks is installed in exactly one place across `.github/` — this job — so nothing else needs the same fix. Dropping `gh api releases/latest` also removes the
  `GH_TOKEN` dependency and a second network hop.
  - **Scan behavior preserved:** `gitleaks detect --source .` is unchanged, so `.gitleaks.toml` still auto-loads and the full-tree/history scan (`fetch-depth: 0`) is intact.

  ### Optional (non-blocking)
  1. **Checksum verification.** You've pinned the *version* but not the *artifact* — a re-tagged/replaced release asset would still be trusted. Since gitleaks publishes
  `gitleaks_8.30.1_checksums.txt`, verifying the SHA-256 completes the supply-chain story cheaply. This is the "optional further hardening" you already noted; I'd endorse folding it in now
  that the version is pinned. (Not a regression — the old code had no checksum either.)
  2. **Nit:** `--retry-all-errors` will also retry permanent failures (e.g. a bad version → 404) 5×5 s ≈ 25 s before failing. Harmless. Consider `--max-time` to bound a
  stalled-but-connected transfer, since `--connect-timeout` only covers the connect phase.

PR #370, opened 2026-08-13
fix(ui): refresh sidebar chat history when conversations change

jtomek-strike48:
## Summary
- The left sidebar's recent-conversations list loaded once at connect and never refreshed. Its fetch `use_effect` (`sidebar.rs`) depended only on `api_url`/`auth_token` — both stable for the whole session — and the fetch runs inside a `spawn`, whose reads don't subscribe the effect. New conversations and `updated_at` bumps never appeared until a reload, while the chat panel's own (disjoint) history dropdown updated normally, so the sidebar looked stuck.
- Added a shared `ConversationRefresh(Signal<u64>)` trigger provided by `AppLayout` — the common ancestor of both `Sidebar` and `ChatPanel` in both app roots. `ChatPanel` bumps it on each `create_conversation` success (new conversation appears immediately) and at turn completion inside `poll_and_update` (one choke point covering all six poll call-sites, so replies/`updated_at` bumps reorder the list). `Sidebar`'s fetch effect reads the trigger, so a bump re-runs the fetch and the list stays current. No self-retrigger: the effect never reads `recent_convos`.

## Test plan
- [x] `cargo check --workspace --locked --features pentest-platform/desktop-pcap` — clean
- [x] `cargo clippy --workspace --locked --features pentest-platform/desktop-pcap -- -D warnings` — clean
- [x] `pentest-ui` unit tests — 95 passed / 0 failed
- [ ] Manual: send a message / receive an agent reply and confirm the sidebar list updates (new conversation appears, ordering/timestamp refresh) without a page reload. This is Dioxus reactive wiring with no behavioural logic to unit-test, and the crate has no render-test harness, so a component-level test is intentionally omitted rather than added as a tautology.

### DIFF
--- a/crates/ui/src/components/app_layout.rs
+++ b/crates/ui/src/components/app_layout.rs
@@ -8,6 +8,19 @@ use super::pane_boundary::PaneBoundary;
 use super::sidebar::{NavPage, Sidebar};
 use super::status_bar::StatusBar;
 
+/// Shared "conversations changed" trigger, distributed via Dioxus context.
+///
+/// The chat panel bumps this whenever it creates a conversation or an agent
+/// reply lands (which moves a conversation's `updated_at`). The sidebar's
+/// recent-conversations effect subscribes to it so the list re-fetches instead
+/// of staying frozen at its initial post-connect load. The wrapped value is an
+/// opaque monotonically-increasing counter — only its *change* matters.
+///
+/// `Signal` is `Copy`, so this newtype is `Copy` too and can be moved into the
+/// spawned tasks that perform the bump.
+#[derive(Clone, Copy)]
+pub struct ConversationRefresh(pub Signal<u64>);
+
 /// Shared layout wrapper used by all three apps.
 ///
 /// The desktop-header serves as the universal page header on both desktop and
@@ -36,6 +49,11 @@ pub fn AppLayout(
     // header actions. ChatPanel writes Some(ChatHeaderCtx) when mounted in
     // full-page mode; we render those actions in the desktop header bar.
     let chat_header_ctx: Signal<Option<ChatHeaderCtx>> = use_context_provider(|| Signal::new(None));
+
+    // Provide the shared conversation-refresh trigger. ChatPanel (a descendant
+    // via `children`) bumps it; Sidebar (a direct child) reads it to re-fetch
+    // its recent-conversations list. See [`ConversationRefresh`].
+    use_context_provider(|| ConversationRefresh(Signal::new(0u64)));
 
     let on_close = move |_: ()| {
         sidebar_open.set(false);
--- a/crates/ui/src/components/chat_panel/mod.rs
+++ b/crates/ui/src/components/chat_panel/mod.rs
@@ -21,6 +21,7 @@ use pentest_core::terminal::TerminalLine;
 use std::collections::HashMap;
 use std::sync::Arc;
 
+use super::app_layout::ConversationRefresh;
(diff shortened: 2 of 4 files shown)


cchao-devo (approved):
## Review of pick PR #370 — fix(ui): refresh sidebar chat history when conversations change
  **Verdict:** APPROVE.

  Correct root-cause fix with sound Dioxus wiring. I traced the reactive graph and the terminal-state coverage end to end; verified `cargo clippy --workspace --locked --features
  pentest-platform/desktop-pcap -- -D warnings` is clean (exit 0). Two minor notes, neither blocking.

  ### What I verified (beyond the diff)
  - **No new panic surface.** `use_context::<ConversationRefresh>()` panics if there's no provider ancestor — but `ChatPanel` *already* consumes
  `use_context::<Signal<Option<ChatHeaderCtx>>>()` unconditionally (mod.rs:190), and `AppLayout` is the **sole** provider of `ChatHeaderCtx` (app_layout.rs:51). So every `ChatPanel` render 
  site already requires an `AppLayout` ancestor today; the new context (provided at app_layout.rs:56, consumed at mod.rs:195) rides the exact same invariant. `Sidebar` consumes it too but
  renders only inside `AppLayout` (app_layout.rs:89), so it's trivially safe. If any path rendered these outside `AppLayout`, it would already panic pre-PR.
  - **Bump coverage is complete and compiler-enforced.** In `poll_and_update`, all three terminal exits bump (`saw_error` active, `done && has_agent_msg` active, and the
  switched-away-mid-await `else if`); the two non-terminal returns (switched-away-before-terminal at loop top, and the `Err` fetch path) correctly do *not* bump. `conversations_refresh` is 
  a required param with no default, so the six poll call-sites + three `create_conversation` sites can't be missed silently — the build enforces it.
  - **No self-retrigger loop.** The sidebar effect reads `convo_refresh`/`api_url`/`auth_token` and writes `recent_convos`; it never reads `recent_convos` and never writes the trigger, so a
  fetch can't re-arm itself. Reading the trigger before the credentials early-return keeps the subscription registered even pre-auth — good.
  - **Write-from-spawn works.** The original bug was reads inside `spawn` not subscribing the effect; signal *writes* schedule subscribers regardless of context, so bumping from spawned
  tasks (via the `Copy` newtype) is correct.

  ### Minor (non-blocking)
  1. **Refresh chattiness.** A new conversation bumps on create *and* again at turn completion (≥2 sidebar re-fetches), and multi-agent flows (validator/report) multiply that. The fetch is
  an idempotent GET and the list converges, so it's fine — but a short debounce/coalesce on the trigger would cut redundant fetches if you see churn.
  2. **Doc nuance.** `poll_and_update`'s doc says it bumps "whenever a turn reaches a terminal state." If the user switches away *before* terminal, the loop-top `!is_active` return stops
  this poller and a later completion goes unobserved (no bump until the next create/turn). That's pre-existing polling-stops-on-switch-away behavior, not a regression — consider wording it
  "when this poller observes a terminal state."

  ### Test coverage
  No unit test, and I agree that's correct here: the only pure logic is `notify_conversations_changed`'s wrapping increment (trivial), and the rest is framework wiring with no render-test
  harness in this crate — a component test would be a tautology. The author's manual test plan is the right verification. (I verified compile + lint across the workspace; I did not re-run
  the 95 pentest-ui unit tests — low regression risk for additive wiring — happy to if you'd like.)

cchao-devo (approved):
LGTM

PR #371, opened 2026-08-13
fix(security): close chart post-processor XSS + pin CDN deps with SRI

### LINKED ISSUE #365: security: sanitize markdown before dangerous_inner_html (file viewer + chat panel)

jtomek-strike48:
## Summary
- Closes the remaining XSS sinks in the chat chart post-processor (`chart_processor.js`) — the out-of-diff half of the #365 class flagged in the #367 review (finding #2).
- Mermaid/ECharts **error messages now render via `textContent`, not `innerHTML`**. A malformed ` ```mermaid ` / ` ```echarts ` fence embeds the offending (untrusted) input in the error message, which was reaching `innerHTML`.
- **ECharts options are sanitized before `setOption`**: every `formatter` (string *or* array — ECharts renders these as raw HTML in tooltips) is dropped, and any `renderMode` is forced to the non-HTML `richText` engine.
- Mermaid `securityLevel: 'strict'` is now set explicitly, so a future default change can't silently disable its DOMPurify output sanitization.
- **Supply chain (#367 review, finding #5):** mermaid/echarts were loaded from jsdelivr on floating majors (`@11` / `@5`) with no integrity check. Pinned to `11.16.1` / `5.6.0` with Subresource Integrity (sha384) + `crossorigin`, so a swapped CDN artifact fails closed.

## Verification
- `node --check` clean; a behavioral test of the option sanitizer confirms string + array formatters and event-handler payloads are neutralized and `renderMode` is forced, while safe fields (titles, data) are preserved.
- The `graphic type:'html'` vector raised in security review was **verified to be a non-sink** against the pinned `echarts@5.6.0` bundle: there is no such graphic element type (every `"html"` literal is the tooltip `renderMode`; graphic text renders on canvas, not HTML). Documented inline.
- Rust build is unaffected — the file is embedded as a string via `include_str!`.

## Residual (follow-up, out of scope)
- ECharts rendered from untrusted JSON is a broad surface (`title.link` phishing; interactive `toolbox`/`dataView`). Deeper isolation (dropping interactive components, CSP) can be a follow-up.
- Remote `https` images/links are allowed by design and can act as tracking beacons on view (same residual noted on #367).

## Test plan
- [x] `node --check crates/ui/src/assets/chart_processor.js`
- [x] Option-sanitizer behavioral test (formatter string/array, graphic, renderMode)
- [x] SRI hashes recomputed for the pinned versions
- [ ] CI green

Closes #365.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -79,6 +79,19 @@ jobs:
       # failure still fails the job.
       - run: cargo test --workspace --locked --no-fail-fast --features pentest-platform/desktop-pcap
 
+  chart-xss-guard:
+    name: Chart XSS Guard
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v5
+      - uses: actions/setup-node@v4
+        with:
+          node-version: '20'
+      # Behavioral regression guard for the chart_processor.js ECharts option
+      # sanitizer (#371). No JS toolchain otherwise exists here; this is a
+      # dependency-free `node` self-test, so it needs no install step.
+      - run: node crates/ui/src/assets/chart_processor.test.js
+
   check-macos:
     name: Check (macOS)
     runs-on: macos-latest
--- a/crates/ui/src/assets/chart_processor.js
+++ b/crates/ui/src/assets/chart_processor.js
@@ -2,23 +2,81 @@
     if (window.__chatChartsInit) return;
     window.__chatChartsInit = true;
 
+    // CDN dependencies. Pinned to exact versions with Subresource Integrity so a
+    // compromised or swapped CDN artifact fails closed instead of executing in the
+    // app origin (#367 review, finding #5). Bump the version AND the integrity hash
+    // together — recompute with:
+    //   curl -sS <url> | openssl dgst -sha384 -binary | openssl base64 -A
+    var MERMAID_SRC = 'https://cdn.jsdelivr.net/npm/mermaid@11.16.1/dist/mermaid.min.js';
+    var MERMAID_SRI = 'sha384-aBQXj4hK6Jm05i7aQAsUV3bLdSUrHX1BGYfMB0166TtWt/RRaw+h0Eelme9OCOvy';
+    var ECHARTS_SRC = 'https://cdn.jsdelivr.net/npm/echarts@5.6.0/dist/echarts.min.js';
+    var ECHARTS_SRI = 'sha384-pPi0zxBAoDu6+JXW/C68UZLvBUUtU+7zonhif43rqj7pxsGyqyqzcian2Rj37Rss';
+
+    // Load a CDN script with SRI. crossOrigin is required for the browser to
+    // verify integrity on a cross-origin resource.
+    function loadScript(src, integrity, onload) {
+        var s = document.createElement('script');
+        s.src = src;
+        s.integrity = integrity;
+        s.crossOrigin = 'anonymous';
+        s.onload = onload;
+        s.onerror = function() {
+            console.error('[PentestConnector] failed to load (SRI or network): ' + src);
+        };
+        document.head.appendChild(s);
+    }
+
+    // Render an error message as TEXT, never HTML. mermaid/jison errors and JSON
+    // parse errors embed a snippet of the offending (untrusted) input, so building
+    // this via innerHTML was a stored-XSS sink (#367 review, finding #2).
+    function showVizError(container, message) {
+        container.textContent = '';
+        var err = document.createElement('div');
+        err.style.cssText = 'color:#f38ba8;font-size:0.75rem;';
+        err.textContent = message;
+        container.appendChild(err);
+    }
+
+    // ECharts renders a `formatter` (e.g. tooltip.formatter) as raw HTML, so
+    // untrusted chart JSON like {"tooltip":{"formatter":"<img src=x onerror=...>"}}
+    // is XSS on hover. Drop every `formatter` regardless of type — a string, or an
+    // array of strings (valid multi-series syntax) — since untrusted content has no
+    // business supplying one; charts still render with the default, escaped tooltip
+    // content. Also force any renderMode to the non-HTML 'richText' engine
+    // (#367 review, finding #2). Note: the confirmed HTML sink is the tooltip
+    // formatter; ECharts 5.6.0 has no `type:'html'` graphic element (verified
+    // against the bundle: every "html" literal is the tooltip renderMode).
+    function stripHtmlSinks(node) {
+        if (Array.isArray(node)) {
+            for (var i = 0; i < node.length; i++) stripHtmlSinks(node[i]);
+        } else if (node && typeof node === 'object') {
+            Object.keys(node).forEach(function(key) {
+                if (key === 'formatter') {
+                    delete node[key];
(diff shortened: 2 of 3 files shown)


cchao-devo (approved):
Review of pick PR #371 — fix(security): close chart post-processor XSS + pin CDN deps with SRI

  Verdict: APPROVE — closes the #365/#367 sinks correctly; one test-coverage follow-up recommended.

  Strong, well-scoped security fix that closes the out-of-diff half of the #367 finding. I verified the sinks by reading the whole file and independently recomputed the SRI hashes against the
  live artifacts rather than trusting the diff. One non-blocking gap (regression test) and one minor residual.

  What I verified

  - SRI hashes are correct. Independently recomputed sha384 for both pinned artifacts and they match byte-for-byte:
    - mermaid@11.16.1 → sha384-aBQXj4hK6Jm05i7aQAsUV3bLdSUrHX1BGYfMB0166TtWt/RRaw+h0Eelme9OCOvy ✓
    - echarts@5.6.0 → sha384-pPi0zxBAoDu6+JXW/C68UZLvBUUtU+7zonhif43rqj7pxsGyqyqzcian2Rj37Rss ✓
  With crossOrigin='anonymous', a swapped CDN artifact fails closed. onerror logs the failure.
  - Error sinks closed. All three div.innerHTML = '<div…>error: ' + msg sites now route through showVizError, which clears the container and sets the untrusted message via textContent — no HTML
  parsing. The only remaining innerHTML is result.svg from mermaid, now with securityLevel:'strict' set explicitly, so its DOMPurify sanitization can't be silently disabled by a future default
  change. The reliance is documented inline. Good.
  - ECharts option sanitizer is broad and correctly wired. stripHtmlSinks recurses every key of the parsed JSON, deletes every formatter (string or array), and forces any renderMode to richText.
  The mutation happens synchronously before the setTimeout, and the same option object flows into setOption (no re-parse); the only post-sanitize mutations are app-controlled (backgroundColor,
  textStyle.color).
  - graphic type:'html' claim holds. ECharts' graphic element types are all zrender/canvas (image/text/group/rect/…); there is no HTML graphic element in 5.6.0, so the tooltip
  formatter/renderMode pair is the right sink to target.
  - node --check clean; no eval, new Function, document.write, insertAdjacentHTML, or outerHTML anywhere in the file.

  Recommend (non-blocking) — commit the sanitizer regression test

  The test plan checks "[x] Option-sanitizer behavioral test", but nothing is committed (the diff is chart_processor.js only), and there's no JS test harness in the repo — so stripHtmlSinks has
  zero regression protection and a future edit could silently reopen the XSS. This is exactly the lesson from #336. Two pragmatic options given there's no JS toolchain here:
  1. A standalone node self-test wired as a CI step (mirrors #336's check-pii.test.sh) asserting: a string formatter is dropped, an array formatter is dropped, a nested/deep formatter is
  dropped, renderMode is forced to richText, and safe fields (title/data) survive.
  2. Or, as a minimum, a Rust unit test in render.rs asserting invariants on the embedded CHART_PROCESSOR_JS string (contains securityLevel: 'strict', both integrity hashes present; no innerHTML
  = '<div error pattern) — runs in the existing cargo test CI with no new tooling.

  Minor residual (optional)

  In the same class as the tracking-beacon residual you already noted: ECharts tooltip.extraCssText injects raw CSS into the tooltip DOM (e.g. background-image:url(...) can beacon on hover). Not
  script-XSS, but if you want to close the CSS-injection surface too, dropping extraCssText in stripHtmlSinks alongside formatter is a one-liner. Fine to leave as documented residual.

jtomek-strike48:
Addressed both review recommendations in a451312:

**1. Committed regression test for `stripHtmlSinks` (recommended)** — went with the behavioral node self-test (your option 1), not the string-invariant Rust test, because a source-grep guard stays green if the recursion/`delete` logic breaks. `crates/ui/src/assets/chart_processor.test.js` exercises the *real* exported sanitizer: string/array/deeply-nested `formatter` dropped, `renderMode` forced to `richText`, and safe fields (title/axis/series data) preserved. Wired as a required CI job (`Chart XSS Guard`) — dependency-free `node`, no new toolchain.
  - **Mutation-verified** (Lens 1): neutering `stripHtmlSinks` turns the test red (`exit=1`); restoring it → green. It genuinely guards the fix.
  - The sanitizer is exposed to node via a CommonJS shim that early-returns before the browser wiring; it's a no-op in the browser (no `module` global — Pick injects the file as a raw `<script>` via `include_str!`, no bundler).

**2. `extraCssText` (optional minor residual)** — dropped alongside `formatter` in `stripHtmlSinks`, closing the CSS-injection / `background-image:url(...)` beacon-on-hover surface. Covered by a test case.

Residuals unchanged and still documented as out-of-scope follow-ups (CSP / dropping interactive components; remote https image/link tracking beacons).

PR #376, opened 2026-08-14
fix(auth): close desktop OAuth callback XSS + add CSRF state (#375)

### LINKED ISSUE #375: Desktop OAuth flow hardening: callback matrix_url XSS + missing loopback CSRF state
## Summary

Surfaced during review of #249. Two hardening gaps in the **desktop** browser-OAuth flow (`crates/core/src/matrix/auth.rs`). The native iOS/Android flows are not affected (they use a verified CSPRNG `state`).

## 1. Callback page XSS via `matrix_url` (auth.rs:853)
The loopback callback HTML interpolates `matrix_url` directly into a single-quoted JS string literal:

```
var MATRIX_URL = '{matrix_url}';
```

A `matrix_url` containing `'` breaks out of the literal and injects script into a page that handles the access token, so XSS here equals token theft. Reachable if an operator is induced to configure a malicious server URL.

**Fix:** `serde_json`-encode the value before interpolation (or otherwise JS-escape it).

## 2. Desktop loopback OAuth omits CSRF `state` (auth.rs:~1053)
The desktop loopback authorize URL is built as `{base}/auth/login?redirect=http://localhost:{port}/callback` with **no `state` parameter**, and `/callback` + `/token` do not verify one. Both native flows generate and verify `state` (auth.rs:436, :550), and the comment at auth.rs:469-475 notes the loopback nonce is more load-bearing than the native one, yet the loopback path is the one without it.

A cross-site page or a local process can reach `http://localhost:PORT/token?access_token=...` while the loopback server is up and fixate an attacker-controlled token into the user's session.

**Fix:** generate a CSPRNG `state`, round-trip it via `?state=`, and reject a callback whose `state` does not match on the desktop path too.

Source: #249 review (SHOULD-FIX).

jtomek-strike48:
## Summary
- **Callback-page XSS (token theft):** the loopback browser-OAuth `/callback` page interpolated the operator-configured `matrix_url` into a single-quoted JS string literal on the page that handles the access token. A `matrix_url` containing `'` or `</script>` could inject script. Now `matrix_url` (and the new state) are interpolated through `js_string_literal()` — `serde_json`-encode plus escaping of `<`, `>`, and the U+2028/U+2029 JS line terminators — so quote / `</script>` / line-terminator breakouts are inert data. The callback HTML is extracted into a testable `build_callback_html()`.
- **Missing CSRF state (token fixation):** `/callback` and `/token` accepted an `access_token` with no `state`, so any cross-site page or local process could hit the loopback server during its ~120s window and fixate an attacker-chosen token into the session. Now a single-use CSPRNG `state` (UUIDv4) is generated per login, carried on the loopback redirect and on every client-side `/token` path, and token acceptance on **both** endpoints routes through one choke point `accept_callback_token()` that requires a non-empty token AND a constant-time-matching `state`.

Both defects are pre-existing on `main` (verified: the helpers don't exist there and the callback HTML interpolates `matrix_url` unescaped with no `state`). Native iOS/Android flows are out of scope here (they already use a verified CSPRNG state, and that code is not on `main`).

Closes #375.

## Known caveat (documented, not a regression)
Strategy 3 (the `#194`-documented-fragile fallback) builds `tokenUrl = …/token?state=X` and hands it to the server as the `?redirect=` target. If a server appends `?access_token=` instead of `&access_token=`, that fallback's state won't round-trip and it will be rejected. The **primary desktop path (Strategy 2, fully client-controlled) is unaffected**, and the existing `#194` fast-fail diagnostic already covers a stranded Strategy 3. Called out so the state requirement isn't overstated.

## Test plan
- [x] `cargo test -p pentest-core --features browser-auth --lib matrix::auth` — 16 pass (7 hardening + 2 callback-HTML wiring + 7 existing)
- [x] `cargo test -p pentest-core --lib matrix::auth::oauth_hardening_tests` (no feature) — 7 pure-helper tests run on the macOS package-scoped lane too, not just the Linux workspace lane
- [x] **Guard mutation tests:** neutering the state gate reddens `accept_callback_token_requires_matching_state`; dropping the `js_string_literal` wiring reddens `malicious_matrix_url_cannot_break_out_of_script`; dropping the U+2028/29 escape reddens `js_string_literal_escapes_js_line_terminators`
- [x] `cargo clippy -p pentest-core --features browser-auth --all-targets -- -D warnings` — clean (also clean without the feature)
- [x] `cargo fmt` — clean

### DIFF
--- a/crates/core/src/matrix/auth.rs
+++ b/crates/core/src/matrix/auth.rs
@@ -405,6 +405,13 @@ pub async fn fetch_matrix_token_browser(matrix_url: &str) -> crate::error::Resul
         local_port
     );
 
+    // CSRF `state`: an unguessable, single-use token that binds the browser
+    // callback to the flow this connector initiated. `/callback` and `/token`
+    // only accept an `access_token` accompanied by a matching `state`, so a
+    // cross-site page or a local process cannot fixate an attacker-chosen
+    // token into the session while the loopback server is up (issue #375).
+    let oauth_state = generate_oauth_state();
+
     // -----------------------------------------------------------------------
     // Callback HTML — tries multiple strategies to obtain the access token:
     //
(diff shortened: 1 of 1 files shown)


jtomek-strike48:
Rebased on current `main` (`590dc40` → `1079fc2`, clean — the only changed file is `matrix/auth.rs`, no overlap with the intervening #371/#336/#357) to refresh the stale CI; it now runs against current base and picks up the new `Chart XSS Guard` lane.

**Author pre-review (empirical, offered as a head start — not a substitute for your pass):**
- **CI-scope (the browser-auth trap):** the 7 pure hardening helpers are un-gated and run on the macOS package-scoped lane too — confirmed by running `cargo test -p pentest-core --lib matrix::auth` with **no feature** → 7 pass; `--features browser-auth` → 16 pass. No "green but untested" gap.
- **Guards are genuine (mutation-tested):** neutering the `state_matches` gate reddens `accept_callback_token_requires_matching_state` (`Some("jwt")` vs `None`); disabling the `<`→`<` escape reddens `malicious_matrix_url_cannot_break_out_of_script` (2 script tags vs 1).
- **Choke point:** both `/callback` and `/token` route through `accept_callback_token()`; no bypass path writes the token channel without a matching state.
- **Constant-time compare:** XOR accumulator, no early return. Minor note: it length-checks before the loop, but both sides are fixed 36-byte UUIDs and single-use, so no exploitable leak.
- **Secret sinks:** token logged as length only; exposure is loopback-only, not argv/process-table.

`cargo fmt --all -- --check` clean.

cchao-devo (approved):
Review of pick PR #376 (@ 1079fc2) — fix(auth): close desktop OAuth callback XSS + add CSRF state (#375)

  Verdict: APPROVE — both the XSS sink and the CSRF gap are correctly closed, with strong negative-control tests.

  Reviewed the full matrix/auth.rs diff and verified gates/tests locally. Both halves of #375 are handled correctly.

  XSS — fixed correctly

  The old callback page interpolated the (operator-supplied, attacker-influenceable) Matrix URL into a single-quoted inline-script literal (MATRIX_URL = '{matrix_url}') — a '</script>
  in a crafted Studio URL meant script injection on the loopback page that handles the access token, i.e. token theft. The fix routes matrix_url and oauth_state through
  js_string_literal(), which uses serde_json quoting plus </> → \u003c/\u003e (so an embedded </script> can't terminate the block) plus U+2028/U+2029 escaping (JS line terminators valid
  in JSON). I confirmed the only remaining interpolations into the HTML are local_port (a u16, inherently safe) and those two escaped values — no unescaped sink remains. The
  malicious_matrix_url_cannot_break_out_of_script test pins exactly one <script>/</script> pair and the unicode-escaped payload.

  CSRF / token fixation — fixed correctly

  A CSPRNG state (UUIDv4, 122 bits, single-use) is generated per login and both /callback and /token are funneled through one accept_callback_token choke point that accepts a token only
  when it is non-empty and carries the matching state (constant-time compare, defense-in-depth). The state is carried on the loopback redirect URI and wired into every client path that
  posts to /token (Strategies 1/2 fetches + the Strategy 3 redirect target), so the legitimate flow still completes while a blind cross-site page or local process hitting
  /token?access_token=… without the state is rejected. Single choke point means the check can't be bypassed by hitting one endpoint over the other.

  Verified locally (@ 1079fc2)

  - cargo fmt --all --check clean.
  - Tests pass in both feature configs: 16 auth:: tests with browser-auth (incl. the XSS malicious_matrix_url… and CSRF accept_callback_token_requires_matching_state guards), 7 pure
  oauth_hardening_tests without it (the pure helpers are ungated so they run on every CI lane; callback_html_tests are correctly browser-auth-gated).
  - cargo clippy -p pentest-core -- -D warnings clean with and without browser-auth.

  Minor (non-blocking)

  - /token's reject path returns Html("missing token") with a 200 OK; a 400 would be more accurate, though there's no security impact (fail-closed, no token handed over).
  - The fix assumes the server preserves the state query param when it redirects back with ?access_token=…. If some server build replaces the query string entirely, a legitimate token
  would be rejected — fail-closed (login fails, nothing leaks), but worth being aware of for the #194-style server-compat cases.
  - js_string_literal's .unwrap_or_else(|_| "\"\"") is dead (serialize of a &str is infallible); harmless.

  Nice, tightly-scoped fix with tests that would catch a regression of either vector.

PR #378, opened 2026-08-14
feat(onboarding): pick connect <url> for one-command Studio setup

jtomek-strike48:
## Summary

- Adds a `pentest-core::onboarding` module and a `pentest-agent connect <url>` subcommand: from just a Studio URL, it does browser sign-in → tenant discovery → a pre-approved OTT → self-registration, and persists the config with a **URL-scoped instance id** so a new Studio never replays another tenant's saved credential (the recurring "Registration failed" cause).
- Reuses the proven `{api}/auth/login` browser mediation (`crate::matrix::fetch_matrix_token_browser`), so there's no direct Keycloak realm/redirect handling and no client-redirect-whitelist risk. No OIDC discovery needed — the mediation runs the OIDC/PKCE dance server-side.
- Degrades gracefully to the **post-approval** flow where a Studio has no `/api/connectors/pre-approve` endpoint (register pending → approve once in Studio).
- The browser/network path is gated behind the `browser-auth` feature (which `pentest-headless` now enables); the pure helpers stay ungated so their unit tests run on every CI lane, not only where `browser-auth` unifies in.

## Test plan

- [x] Unit tests (URL validation, userinfo rejection, ws-host/TLS derivation, tenant JSON incl. stringified `details`) — pass **with and without** `browser-auth`
- [x] `cargo clippy -D warnings` clean in **both** feature configs; `cargo fmt` clean
- [x] `cargo check`/`clippy`/`test` on the headless binary
- [x] **Live end-to-end** against a Studio: browser → tenant resolved → pre-approved OTT → "Registered successfully"; credential persisted for reconnect
- [ ] Reviewer: exercise the post-approval fallback on a Studio without the pre-approve endpoint

Part of #377. The desktop first-run wizard (same `onboarding` module, second surface) follows in a separate PR.

Issue #377, opened 2026-08-14
Onboarding: connect Pick to any Prospector from just a URL


jtomek-strike48:
## Summary

Connecting a standalone Pick connector to a new Prospector/Studio instance today means hand-editing ~5 env vars in `.env` and dodging several footguns:
- duplicate `STRIKE48_HOST` (last-wins under `set -a; source .env`)
- wrong TLS scheme for the derived Matrix API URL
- **stale credentials are keyed by instance id, not tenant** → the old client/key is replayed via `private_key_jwt` against the new Studio and rejected with "Registration failed"

StrikeHub has a smooth "just the URL" flow, but it bundles an old Pick and is not the recommended runtime path. We want the same UX inside Pick.

## Phase 1 — DONE

`scripts/connect.sh <studio-url> [tenant-uuid]`: derives `STRIKE48_HOST`/`STRIKE48_API_URL`/`STRIKE48_TLS` from the URL scheme, auto-derives a URL-scoped `STRIKE48_INSTANCE_ID` (permanently prevents the stale-credential class), idempotent `.env` upsert (the duplicate-host bug can't recur), backs up existing `.env`, and warns on missing tenant / stale credential / JSON-or-localhost `--ott`. Flags: `--ott`, `--launch`, `--reset-cred`. shellcheck-clean; verified with 9 hermetic cases.

Limitation: the tenant UUID is still a required input — the SDK refuses to register without one (`connector.rs`: "tenant_id is required"), and there is no way to discover it without an interactive login.

## Phase 2 — planned

In-pick OIDC `pick connect <url>` on **both** surfaces (Dioxus desktop first-run wizard + headless CLI) via a shared `pentest-core` onboarding module. Reuse SDK primitives already present in `strike48-connector`:
- `OAuthManager::login_interactive` (browser OIDC)
- server-pushed `OidcConfig` (no Keycloak URL/realm to type)
- `OttProvider::register_with_ott`

Port StrikeHub's thin `sh-core` glue: `validate_studio_url`, `fetch_tenant_id` (resolves tenant from the authenticated session), `create_pre_approved_token` (mints an OTT embedding the correct studio URL — also closes the localhost-OTT origin footgun). Result: type a URL → browser login → tenant discovered + OTT minted → registered, with no tenant/Keycloak/token typed. Needs a device-code / print-link fallback for browserless hosts.

## Acceptance criteria

- [x] `scripts/connect.sh` writes a correct, idempotent `.env` from a URL; verified.
[...]

PR #379, opened 2026-08-14
feat(connect): scripts/connect.sh for one-command Studio .env setup

jtomek-strike48:
## Summary

- Adds `scripts/connect.sh <studio-url> [tenant-uuid]`: derives `STRIKE48_HOST` (wss/ws), `STRIKE48_API_URL` (https/http), and `STRIKE48_TLS` from the URL scheme, plus a **URL-scoped `STRIKE48_INSTANCE_ID`** so a new Studio never replays another tenant's saved credential (the recurring "Registration failed" cause).
- Idempotent `.env` upsert (exactly one active line per key — the duplicate-`STRIKE48_HOST` bug can't recur), backs up any existing `.env`, and warns on a missing tenant, a stale credential for the instance, and a JSON/`localhost` `--ott` value.
- Flags: `--ott <bare token>`, `--instance`, `--reset-cred`, `--launch`.

This is the zero-code onboarding path (no Rust). The in-app OIDC flow (`pick connect <url>`) is #378; the desktop wizard follows.

## Test plan

- [x] 9 hermetic cases in a throwaway sandbox with a fake `HOME` (prod https, local http, `/path` stripping, no-scheme, idempotency, `--help`, missing-tenant, JSON/localhost `--ott` warning, unknown-flag)
- [x] `shellcheck -S warning` clean; `bash -n` clean
- [ ] Reviewer: run against your own Studio and confirm the written `.env` connects

Part of #377.

### DIFF
--- a/scripts/connect.sh
+++ b/scripts/connect.sh
@@ -47,7 +47,7 @@ log() { printf '\033[0;36m[connect]\033[0m %s\n' "$*"; }
 warn() { printf '\033[0;33m[connect] WARN:\033[0m %s\n' "$*" >&2; }
 die()  { printf '\033[0;31m[connect] ERROR:\033[0m %s\n' "$*" >&2; exit 1; }
 
-usage() { sed -n '2,40p' "${BASH_SOURCE[0]}" | sed 's/^# \{0,1\}//'; exit "${1:-0}"; }
+usage() { sed -n '2,/^$/p' "${BASH_SOURCE[0]}" | sed 's/^# \{0,1\}//'; exit "${1:-0}"; }
 
 # ---- parse args ----
 STUDIO_URL=""
@@ -60,8 +60,8 @@ LAUNCH=false
 while [[ $# -gt 0 ]]; do
     case "$1" in
         -h|--help)     usage 0 ;;
-        --ott)         OTT="${2:-}"; shift 2 ;;
-        --instance)    INSTANCE_OVERRIDE="${2:-}"; shift 2 ;;
+        --ott)         [[ -n "${2:-}" ]] || die "--ott requires a token (see --help)"; OTT="$2"; shift 2 ;;
+        --instance)    [[ -n "${2:-}" ]] || die "--instance requires an ID (see --help)"; INSTANCE_OVERRIDE="$2"; shift 2 ;;
         --reset-cred)  RESET_CRED=true; shift ;;
         --launch)      LAUNCH=true; shift ;;
         --*)           die "unknown option: $1 (see --help)" ;;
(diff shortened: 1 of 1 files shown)


cchao-devo (approved):
Review of pick PR #379 (@ a468b63) — feat(connect): scripts/connect.sh for one-command Studio .env setup

  Verdict: APPROVE with two should-fixes — the script is sound and shellcheck-clean, but please gitignore the secret-bearing backups and reconcile the instance-id scheme with #378
  before merge.

  A well-built convenience wrapper: strict mode, idempotent key upsert, portable hash fallback, careful UX warnings, and it doesn't echo the OTT in its summary. shellcheck -s bash is
  clean. Two issues stand out.

  Should-fix 1 (Medium) — secret backups aren't gitignored, in a public repo

  .env itself is gitignored (good), but the script writes "$ENV_FILE.bak.<timestamp>" (line 137-138) which contains the full .env — including STRIKE48_REGISTRATION_TOKEN (the OTT) when
  --ott was used on a prior run. .env.bak.* is not matched by .gitignore (confirmed with git check-ignore), so a git add . in the repo root would stage a live credential into a public
  repo. Please add .env.bak.* (and consider *.bak.*) to .gitignore, or write the backup outside the repo tree (e.g. under $HOME/.strike48/). This is the same secret-in-git class as
  #336.

  Should-fix 2 (Medium) — instance-id derivation diverges from #378's pick connect

  This script computes pick-<first-host-label>-<sha256(ws://hostport)[:6]>, with no device id. The Rust onboarding in #378 uses ConnectorConfig::env_scoped_instance_id(device_id, host)
  = <device_id>-<host_slug>, with no hash. For the same Studio the two "one-command setup" paths therefore produce different STRIKE48_INSTANCE_IDs → different
  pentest-connector_<id>.json credential files → two separate registrations, and neither reuses the other's saved credential. Since URL-scoped instance ids exist precisely to avoid
  credential-replay confusion, and both PRs are in flight, this is the moment to converge them (share one derivation, or document that the script and pick connect are mutually
  exclusive). Note the script's stale-credential check (line 158) is also scoped to its own id scheme, so it won't detect a credential written by the other path.

  Low / nits

  - No userinfo rejection. https://user:pass@host parses host_only as user (line 97) and writes wss://user:pass@host into .env. #378's validate_studio_url rejects userinfo; consider
  matching that here (reject or strip) both for correctness and to avoid persisting credentials in the host.
  - .env / .env.bak permissions. The upsert path lands .env at 600 (via mktemp+mv), but the initial : > .env / cp .env.example and the cp backup can be 644 — a world-readable file
  holding the OTT on a shared host. A chmod 600 "$ENV_FILE" "$backup" would make the intent explicit.

  Verified

  - shellcheck -s bash scripts/connect.sh clean; .env is gitignored; scheme handling (http/ws→plaintext, https/wss/grpcs→TLS, bare/grpc→secure-default) matches #378's coercion approach;
  --launch refuses without a tenant.

PR #381, opened 2026-08-14
chore(deps): bump strike48-connector and strike48-proto 0.4.4 to 0.6.2

cchao-devo:
Bumps the connector SDK 0.4.4 → 0.6.2, picking up reconnect-path hardening: bounded inline JWT re-mint in the reconnect loops, and clearing of a rejected session token
  so a reconnect re-registers.
  
  ## Why
  
  Empirically, 0.6.2 self-heals where 0.4.4 could stall:
  
  - Clean stream disconnect → re-registers in ~130ms (verified: froze Pick's container past the reap threshold, thawed → auto-recovered).
  - Half-open socket → detects and reconnects after backoff (verified: blackholed the gRPC endpoint → recovered on restore).
  
  This complements the server-side reap fix in Strike48/matrix#3685 (which stops the server reaping a *live* connector). Together they close the "overnight deadlock"
  from both sides. Note: 0.6.2 does not by itself fix the resume-race deadlock — that is the matrix-side guard — but it is a genuine improvement and reduces reap→stuck
  frequency.
  
  ## Testing
  
  - Non-breaking: Pick uses none of the APIs removed between 0.4.4 and 0.6.2. `cargo update` touches only these two crates (291 deps unchanged).
  - `cargo fmt --all --check`, `cargo check --workspace --locked`, `cargo clippy --workspace --locked -- -D warnings` — all clean.
  - `cargo test --workspace --no-fail-fast` — green except `all_tools_registered`, which is a pre-existing macOS-only OS-gating artifact (`autopwn_crack` registers on
  Linux only). It fails identically on clean `main` and passes on the Linux CI `Test` lane (the authoritative one).
  
  Relates to Strike48/sdk-rs#59, Strike48/matrix#3685.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4245,7 +4245,7 @@ dependencies = [
  "libc",
  "percent-encoding",
  "pin-project-lite",
- "socket2 0.5.10",
+ "socket2 0.6.3",
  "system-configuration 0.7.0",
  "tokio",
  "tower-layer",
@@ -4266,7 +4266,7 @@ dependencies = [
  "js-sys",
  "log",
  "wasm-bindgen",
- "windows-core 0.57.0",
+ "windows-core 0.62.2",
 ]
 
 [[package]]
@@ -7277,7 +7277,7 @@ dependencies = [
  "quinn-udp",
  "rustc-hash 2.1.1",
  "rustls",
- "socket2 0.5.10",
+ "socket2 0.6.3",
  "thiserror 2.0.18",
  "tokio",
  "tracing",
@@ -7314,7 +7314,7 @@ dependencies = [
  "cfg_aliases 0.2.1",
  "libc",
  "once_cell",
- "socket2 0.5.10",
+ "socket2 0.6.3",
  "tracing",
  "windows-sys 0.60.2",
 ]
@@ -8236,7 +8236,7 @@ version = "1.4.8"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "c4db69cba1110affc0e9f7bcd48bbf87b3f4fc7c61fc9155afd4c469eb3d6c1b"
 dependencies = [
- "errno 0.2.8",
+ "errno 0.3.14",
  "libc",
 ]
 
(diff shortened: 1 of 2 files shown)


jtomek-strike48 (approved):
Approve. Verified empirically in an isolated worktree at head 5d2fd5a against base main 3384d52f, on Linux with rustc 1.92.0 (the repo MSRV, driven by egui/eframe 0.34 in Cargo.toml:21 and unchanged by this PR).

What I verified sound:

- Non-breaking for Pick (the central claim): `cargo check --workspace --locked --features pentest-platform/desktop-pcap` is clean (exit 0, 0 errors). Every connector-SDK symbol Pick consumes (BaseConnector, ConnectorRunner, ConnectorConfig, ConnectorError, ConnectorBehavior, AppManifest, AppPage{Request,Response}, InvokeCapabilityOptions, parse_url, Result) still resolves under 0.6.2. `cargo clippy --workspace --locked ... -D warnings` and `cargo fmt --all --check` are also clean.

- "Touches only these two crates" is accurate. I diffed the full set of resolved (name, version) pairs between the two lockfiles: the only delta is strike48-connector and strike48-proto 0.4.4 -> 0.6.2. Nothing added, removed, or re-versioned otherwise (1062 -> 1062 pairs). The socket2 0.5.10->0.6.3, windows-core 0.57.0->0.62.2, and errno 0.2.8->0.3.14 lines in the Cargo.lock diff are internal dependency-edge re-points to versions that already existed in the lock (both majors were present on base), not new resolutions. No supply-chain surface change.

- all_tools_registered is pre-existing and macOS-only. On Linux HEAD the test passes (test result: ok, 1 passed). The divergence is a platform gate: pentest_tools registers AutoPwnCrackTool unconditionally (tools/src/lib.rs:114), while the connector capability set is filtered by cfg(target_os) in core/src/tools.rs, which this PR does not touch. So it is inherent to the platform, not introduced by the bump. (macOS side verified by reading the cfg gating; I cannot run macOS locally.)

- init-dev relevance: this is the primary beneficiary, not a no-op. The headless connector (apps/headless/main.rs:189) drives connect_and_run() -> the ConnectorRunner reconnect loop, which is exactly where 0.6.2's bounded inline JWT re-mint and rejected-token clearing live. The overnight-deadlock target path is the one that benefits.

Honest disclosure: I did not independently reproduce the SDK-internal reconnect behavior (clean-disconnect re-register, half-open socket recovery) - that depends on the SDK crate internals plus a live cluster. I verified the Pick-side integration compiles and that the headless connector uses that path; I take the cluster self-heal results as author-verified per the PR body.

Out of scope, non-blocking: check/clippy surface a future-incompat lint from the transitive screenshots v0.8.10 (never-type-fallback). It is not in this PR's lock diff and is pre-existing on main, so it is not a concern here - possibly worth a separate tracking issue.

PR #382, opened 2026-08-17
fix(tools): coerce string and numeric booleans in param_bool (#307)

### LINKED ISSUE #307: param_bool ignores string and numeric booleans that LLM callers emit
## Problem

`param_bool` accepts only a real JSON boolean, while its sibling `param_u64` in the same file coerces integers, floats and strings. An LLM tool caller that emits `"true"` or `1` silently gets the default value instead of the flag it asked for.

`crates/tools/src/util.rs:40-42`:

```rust
pub fn param_bool(params: &Value, key: &str, default: bool) -> bool {
    params.get(key).and_then(|v| v.as_bool()).unwrap_or(default)
}
```

Compare `param_u64` at `:28`, which chains `as_u64()` then `as_f64()` then a string parse. The file even documents this exact hazard for the numeric case:

> LLM tool callers frequently emit numeric args as floats (`22.0`), so a raw `as_u64()` here silently rejected valid ports.

The same reasoning applies to booleans and was not carried over.

## Impact

Verified by executing the function bodies directly:

```
param_bool({"append": "true"},  default=false) -> false   # expected true
param_bool({"recursive": 1},    default=false) -> false   # expected true
param_bool({"passive": true},   default=false) -> true    # only real bools work
param_u64  ({"timeout_ms": "5000"}, default=2000) -> 5000  # sibling coerces
```

Every call site silently runs with the wrong flag. Current callers include `write_file` (`append`), `dirb` (`recursive`), `enum4linux` (`enumerate_all`, `verbose`), `rustscan` (`accessible`), `amass` (`passive`, `brute`), `responder` (`analyze`, `wpad`, `force_wpad_auth`, `fingerprint`), `bettercap` (`silent`), `john` (`show`, `incremental`), and `exiftool` (`recursive`).

The failure is silent in both directions: a caller asking to append overwrites instead, and a caller asking for a passive scan gets an active one. The latter has real consequences on a client engagement.

## Reproduction

```bash
rg -n 'fn param_bool' -A 3 crates/tools/src/util.rs
rg -n 'param_bool' crates/tools/src   # enumerate affected call sites
```

## Suggested fix

Mirror `param_u64`'s coercion chain:

```rust
pub fn param_bool(params: &Value, key: &str, default: bool) -> bool {
    params
        .get(key)
        .and_then(|v| {
            v.as_bool()
                .or_else(|| v.as_u64().map(|n| n != 0))
                .or_else(|| match v.as_str()?.trim().to_ascii_lowercase().as_str() {
                    "true" | "1" | "yes" => Some(true),
                    "false" | "0" | "no" => Some(false),
                    _ => None,
                })
        })
        .unwrap_or(default)
}
```

`param_bool` currently has no test at all, unlike `param_u16_opt`. Add one covering `true`, `"true"`, `1`, `"false"`, `0`, and an unparseable value falling back to the default.

## Notes

Pre-existing on `main` (verified at `crates/tools/src/util.rs:40` on current `main`). Surfaced while reviewing #249, but independent of it.

jtomek-strike48:
## Summary
- `param_bool` accepted only a real JSON boolean, so an LLM tool caller emitting `"true"` or `1` for a flag silently got the default. This ran passive scans as active and turned appends into overwrites across `write_file`, `dirb`, `enum4linux`, `rustscan`, `amass`, `responder`, `bettercap`, `john`, and `exiftool`.
- Mirror the sibling `param_u64` coercion chain: accept bool, integer/float (`0` false, non-zero true), and the common string shapes (`"true"`/`"1"`/`"yes"`, `"false"`/`"0"`/`"no"`, case- and whitespace-insensitive). Unrecognized values fall back to the caller's `default` rather than a silent `false`.
- Adds the previously-absent `param_bool` unit-test coverage.

Closes #307.

## Test plan
- [x] `cargo test -p pentest-tools --lib util::` — 4 new tests pass
- [x] Mutation-verified: reverting to the old `as_bool()`-only body turns the numeric/string coercion tests RED
- [x] `cargo clippy -p pentest-tools --all-targets -- -D warnings` clean
- [x] `cargo fmt --all` applied
- [x] Blast radius: change is strictly additive (accepts more input forms, never fewer) — no regression for callers relying on the default

### DIFF
--- a/crates/tools/src/util.rs
+++ b/crates/tools/src/util.rs
@@ -37,8 +37,26 @@ pub fn param_u64(params: &Value, key: &str, default: u64) -> u64 {
 }
 
 /// Extract a `bool` parameter with a default value.
+/// Accepts a real JSON boolean, an integer/float (`0` is false, non-zero true),
+/// or a string (`"true"`/`"1"`/`"yes"` and `"false"`/`"0"`/`"no"`, case- and
+/// whitespace-insensitive). LLM tool callers frequently emit `"true"` or `1`
+/// for a flag; a raw `as_bool()` here silently dropped those to the default,
+/// so a passive scan could run active or an append could overwrite. An
+/// unrecognized value falls back to `default`.
 pub fn param_bool(params: &Value, key: &str, default: bool) -> bool {
-    params.get(key).and_then(|v| v.as_bool()).unwrap_or(default)
+    params
+        .get(key)
+        .and_then(|v| {
+            v.as_bool()
+                .or_else(|| v.as_u64().map(|n| n != 0))
+                .or_else(|| v.as_f64().map(|f| f != 0.0))
+                .or_else(|| match v.as_str()?.trim().to_ascii_lowercase().as_str() {
+                    "true" | "1" | "yes" => Some(true),
+                    "false" | "0" | "no" => Some(false),
+                    _ => None,
+                })
+        })
+        .unwrap_or(default)
 }
 
 /// Convert dBm signal strength to signal quality percentage (0-100)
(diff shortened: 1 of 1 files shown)


cchao-devo (approved):
Review of pick PR #382 — fix(tools): coerce string and numeric booleans in param_bool (#307)

  Verdict: APPROVE the change — needs a rebase before it can merge (currently conflicting/dirty).

  Correct, well-scoped fix for a real correctness/safety bug: an LLM emitting "true"/1 for a flag silently got the default, so a passive scan could run active or an append become
  an overwrite.

  Verified
  
  - Coercion is correct and safe. as_bool → as_u64(≠0) → as_f64(≠0.0) → string-match ("true"/"1"/"yes" vs "false"/"0"/"no", trimmed + lowercased), and — the key property — an
  unrecognized value falls back to default, not a silent false. Mirrors the existing param_u64 coercion philosophy, so the two stay consistent.
  - Edges hold (traced): null/array/""/"maybe"/"2" → default; negative int -1 → true (via as_f64); -0.0 → false. No panic path.
  - Tests + gates green (ran locally): the 4 new util::tests pass (real bool, numeric 1/0/1.0/0.0, strings incl. " TRUE ", and fallback for missing/unrecognized/wrong-type), and
  cargo clippy -p pentest-tools -- -D warnings is clean. Good that the fallback cases assert default, not false, so a regression to silent-false would go red.
  - Blast radius: additive in the safe sense — a real JSON bool behaves identically; only the previously-dropped numeric/string forms now coerce (which is the fix). No caller that
  legitimately passed a bool changes behavior.

  Blocker (state, not code)
  
  mergeable: false / dirty — please rebase onto main (should be a trivial util.rs conflict). No code changes requested.

  Minor (optional)

  Non-"1" digit strings ("2", "-1") fall to default while their integer forms coerce to true — intended and documented (only 1/yes/true are recognized as strings), just noting the
  string/number asymmetry. A couple of the traced edges (negative int, empty string, null) aren't in the test set; cheap to add but not necessary.

jtomek-strike48:
Closing as superseded. The core bug this PR fixes (#307 — `param_bool` dropping `"true"`/`1` from LLM callers) was already fixed independently on `main` in commit 1facaea ("fix(tools): param_bool accepts string and int forms", 2026-08-01), which is why this branch shows as conflicting — it is 474 commits behind and predates that fix. A rebase would be reconciling two implementations of the same fix rather than a trivial conflict.

I compared both implementations empirically. The only behavior `main`'s version lacked is float-valued flag coercion (`{"append": 1.0}`): `main` matches the number arm with `as_i64`, which returns `None` for a JSON float and silently falls back to the default. That single gap — the one thing this PR handled that `main` did not — is carried forward in #390, which switches the number arm to `as_f64` (consistent with `param_u64`/`param_u16_opt`) and closes #307.

Thanks for the fix and the thorough test coverage here; the approach was right, it just landed on `main` first via another PR.

PR #384, opened 2026-08-17
Validate IPv4 dash ranges as ranges, not pseudo-hostnames (#383)

### LINKED ISSUE #383: validate_target does not validate IPv4 dash ranges (accidental hostname pass-through)
## Summary

`crates/core/src/validation.rs` `validate_target` tries CIDR → IP → hostname.
An IPv4 dash-range like `10.0.0.1-50` or `10.0.0.1-10.0.5.20` fails CIDR and IP,
then **passes `validate_hostname`** because each dot-label is hostname-legal
(`10`, `0`, `0`, `1-50`). So ranges are accepted **as pseudo-hostnames**, with no
bounds check and no deliberate range logic. The nmap/masscan `target` param
descriptions promise structured range support the validator does not implement
(description-vs-validation mismatch).

This is the Pick (Rust) half of the cross-repo issue tracked in
Strike48/matrix#3019 (Finding 1 there). The Matrix (Elixir) scope-enforcement
half — Finding 2 in that issue — is already handled on Matrix `develop` by
`Matrix.Net.ScopeGate`, which scope-checks CIDR and IPv4 dash-ranges via
interval math; the `scope_validator.ex` the issue cited is not on the live
enforcement path. So this Pick change is the remaining gap.

## Expected

`validate_target` should validate IPv4 dash-ranges as ranges, with a grammar
that matches Matrix's `ScopeGate` so anything Pick accepts, Matrix can
scope-check:

- Accept **last-octet** ranges (`10.0.0.1-50`) and **full-IP** ranges
  (`10.0.0.1-10.0.5.20`), requiring `start <= end`.
- Reject **multi-octet** cross-products (`10.0.0-1.1-254`), reversed bounds, and
  out-of-range octets (fail closed).
- Leave hostnames-with-dashes (`my-host.example.com`) classified as hostnames.
- Align the nmap/masscan `target` param descriptions with the validated grammar.

## Acceptance

- [ ] `validate_target` routes dash-range literals to a real range validator, not
      the hostname branch.
- [ ] Last-octet and full-IP ranges accepted; multi-octet, reversed, and
      out-of-range rejected.
- [ ] nmap/masscan param descriptions match the validated grammar.
- [ ] Unit, doctest, and integration tests cover accepted forms and fail-closed
      rejections.

## References

- Parent (cross-repo): Strike48/matrix#3019
- Files: `crates/core/src/validation.rs`, `crates/tools/src/external/{nmap,masscan}.rs`

ngdevo:
## Summary

`validate_target` previously let IPv4 dash-ranges (`10.0.0.1-50`,
`10.0.0.1-10.0.5.20`) fall through to the hostname branch, so they passed as
pseudo-hostnames with no bounds check and no agreement with Matrix scope
enforcement. This adds a real range validator and routes range literals to it.

The grammar matches Matrix's `Matrix.Net.ScopeGate`, so every range Pick accepts
is one Matrix can scope-check against out-of-scope. Closes #383 (the Pick half of
Strike48/matrix#3019, Finding 1). The Matrix scope-enforcement half (Finding 2)
is already handled on Matrix `develop` by `ScopeGate`.

## What changed

- Add `validate_ipv4_range`: accepts last-octet (`10.0.0.1-50`) and full-IP
  (`10.0.0.1-10.0.5.20`) forms requiring `start <= end`; rejects multi-octet
  cross-products (`10.0.0-1.1-254`), reversed bounds, and out-of-range octets
  (fail closed).
- Route dash-range literals through it in `validate_target`, before the hostname
  branch. Hostnames-with-dashes (`my-host.example.com`) still classify as
  hostnames.
- Tighten the nmap/masscan `target` param descriptions to state multi-octet
  ranges are rejected (previously "may be rejected").
- Add unit, doctest, and integration tests covering accepted forms and the
  fail-closed rejections.

## Test plan

- [x] `cargo test -p pentest-core validation` (unit) — 62 passed
- [x] `cargo test -p pentest-core --doc validation` (doctests) — 10 passed
- [x] `cargo test -p pentest-tools --test security_tests target_validation` — 5 passed
- [x] `cargo fmt --all -- --check` clean, `cargo clippy -p pentest-core -p pentest-tools --all-targets` clean

## References

- Pick issue: #383
- Cross-repo parent: Strike48/matrix#3019

### DIFF
--- a/crates/core/src/validation.rs
+++ b/crates/core/src/validation.rs
@@ -287,13 +287,88 @@ pub fn validate_cidr(cidr: &str) -> Result<String> {
     Ok(cidr.to_string())
 }
 
-/// Validates target specification (IP, hostname, or CIDR)
+/// Validates an IPv4 dash range.
+///
+/// Accepts exactly the two forms the scan tools advertise and that the Matrix
+/// scope gate (`Matrix.Net.ScopeGate`) can scope-check, so a range Pick accepts
+/// is always one Matrix can enforce against out-of-scope (issue #3019):
+///
+/// - Last-octet: `10.0.0.1-50` — same first three octets, last octet is the
+///   upper bound (`0-255`), and the upper octet must be `>=` the base's.
+/// - Full-IP:    `10.0.0.1-10.0.5.20` — both sides valid IPv4, upper `>=` lower
+///   as a 32-bit integer (no same-/24 restriction).
+///
+/// Rejected (fails closed, matching the scope gate):
+/// - Multi-octet cross-product (`10.0.0-1.1-254`) — the left side is not a valid
+///   IPv4, so nmap's expansion is not scope-checkable and is not accepted here.
+/// - Reversed bounds (`10.0.0.50-1`, `10.0.0.5-10.0.0.1`).
+/// - Out-of-range octet, or any malformed token.
+///
+/// # Examples
+///
+/// ```
+/// use pentest_core::validation::validate_ipv4_range;
+///
+/// assert!(validate_ipv4_range("10.0.0.1-50").is_ok());
+/// assert!(validate_ipv4_range("10.0.0.1-10.0.5.20").is_ok());
+/// assert!(validate_ipv4_range("10.0.0.50-1").is_err());
+/// assert!(validate_ipv4_range("10.0.0-1.1-254").is_err());
+/// ```
+pub fn validate_ipv4_range(range: &str) -> Result<String> {
+    let range = range.trim();
+
+    let err = || Error::InvalidParams(format!("Invalid IPv4 range: {}", range));
+
+    // Only the first dash separates the bounds. A left side like "10.0.0" (from
+    // a multi-octet form) is not a valid IPv4 and is rejected below.
+    let (left, right) = range.split_once('-').ok_or_else(err)?;
+
+    let lower = validate_ipv4(left.trim()).map_err(|_| err())?;
+    let lower_int = u32::from(lower);
+
+    let right = right.trim();
+
+    // Bare last-octet upper bound (e.g. "50" in "10.0.0.1-50"): reuse the base's
+    // first three octets. A multi-octet token like "1.1-254" is not `\d{1,3}`,
+    // so it falls to the full-IP branch and is rejected there.
+    let upper_int = if right.chars().all(|c| c.is_ascii_digit()) && !right.is_empty() {
+        let last = right.parse::<u16>().map_err(|_| err())?;
+        if last > 255 {
+            return Err(err());
+        }
+
+        let [a, b, c, _] = lower.octets();
+        u32::from(Ipv4Addr::new(a, b, c, last as u8))
+    } else {
+        u32::from(validate_ipv4(right).map_err(|_| err())?)
+    };
+
+    if lower_int > upper_int {
+        return Err(err());
+    }
+
+    Ok(range.to_string())
+}
+
+/// True when a target is an IPv4 dash-range candidate: it contains a dash and is
+/// made up only of digits, dots, and dashes. This excludes hostnames that
+/// contain dashes (they carry letters) and IPv6 (colons), mirroring the Matrix
+/// scope gate's `range_literal?/1` so both sides classify targets identically.
+fn is_ipv4_range_literal(target: &str) -> bool {
+    target.contains('-')
+        && target
+            .chars()
+            .all(|c| c.is_ascii_digit() || c == '.' || c == '-')
+}
+
+/// Validates target specification (IP, hostname, CIDR, or IPv4 dash range)
 ///
 /// Accepts:
 /// - IPv4: "192.168.1.1"
 /// - IPv6: "2001:db8::1"
 /// - Hostname: "example.com"
 /// - CIDR: "192.168.1.0/24"
+/// - IPv4 dash range: "10.0.0.1-50" or "10.0.0.1-10.0.5.20"
 ///
 /// # Examples
 ///
@@ -303,6 +378,7 @@ pub fn validate_cidr(cidr: &str) -> Result<String> {
 /// assert!(validate_target("192.168.1.1").is_ok());
 /// assert!(validate_target("example.com").is_ok());
 /// assert!(validate_target("192.168.1.0/24").is_ok());
+/// assert!(validate_target("10.0.0.1-50").is_ok());
 /// assert!(validate_target("; rm -rf /").is_err());
 /// ```
 pub fn validate_target(target: &str) -> Result<String> {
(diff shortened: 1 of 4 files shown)


cchao-devo (approved):
Review of pick PR #384 — Validate IPv4 dash ranges as ranges, not pseudo-hostnames (#383)

  Verdict: APPROVE — correct, fail-closed range validator whose grammar I verified matches Matrix's ScopeGate; ran it locally.

  Clean close of the Pick half of #3019 Finding 1. The bug was real: dash-ranges previously fell through to the hostname branch and passed unbounded.

  Verified — grammar agrees with Matrix ScopeGate (the security invariant)

  The point of #3019 is that Pick must accept only ranges Matrix can scope-check. I compared against Matrix.Net.ScopeGate on develop:
  - Classification is equivalent. ScopeGate's range_literal? = contains("-") and ~r/^[0-9.\-]+$/ matches Pick's is_ipv4_range_literal (contains('-') and all chars in [0-9.-])
  exactly — and both correctly leave my-host.example.com as a hostname (it carries letters).
  - Parse matches. Both split on the first dash, require the left to be a valid IPv4, reuse the base's first three octets for a bare last-octet, require lo <= hi, and reject
  out-of-range octets / reversed bounds / multi-octet cross-products.
  - The one internal difference is benign. ScopeGate detects the last-octet form via \d{1,3}; Pick uses "all-digits, then ≤255 via a u16 parse." I traced the boundary cases (-256,
  -1234, -65536, leading-zero -050, dotted right like -1.2) and both sides reach the same accept/reject on every one — so Pick never accepts a range ScopeGate can't parse. Where
  they differ at all (leading zeros on the left: Pick's Ipv4Addr rejects 010.0.0.1), Pick is the stricter side, which is the safe direction. 

  Ran locally

  - cargo test -p pentest-core validate_ipv4_range (unit) and --doc — green.
  - cargo test -p pentest-tools --test security_tests target_validation — 5/5 green (accept valid, reject invalid ranges, reject command injection, reject empty, integration).
  - cargo clippy -p pentest-core -p pentest-tools --all-targets -- -D warnings — clean.

  Coverage
  
  The tests exercise both forms (last-octet + full-IP, incl. cross-/24), single-host lo == hi, and the fail-closed rejections (reversed, equal-as-reversed, multi-octet
  cross-product, out-of-range last octet, u16 overflow, empty sides, 999.x). The routing test confirms ranges hit the validator before the hostname branch and dash-hostnames still
  classify as hostnames. The nmap/masscan param-description tightening ("may be rejected" → "are rejected") matches the new behavior.

  Nicely scoped — the Matrix enforcement half (Finding 2) is correctly noted as already handled by ScopeGate on develop.

jtomek-strike48 (commented):
Reviewed empirically at head d441b16 in an isolated worktree (base origin/main 2cef58d; head is exactly 1 commit ahead, not stale). Tests run on Rust 1.92 (repo MSRV; mise-pinned 1.91.1 is rejected by the workspace).

## Verified sound

- **The fix is real and its tests genuinely guard it.** I neutered the new routing branch (`if false && is_ipv4_range_literal(...)`) so ranges fall back to the old hostname path. Both reject tests went RED at exactly `assert!(validate_target("10.0.0.50-1").is_err())`: `test_validate_target` (validation.rs:517) and integration `test_reject_invalid_ranges` (security_tests.rs:417). Restored -> green. Pre-PR, `validate_hostname` accepted `10.0.0.50-1`, `10.0.0-1.1-254`, `10.0.0.1-300` as pseudo-hostnames (their dot-split labels are all valid), so this closes a genuine fail-open.
- **Tests run in the required CI job.** The Linux `Test` job is `cargo test --workspace` (ci.yml:80), which runs the new unit, doctest, and integration tests. The macOS lane (`--package pentest-platform --package pentest-core --lib`, ci.yml:109) is a narrower redundant check. No feature-gating hides the code.
- **Classification matches ScopeGate.** Pick `is_ipv4_range_literal` (`contains('-') && all [0-9.\-]`) is intent-identical to `Matrix.Net.ScopeGate.range_literal?/1` (`~r/^[0-9.\-]+$/`). No misclassification fail-open.
- **Last-octet form is genuinely aligned.** `nmap -sL -n 10.0.0.1-50` expands to 50 hosts in the same /24; Pick, ScopeGate, and nmap agree. The `"is rejected"` doc tightening for multi-octet is true.

## SHOULD-FIX: the full-IP form is a false, safety-relevant claim for nmap

nmap.rs:52-54 and validation.rs:298 accept and now document the full-IP form `10.0.0.1-10.0.5.20` as a scope-aligned range for nmap, but nmap does not range-expand it:

```
$ nmap -sL -n 10.0.0.1-10.0.5.20
Nmap scan report for 10.0.0.1-10.0.5.20 (100.100.172.23)
Nmap done: 1 IP address (0 hosts up) scanned in 0.08s
```

nmap treats the whole token as a single hostname and forward-resolves it (Pick's nmap adds only `-Pn`, never `-n`, so DNS is live at dispatch: nmap.rs:225). It scanned 1 arbitrary resolved host, not the interval. By contrast, the multi-octet form the PR rejects (`10.0.0-1.1-254`) is the one nmap natively expands (508 hosts, all in 10.0.x).

Consequence: the PR/docstring claim "every range Pick accepts is one Matrix can scope-check, so nmap's expansion matches what the scope gate checks" is false for nmap + full-IP. `ScopeGate` reasons about `10.0.0.1-10.0.5.20` as the integer interval and may allow or inject `--exclude` for that interval, but nmap ignores the interval and scans whatever the token resolves to, a host ScopeGate never checked. That is the scope-divergence class #3019 exists to close, left open for nmap + full-IP.

Scope-honest note: this is not a new regression. Pre-PR the same token passed `validate_hostname` and reached nmap identically. But this PR is a scope-alignment change that newly blesses and documents the nmap + full-IP path as aligned when it isn't. masscan is unaffected: it supports full-IP ranges natively, so for masscan the form and claim are correct.

Fix options: (a) cheapest and honest: drop the full-IP "supported" claim from the nmap `target` description and the "matches nmap's expansion" framing; (b) for the nmap tool, reject the full-IP form or translate it to nmap octet-range/CIDR before dispatch. Adding `-n` is not a fix (it would still scan the wrong single host).

## LOW / note: minor Pick vs ScopeGate grammar divergences (both fail-closed)

- Zero-padded octet of 4+ digits: Pick accepts `10.0.0.1-0050` (all-digit -> 50); ScopeGate `^\d{1,3}$` rejects -> deny. Matrix stricter.
- Leading-zero left IP `010.0.0.1-50`: Rust `Ipv4Addr::from_str` rejects; Erlang `:inet.parse_address` may accept. Pick stricter.

Both directions have one side stricter (fail-closed), so neither is a bypass. Worth a one-line note, not blocking.

## Verdict

Solid, well-tested hardening that closes a real fail-open. Not blocking. One SHOULD-FIX: the nmap full-IP claim is false and leaves a scope-divergence for nmap; correcting the nmap description (or rejecting/translating the form for nmap) closes it.

PR #388, opened 2026-08-18
ci(release): default desktop + headless release builds to easy mode

josh-devo:
## What

Bakes `PICK_EASY_MODE=true` into the **desktop** (×3) and **headless agent** (×3) release jobs. Desktop additionally bakes `STRIKE48_HOST=https://studio.strike48.com`.

## Why

Only the **Android** artifacts previously shipped easy-mode-first. Two gaps:

- **Desktop**: shipped neutral. A distributed desktop binary (no runtime `.env`) had no host to auto-connect to, so easy-mode fell through to the connect form. Now it boots into easy mode and auto-connects to studio, matching Android.
- **Headless agent**: this is the artifact **StrikeHub embeds** (registry asset `pentest-agent-{os}-{arch}`, `binary_hint: pentest-agent`). It serves the LiveView workspace UI, whose easy-mode resolves as *persisted setting > build-time `PICK_EASY_MODE` > app default (`false`)*. So a **local** easy-mode build of `pentest-agent` showed easy mode, but the **released** one booted non-easy — StrikeHub inherited that. Baking `PICK_EASY_MODE=true` makes the released agent match.

## Scope / deliberately unchanged

- **Web** stays neutral (compiled `easy_mode=false`, host from runtime env).
- **Headless bakes the flag only, not `STRIKE48_HOST`**: under StrikeHub the host arrives over the IPC socket (`ipc_runner.rs`), and standalone `pentest-agent connect <url>` supplies it via the onboarding flow.
- **iOS** unchanged (disabled).
- Env is set at each **Build step**, so it doesn't leak into signing/packaging.

## Bake matrix after this PR

| Job | `PICK_EASY_MODE` | `STRIKE48_HOST` |
|---|---|---|
| desktop ×3 | ✅ | ✅ studio |
| headless ×3 | ✅ | ❌ (IPC / connect supplies) |
| android | ✅ (justfile default) | ✅ studio (pre-existing) |
| web | ❌ | ❌ |
| iOS | disabled | — |

## Verification

- `yq` parses the workflow; per-job `awk` map confirms the flags land only on the six intended jobs, web/release untouched.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -382,7 +382,15 @@ jobs:
           sudo apt-get update
           sudo apt-get install -y protobuf-compiler libpcap-dev
 
+      # Easy-mode-first: bake PICK_EASY_MODE so the served LiveView UI defaults to
+      # easy mode (resolve_easy_mode: persisted setting > build-time PICK_EASY_MODE
+      # > app default). This is the agent StrikeHub embeds (registry asset
+      # pentest-agent-{os}-{arch}), so the released StrikeHub connector matches a
+      # local easy-mode build. No STRIKE48_HOST is baked: under StrikeHub the host
+      # arrives over IPC, and standalone `pentest-agent connect <url>` supplies it.
       - name: Build
+        env:
+          PICK_EASY_MODE: "true"
         run: cargo build --locked --package pentest-headless --release --features pentest-platform/desktop-pcap
 
       - name: Package
@@ -410,7 +418,10 @@ jobs:
           version: "27.x"
           repo-token: ${{ secrets.GITHUB_TOKEN }}
 
+      # Easy-mode-first headless build (see the Linux headless job for rationale).
       - name: Build
+        env:
+          PICK_EASY_MODE: "true"
         run: cargo build --locked --package pentest-headless --release
 
       - name: Sign Windows binary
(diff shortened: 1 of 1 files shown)


jtomek-strike48 (approved):
Reviewed empirically against the Pick source (not just the diff). Approving.

## Verified sound
- **Both env var names resolve at BUILD time, so baking them is effective (not a no-op).** `PICK_EASY_MODE` is read via `option_env!` in `crates/core/src/config.rs:820` (inside `resolve_easy_mode/2`), and `STRIKE48_HOST` via `option_env!("STRIKE48_HOST").or(STRIKE48_URL).or(STRIKE48_API_URL)` at `config.rs:450`. Confirmed these are the macros the running code actually consumes.
- **Precedence claim is accurate.** `resolve_easy_mode` is `persisted Settings > build-time PICK_EASY_MODE > app_default`, so the headless jobs baking only the flag (app_default stays `false` at `workspace_app.rs:322`) still flip to easy mode, and the in-app toggle still wins. Desktop consumes the baked host via `ConnectorConfig::from_baked_or_env()` (`connector_app.rs`), which is why desktop also needs `STRIKE48_HOST`.
- **Scope is exactly as claimed.** The six `env:` blocks land on the desktop (`pentest-desktop` x3: release.yml ~55/83/133) and headless (`pentest-headless` x3: ~369/397/447) Build steps. The web Build step (`pentest-web`, ~217) is untouched; Android (`Build APK`/`Build AAB`) and iOS (`Build (Simulator)`) are untouched. Web/iOS stay neutral as documented.
- **Host+scheme matches the already-shipping Android bake.** Android release jobs already bake the identical `STRIKE48_HOST: https://studio.strike48.com` (release.yml:277/287), so "desktop matches Android" is literally true and the scheme is the one already in production.
- Env is scoped to each Build step, so it does not leak into signing/packaging. No secrets (public host, already present in the justfile defaults).

Title honestly covers the whole single-file diff; no scope creep. LGTM.

Issue #251, opened 2026-07-17
Raw sockets are faked under userspace sandbox (nmap -sS etc. silently degrade)


jtomek-strike48:
## Summary

Tools that need `CAP_NET_RAW` (nmap SYN scan `-sS`, ARP scan, masscan, responder) do not get real raw-socket privileges when Pick runs unprivileged with a userspace sandbox. proot reports uid 0 inside the sandbox (so pacman works), but the kernel credentials remain the unprivileged host user and proot cannot honor file capabilities. Net effect: `nmap -sS` likely silently falls back to a TCP connect scan — a **quiet capability degradation**, not an error, which is dangerous for a pentest tool (results look fine but are weaker/different than requested).

## Why it matters

Raw-socket reach is the Tier-1 capability that distinguishes a real pentest connector from a generic tool runner. Silent degradation means the operator (and the agent) believe a SYN scan ran when it didn't. Related: #183 already notes nmap raw-socket privilege is hardcoded `false` on Windows.

## Confidence

Mechanism is well-established (proot fake-root, no CAP_NET_RAW, can't honor file caps). The specific claim "nmap -sS currently degrades to connect scan on this setup" is **not yet empirically confirmed** — needs a live capability test. Connect scan (`-sT`) is verified working.

## Real fixes (design)

- Run the connector **as root on the host** -> bwrap path can `--cap-add ALL` for genuine raw sockets (bwrap only, not proot).
- Or grant the connector process `CAP_NET_RAW` (setcap / systemd `AmbientCapabilities`).
- Or Docker-on-Linux with `--network=host --cap-add=NET_RAW`.
- On Mac/Windows this is additionally constrained: Docker/WSL2 network is a VM, so `--network=host` binds the VM's network, not the target LAN. Raw scans there hit the VM, not the LAN — a hard platform limit, not a bug we can code away.

## Acceptance criteria (draft)

- [ ] Empirically determine current `nmap -sS` behavior under each backend (does it degrade? silently?).
- [ ] When raw sockets are unavailable, the tool/agent is told **explicitly** (typed capability signal) rather than silently degrading.
- [ ] Document per-backend raw-socket support (proot: no; bwrap-as-root: yes; docker `--cap-add`: yes; Mac/Win VM: LAN-limited).

## Context

Surfaced during #248 proot investigation. See also the cross-platform capability tiering (Tier 1 = raw/LAN, Linux-host-first; Tier 2 = portable tool exec).

PR #389, opened 2026-08-19
feat(agent): fact-first network discovery (inject subnets, scope hints)

### LINKED ISSUE #347: Agent-facing: inject live subnets + fact-first prompt rule + scope-aware errors

jtomek-strike48:
## Summary

- Inject the connector's live subnets into the Red Team agent's system message at registration, and add a mandatory "Fact-First Targeting" rule to the persona, so the agent scans a known network (or `target="auto"`/`"current"`) instead of guessing a range like `192.168.1.0/24` - the root cause of the scan-nothing-on-the-wrong-subnet failure.
- On a zero-host nmap sweep, attach a `scope_hint` that names the host's actual subnets and, when raw-socket capability was unavailable in-sandbox, states the discovery was degraded (#251/#309) rather than masking it as an empty network.
- Closes #347 (child of #344). Subnets come from `network_context()` (the #348/#349 foundation); `pentest-core` stays dependency-clean because `core -> tools` is a dev-only dep, so the fetch lives in the `crates/ui` wrapper.

## Changes

- `crates/core/src/matrix/agent_defaults.rs`: `default_pentest_agent_input` takes `active_subnets`; new `host_network_facts_section`; extracted `available_tools_section`; "Fact-First Targeting (mandatory)" rule added to `RED_TEAM_SYSTEM_PROMPT`.
- `crates/ui/src/components/chat_panel/{constants.rs,mod.rs}`: wrapper is now async and sources subnets from `network_context()` (empty on error); two call sites await it.
- `crates/tools/src/scan_scope_hint.rs` (new, pure + unit-tested) wired into `crates/tools/src/external/nmap.rs` on the zero-host path.

## Test plan

- [x] `cargo test -p pentest-core --lib matrix::agent_defaults` - 7/7
- [x] `cargo test -p pentest-tools --lib scan_scope_hint` - 3/3
- [x] Guard mutations: neutering `host_network_facts_section`, the fact-first prompt text, and the degradation clause each reddens its test; all reverted
- [x] `cargo clippy --workspace --all-targets --features pentest-platform/desktop-pcap -- -D warnings` - clean
- [x] `cargo fmt --all` - clean
- [ ] Behavioral (AC #4, not unit-testable per the issue): run `./run-pentest.sh headless dev` against a live workspace and confirm the agent reads the injected subnets / uses `target="auto"` instead of guessing a range. Deferred as a documented manual step.

CI-scope note: `agent_defaults` tests run in both CI lanes; `scan_scope_hint` (pure) runs in the Linux workspace lane.

### DIFF
--- a/crates/core/src/matrix/agent_defaults.rs
+++ b/crates/core/src/matrix/agent_defaults.rs
@@ -21,31 +21,14 @@ fn build_tool_configs(names: &[String]) -> serde_json::Value {
     serde_json::Value::Object(map)
 }
 
-/// Build the default CreateAgentInput for auto-creating a pentest-connector persona.
+/// Build the "tools available on THIS connector" preamble.
 ///
-/// `tenant_id` is the tenant/realm name (e.g. "non-prod") used to build the
-/// connector address pattern `{tenant}.{connector_name}.*` so the Matrix
-/// backend can match registered connector tools to this agent.
-///
-/// `connector_name` controls the gateway identity. Instances sharing the same
-/// name are round-robin'd; use a unique name (e.g. `pentest-connector-<hostname>`)
-/// to get a dedicated agent view.
-///
-/// `tool_names` are the registered tool names to auto-approve. Callers pass
-/// this explicitly (the Dioxus app sources it from its global session; the
-/// crux shell passes the tools it knows about) so this builder stays pure.
-/// Build the agent system message: the static red-team persona plus a dynamic
-/// preamble listing the tools THIS connector actually advertises. The persona
-/// documents tools like `nmap` / `execute_command` for the full desktop
-/// toolset, but a given connector (notably iOS — no fork/exec, no sandbox) only
-/// exposes a subset. Without being told, the agent reaches for tools it "knows"
-/// but that aren't advertised here, and those calls just fail (the observed
-/// "tried nmap / execute_command on iOS" behavior). Prepend the real list and
-/// tell it not to reach outside that set.
-fn system_message_with_available_tools(tool_names: &[String]) -> String {
-    if tool_names.is_empty() {
-        return RED_TEAM_SYSTEM_PROMPT.to_string();
-    }
+/// The persona documents the full desktop toolset (`nmap`, `execute_command`,
+/// …), but a given connector (notably iOS — no fork/exec, no sandbox) exposes
+/// only a subset. Listing the real set and telling the agent not to reach
+/// outside it prevents calls to tools this platform can't run (the observed
+/// "tried nmap / execute_command on iOS" behavior).
+fn available_tools_section(tool_names: &[String]) -> String {
     let mut names: Vec<&str> = tool_names.iter().map(String::as_str).collect();
     names.sort_unstable();
     format!(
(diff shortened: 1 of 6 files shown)


cchao-devo (approved):
Review of pick PR #389 — feat(agent): fact-first network discovery (inject subnets, scope hints)

  Verdict: APPROVE — correct, well-tested, and the #251 degradation signal is wired the right way.

  Good fix for the scan-the-wrong-subnet failure. Verified the pieces that matter and ran it locally.

  Verified

  - scan_scope_hint (pure) is correct and tested. On a zero-host sweep it names the scanned target, lists the connector's real subnets (or steers to target="auto"/discovery when
  none are known), and — the meaningful part — states raw-socket degradation explicitly rather than masking a degraded scan as an empty network. 3/3 tests, mutation-guarded.
  - The #251 degradation mapping is right (I checked the flag). unprivileged = !has_raw_socket_privilege() (nmap.rs:216), and the hint is built with raw_socket_available = 
  !unprivileged, so the "discovery was degraded" clause fires exactly when raw sockets were unavailable. An inverted mapping here would have re-introduced the masked-degradation
  bug; it doesn't.
  - Graceful degradation, no crash. The nmap path and the UI wrapper both fetch subnets via network_context() with unwrap_or_default() — on error, subnets are empty, the fact-first
  prompt section is omitted (bare persona returned, so pre-registration paths don't break), and the scope hint steers to auto.
  - Prompt injection of subnets is safe. Only the host's own CIDR strings are formatted into the system prompt (structurally constrained, host-derived, not attacker-controlled),
  and the section is dropped when empty.
  - Architecture is correct. The subnet fetch lives in the crates/ui wrapper (async) so pentest-core stays dependency-clean (core→tools is dev-only) — the right placement.
  - Gates green locally: agent_defaults 7/7, scan_scope_hint 3/3, and cargo clippy --workspace --all-targets -- -D warnings clean (the async-signature change ripples cleanly
  through both call sites).

  The nmap target param doc tightening ("multi-octet range … is rejected") is consistent with the ScopeGate/#384 behavior.

  Note

  AC #4 (behavioral — confirm the agent actually reads the injected subnets / uses target="auto" instead of guessing) is correctly deferred as a manual step: the prompt/hint
  machinery is unit-tested, but whether the model obeys the fact-first rule is a live-workspace check I can't run from here. Worth doing that manual pass before relying on it in
  the field.

PR #390, opened 2026-08-19
fix(tools): coerce float-valued booleans in param_bool (#307)

jtomek-strike48:
## Summary
- Completes the fix for #307. The string/int coercion for `param_bool` already landed on `main` in `1facaea`, but its number arm matches with `as_i64`, which returns `None` for a JSON float like `1.0` — so float-valued flags (`{"append": 1.0}`) still silently fell back to the default.
- Switch the number arm to `as_f64` (non-zero is true) so integer and float numbers coerce identically, matching what `param_u64`/`param_u16_opt` already do for the same reason (LLM tool callers emit numeric args as floats).
- Supersedes #382, which fixed the same underlying issue on a stale 474-commits-behind base and now conflicts with `main`'s existing coercion. This PR carries the one behavior `main` still lacked.

## Test plan
- [x] `cargo test -p pentest-tools --lib util::` — 5 tests pass
- [x] Mutation-verified: reverting the number arm to `as_i64` turns `param_bool_coerces_float_numbers` RED (`assertion failed ... 1.0`)
- [x] `cargo clippy -p pentest-tools --all-targets -- -D warnings` clean (exit 0)
- [x] `cargo fmt --all` applied
- [x] Blast radius: additive — real JSON bools, strings, and integers behave exactly as before; only previously-dropped float numbers now coerce. No caller relying on the default regresses.

### DIFF
--- a/crates/tools/src/util.rs
+++ b/crates/tools/src/util.rs
@@ -56,9 +56,9 @@ pub fn param_u16_opt(params: &Value, key: &str) -> Option<u16> {
 /// Extract a `bool` parameter with a default value.
 ///
 /// Accepts real JSON booleans, plus the string forms LLM tool callers emit
-/// (`"true"`/`"false"`/`"yes"`/`"no"`/`"1"`/`"0"`, case-insensitive) and the
-/// integers `1`/`0` — mirroring the loose coercion `param_u64` does for numbers.
-/// Unparseable or missing values yield `default`.
+/// (`"true"`/`"false"`/`"yes"`/`"no"`/`"1"`/`"0"`, case-insensitive) and numbers
+/// (`1`, `1.0`, `0`, `0.0`; non-zero is true) — mirroring the loose coercion
+/// `param_u64` does for numbers. Unparseable or missing values yield `default`.
 pub fn param_bool(params: &Value, key: &str, default: bool) -> bool {
     match params.get(key) {
         Some(Value::Bool(b)) => *b,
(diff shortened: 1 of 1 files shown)


cchao-devo (approved):
Review of pick PR #390 — fix(tools): coerce float-valued booleans in param_bool (#307)

  Verdict: APPROVE — correct, minimal fix on current main; supersedes #382.

  Closes the one remaining gap in main's param_bool: the number arm matched with as_i64, which returns None for a JSON float, so {"append": 1.0} silently fell back to the default.
  Switching to as_f64 (non-zero is true) makes integers and floats coerce identically, matching param_u64/param_u16_opt.

  Verified (ran it)

  - The fix is right. as_f64 on any Value::Number (int or float) returns Some, so 1→true, 0→false, 1.0→true, 0.0→false, 0.5→true, -1→true all resolve correctly; the None arm is a
  harmless defensive default (unreachable for std serde_json numbers). -0.0 → != 0.0 is false → false, which is sensible.
  - Additive / no regression. Real bools, strings, and integers behave exactly as before (integers already coerced via as_i64, now via as_f64 with the same result); only
  previously-dropped floats now coerce.
  - Test coverage closes the loop. The new param_bool_coerces_float_numbers guards the regression (mutation-verified per the description), and the fallback test now also pins
  empty-string and null → default, not silent false — exactly the edges worth locking for an LLM-facing coercion.
  - Local: util:: tests 5/5, cargo clippy -p pentest-tools -- -D warnings clean.

  Note
  
  This supersedes #382 (same underlying #307 fix, but on a 474-commits-stale base that now conflicts). Close #382 in favor of this. mergeable: true; the only gate is review.

PR #394, opened 2026-08-21
fix(connector): own the OTT callback origin, and keep WS query strings intact

edbonddevo:
Two independent faults that together kept a headless connector from completing against a multi-tenant studio. **Both are fixed client-side — neither needs a server or cluster change.**

Verified end to end against `jason-validation-v1.staging.strike48.engineering`.

---

## 1. The connector now owns its OTT callback origin

Vendored fork of `strike48-connector` 0.6.2. Upstream PR: **Strike48/sdk-rs#65** — drop the fork once that ships.

**Problem.** On post-approval registration the connector POSTs its public key to whatever host the server names in `CredentialsIssued.matrix_api_url`. `STRIKE48_API_URL` could only *veto* that value, never replace it, giving two reachable states: unset and follow the server anywhere, or set and be stranded.

Server-side that value falls back to a single cluster-global setting whenever the per-tenant Studio host is unavailable, and to `http://localhost:4001` when the deployment configures none. On the WebSocket transport the per-tenant path never fires at all, because `ConnectorSocket.connect/3` discards `connect_info`, so `realm_info` is always `nil`.

Two consequences:

- **A global cannot express multi-tenancy.** One value cannot name each tenant's callback host — configuring it for one tenant breaks the others.
- **Following the placeholder is a security problem.** `localhost:4001` resolves on the *connector's* machine. A credential exchange gets aimed at an unrelated local listener, over plaintext.

Observed before the fix:

```
OTT registration: api_url=http://localhost:4001
Failed to complete OTT registration: ... error sending request for url (http://localhost:4001/...)
```

and with the allowlist on, the correct-but-terminal:

```
OTT register URL origin ("http", "localhost", 4001) does not match STRIKE48_API_URL origin
("https", "<tenant>", 443); refusing to send credentials to an unapproved host
```

**Fix.** The connector already knows which host it dialed. It resolves its own base: `STRIKE48_API_URL` → the dialed host (scheme normalised, `connectors-` label and default port stripped) → the server value, last resort. Per-connector by construction. An empty `matrix_api_url` is no longer fatal; the origin allowlist is kept for the server-value-only case. Derivation mirrors `pentest_core::connector_registration::derive_api_url` so both ends agree.

## 2. WebSocket query strings survive the Studio proxy

**Problem.** `phoenix_shim.js` put the query string in the channel **join payload**, while `app_ws_proxy_channel` discards the payload and recovers the query by splitting the **topic** on `?` (`parse_path_and_query/1`). The topic was path-only, so every query param was silently dropped:

```
[WsOpen] connection_id=... path=/ws/shell query_string_len=0
[shell_ws] rejecting /ws/shell: missing or invalid token
Failed to connect to backend WS: HTTP error: 401 Unauthorized
```

This affected every query-bearing WebSocket from a Pick app; the shell just failed loudest because it's the one that authenticates.

**Fix.** Put the query in the topic — which the **deployed** server already parses, so this needs no matrix change. Both the `join` and the `send` topic-builders change together; if they diverge, frames route nowhere. The payload is retained: harmless today, correct if the server ever reads it.

## 3. `.gitignore`: `.env.bak.*`

`scripts/connect.sh` writes `.env.bak.<timestamp>`. The existing `.env.bak-*` rule uses a hyphen and does not match, leaving OTT-bearing backups committable.

---

## Verification

```
OTT callback base overridden: server advertised "http://localhost:4001",
  using "https://jason-validation-v1.staging.strike48.engineering"
OTT registration: full_url=https://jason-validation-v1.staging.strike48.engineering/api/connectors/register-with-ott
OTT registration response status: 200 OK (attempt 1)
Registered public key with OTT. Client ID: matrix:connector:non-prod:...:pentest-connector:pick-jason-v1
[status] Registered
```

Restart reuses the saved credential and mints a `private_key_jwt` against the tenant realm — no re-approval. Server-side the connector moves from `PendingRegistrationStore` into `Registry`. Chat then reaches the API and provisions its persona (`updated agent tools for pentest-connector`).

Upstream SDK suite: **379 lib tests pass, 0 failed** (8 new covering derivation, precedence, and the per-connector property); `cargo fmt --check` and `cargo clippy --lib --all-features` clean.

## Notes for reviewers

- **Behaviour change:** a deployment relying on the server redirecting connectors to a host other than the one they dialed must now set `STRIKE48_API_URL`. That redirect is the property being removed on purpose.
- **Vendoring scope:** only what `[patch.crates-io]` compiles — the lib. `tests/`, `test_fixtures/` and `examples/` are not vendored; upstream stays the source of truth. Changes are marked in-source with `[STRIKE48-PATCH connector-owns-callback-origin]`; audit with `git grep -n STRIKE48-PATCH vendor/`.
- **Not fixed here**, seen once during testing: a duplicate/concurrent `credentials_issued` can start a second registration that 401s on the already-claimed OTT and logs a failure even though the first succeeded. Benign but noisy — worth a separate look.
- The underlying server-side gaps (`ConnectorSocket` dropping `connect_info`; tenant HTTPRoutes not injecting `X-Matrix-Studio-Host`) are unaddressed and still worth fixing in matrix — this PR just stops the connector from depending on them.

---

## Review response (cchao-devo, approved-with-comments)

**Security hardening (commit `94afeed`, re-vendored from sdk-rs#65 `870d851`):**
- **#1** `resolve_register_base` validates the server-advertised value through `parse_origin` — was a trivial self-compare when `STRIKE48_API_URL` is unset.
- **#2** `register_with_ott` (pre-approval OTT-blob path) now enforces the origin allowlist — was previously unchecked.
- **#3** `resolve_register_base_verbose` emits **Warn** (survives the default warn-level EnvFilter) when the base is server-supplied-unpinned or a loopback origin.

**Nits:**
- **#4** PATCHES.md audit claim corrected — this fork mirrors the upstream PR branch and carries no in-tree markers; audit by diff vs crates.io 0.6.2 / sdk-rs#65.
- **#5** `derive_api_url` now strips the scheme default port, matching the SDK's `derive_api_base`.

## Scope disclosure (#6)
This PR carries three changes beyond the two headline faults, all intentional:
1. **`agent_defaults.rs` tool-optimizer disable** — required to make the connector's ~115 tools usable (the optimizer hid them behind find_tools/call_tool and models never invoked them). Adds ~115 tool schemas per pentest-agent call; relies on the tier models' ≥200k context (Qwen3/Claude on staging both qualify).
2. **`wsl.rs` `chunks_exact`→`as_chunks`** — unrelated clippy 1.98 fix; `main` was pre-1.98 and had not rebuilt.
3. **`docker-multiarch.yml` publishes on PRs** — infra-policy change: same-repo PR branches now publish `pr-N` images to ghcr (needed to hand off a testable image). `latest` stays gated to the default branch.

## Out-of-repo follow-up
Server-side `app_ws_proxy_channel.parse_path_and_query/1` should split the topic on the first `?` and treat the recovered query as untrusted. Tracked separately; not in this diff.


cchao-devo (approved):
Review of Strike48-public/pick PR #394 — "fix(connector): own the OTT callback origin, and keep WS query strings intact"

  Verdict: Approve with comments. The two headline fixes are correct and well-guarded, and this is the right client-side resolution of the multi-tenant callback problem (no
  server/cluster change needed). Local gates pass: cargo fmt --all --check, cargo clippy --workspace --locked --features pentest-platform/desktop-pcap -- -D warnings, and cargo 
  test -p pentest-core --lib (462/0) all green. Findings below are non-blocking but I'd harden #1 and #3 before merge — they're silent failure modes.

  Verified good

  - phoenix_shim WS fix is complete and can't diverge — both topic builders (_joinChannel :138, send :194) read the single _wsTopicPath computed once at :75; the "if they
  diverge, frames route nowhere" failure mode is structurally impossible. Retained payload is harmless as documented.
  - resolve_register_base precedence (explicit STRIKE48_API_URL → dialed host → server value) is per-connector by construction and closes the common-case localhost:4001
  plaintext-redirect hole. Absolute foreign register_url is refused and tested.
  - No TLS downgrade (wss/https/grpcs force https regardless of use_tls:false, tested); empty/blank server value fails loud (returns without POSTing), not silently.

  Security — worth hardening before merge

  1. Allowlist collapses to a self-check on the last-resort server_supplied path when STRIKE48_API_URL is unset (ott_provider.rs:257 → :850-852). When the base comes from the
  server value, enforce_same_origin compares the server value against itself (passes trivially), and the independent validate_register_origin is skipped with only a warn! when
  no override is set. Reachability is low (an underivable dialed host normally means no CredentialsIssued arrives), but the invariant isn't enforced. Suggest: on that branch,
  either refuse when STRIKE48_API_URL is unset, or run server_supplied through parse_origin in resolve_register_base itself, and emit a Warn when the base originated from the
  server value.
  2. Pre-approval flow (register_with_ott :534-540) has no origin allowlist at all — the hardening is confined to the post-approval CredentialsIssued path. A tampered OTT-blob
  api_url directs the token+public-key POST anywhere unchecked. Suggest running the resolved register_url through validate_register_origin here too, or documenting why it's
  intentionally exempt.
  3. Logging asymmetry (ott_provider.rs:281-293, logger.rs:38): the override note is Info, but the SDK's default EnvFilter is warn — so the "security-relevant redirect an
  operator should see in normal logs" is actually dropped at the default level. And the risky case (falling back to a server-advertised localhost) emits no note at all
  (advertised == base). Suggest Warn when the resolved base is a loopback/server-value origin, and reconcile the "visible in normal logs" comment with the warn default.

  Nits / scope-creep to disclose

  4. PATCHES.md audit claim is untrue as written — it says every change is marked in-source with [STRIKE48-PATCH <slug>] and auditable via git grep -n STRIKE48-PATCH 
  vendor/strike48-connector, but that grep returns only the PATCHES.md sentence itself; the patched sources carry zero markers. Either add the markers or correct the doc to the
  real audit method (diff vs crates.io 0.6.2 / sdk-rs#65).
  5. "Both ends agree" is only approximately true — pick's connector_registration::derive_api_url (:22) keeps the default port (…:443) while the vendored derive_api_base strips
  it. Functionally benign today (parse_origin normalizes default ports for the allowlist), but the vendored side's own rationale is that the port-carrying form trips
  hostname-based gateway routing — which is exactly what pick's version still emits. Align pick's to also strip, or soften the wording.
  6. Three changes ride outside the titled "two faults" — the agent_defaults.rs tool-optimizer disable (sound + tested, but adds ~115 tool schemas to every pentest-agent call;
  the "≥200k context" assumption is load-bearing and untested), the unrelated wsl.rs chunks_exact→as_chunks refactor, and docker-multiarch.yml now pushes images on every PR
  (removed the != 'pull_request' guards, push: true unconditional — an infra-policy change; same-repo branch PRs will now publish pr-N images). Please disclose these in the PR
  body.

  Out-of-repo to confirm
  
  - Server-side app_ws_proxy_channel.parse_path_and_query/1 should split the topic on the first ? and treat the recovered query as untrusted (the query can now legitimately
  contain literal ?). Not in this diff.

Issue #52, opened 2026-04-17
[Report Quality] Tool provenance & reproducible probe commands


jtomek-strike48:
> **Status: re-grounded against `main` on 2026-08-25.** The provenance schema, the fail-closed grounding gate (pick#184), report rendering, and redaction (pick#317) are already shipped. The remaining work is tool coverage (promoting provenance into the evidence graph), closing the easy-mode `document_write` leak, and wiring `autopwn_webapp`. Full evidence, re-scoped acceptance criteria, and the delivery plan are in [this comment](https://github.com/Strike48-public/pick/issues/52#issuecomment-5419785537). Owner: @jtomek-strike48.

## Priority
P0 (credibility with pentester persona)

## Effort
Small to Medium (original estimate covered the schema, now shipped; the remaining coverage work is larger - see re-scope comment)

## Credibility Impact
**Critical**

## Problem

Current Pick reports reference internal agent names in the methodology and evidence sections (e.g., `autopwn_webapp`, `default_creds_test`, `web_vuln_scan`, `port_scan`). These names do not identify:

- What underlying tool or technique was actually executed
- What version of that tool/ruleset was used
- The exact command, API call, or probe that was issued
- The raw response observed

A senior red teamer cannot reproduce a Pick finding from what is currently in the report. Seeing wrapper agent names where they expect tool names causes immediate loss of trust.

## Proposed Solution

Add required fields to every agent's finding output schema:

```yaml
finding:
  id: string
  title: string
  severity: enum
  provenance:
    underlying_tool: string        # e.g., "nuclei", "nikto", "httpx", "custom-s48-ssl", "curl"
    tool_version: string           # runtime-detected, e.g., "3.2.4"
    probe_commands:
      - command: string            # exact command as executed
        effective_command: string  # sanitized/redacted version safe for reports
        description: string        # optional one-line purpose
    raw_response_excerpt: string   # first N bytes/lines of actual target response
    timestamp: ISO8601
```

**`effective_command` handling:** Pentests routinely involve session tokens, cookies, internal URLs, or credentials that should not appear verbatim in a shareable report. The `effective_command` field provides a safe-to-publish version showing the command **structure** while `command` retains the exact form for internal traceability.

> Note (2026-08-25): this schema is implemented in `crates/core/src/provenance.rs` and wired into `EvidenceNode`/`ManifestFinding`. The open work is populating and promoting it from more tools, not defining it.

## Implementation Notes

[...]

PR #401, opened 2026-08-26
feat(evidence): promote web-wrapper provenance into evidence graph

jtomek-strike48:
## Summary

- Wires `web_vuln_scan`, `default_creds`, and `http_request` to promote their findings into the evidence graph. Each already built a full `Provenance` but never pushed it, so their findings could never reach a report (among the tool wrappers, only `nmap`/`whatweb`/`service_banner` promoted evidence beforehand; the `webwright` browser path has its own). Closes pick#52 **Gap A** for the three web wrappers - 3 of ~10; `port_scan`/`execute_command` and the external-subprocess wrappers (`zap`/`metasploit`/AD) follow in later PRs.
- Adds a shared `evidence_from_generic` builder plus per-wrapper adapters (`evidence_from_web_vuln_scan` / `_default_creds` / `_http_request`), so the `EvidenceNode` construction boilerplate lives in one place and each adapter only maps its own result shape.
- `default_creds` emits a node only per **successful** login (High severity); failed/errored attempts are not findings, and the working password is withheld from every node field.
- **Security fix folded in:** wiring `default_creds` evidence would otherwise carry a plaintext credential into the published report. `truncate_excerpt`'s redaction does not match JSON-quoted `"password":"..."` pairs (its secret regex expects `password:`, not `password":`), so passwords are now masked before reaching `raw_response_excerpt` (pick#52 / pick#317).
- `http_request` emits one Info node per response as a grounding anchor for its reproducible curl; the Validator can downgrade or drop it. (Trade-off: on a busy engagement this adds Info-node volume to the graph - accepted deliberately over adding classification logic to a primitive.)
- **Follow-up hardening (`ec7b129`):** closes two more report-bound secret sinks the first commit's redaction missed. `http_request` now runs the request URL through `redact()` before it fills the node `title`/`description`/`target`, so a `?api_key=...` query secret or `user:pass@` userinfo no longer reaches the report (the sibling provenance curl was already redacted). `default_creds` builds its probes via `from_exact_redacting_secret`, so the accepted SSH default password no longer survives in `effective_command` (the `sshpass -p '<pw>'` shape matches no `redact()` pattern), and the returned `data` attempts log is password-masked too, since the connector serializes it whole onto the Matrix wire.

## Test plan

- [x] `cargo test -p pentest-tools --lib` - unit tests for all three builders plus guards that the SSH probe, the returned `data`, and an `?api_key=` URL never leak a secret into a node field, excerpt, or `effective_command` (each guard-checked to go red when its fix is reverted). 407 passed.
- [x] `cargo test -p pentest-ui --test evidence_pipeline` - new `web_wrapper_findings_reach_the_report_after_validation` proves all three builders flow tool -> buffer -> graph -> gate -> report manifest; existing 4 pipeline tests still green.
- [x] `cargo clippy` on `pentest-tools` and `pentest-ui` with `-D warnings` (CI feature set `pentest-platform/desktop-pcap`) - clean.
- [x] `cargo fmt --all -- --check` - clean.
- Not run: live/network exercise of each wrapper's `execute()` tail (the push happens against a real target). The `execute()` tail is the mechanical mirror of the shipped `nmap.rs` wiring; builder correctness and the buffer->report flow are covered by the unit + integration tests above.

### DIFF
--- a/crates/tools/src/default_creds.rs
+++ b/crates/tools/src/default_creds.rs
@@ -15,17 +15,18 @@ use crate::util::{param_str, param_u64};
 /// Default credentials testing tool
 pub struct DefaultCredsTool;
 
-/// Serialize the attempt log for the provenance `raw_response_excerpt` with
-/// every password masked.
+/// Copy the attempt log with every password masked.
 ///
-/// The excerpt reaches the published report, and `truncate_excerpt`'s redaction
-/// pass does not catch JSON-quoted `"password":"..."` pairs (its secret regex
-/// expects `password:`, not `password":`), so a successful login would otherwise
-/// publish the plaintext credential in the evidence appendix (pick#52 /
-/// pick#317). Username and status are preserved so a reviewer still sees the
-/// attempt structure.
-fn masked_attempts_excerpt(attempts: &[Value]) -> String {
-    let masked: Vec<Value> = attempts
+/// The attempt log leaves this tool in two report-/wire-bound places: the
+/// provenance `raw_response_excerpt`, and the returned `ToolResult.data`, which
+/// the connector serializes whole onto the Matrix wire (`pick_connector.rs`
+/// `ToolCompleted`). `truncate_excerpt`'s redaction pass does not catch
+/// JSON-quoted `"password":"..."` pairs (its secret regex expects `password:`,
+/// not `password":`), so a successful login would otherwise publish the
+/// plaintext credential. Username and status are preserved so a reviewer still
+/// sees the attempt structure (pick#52 / pick#317).
+fn mask_attempts(attempts: &[Value]) -> Vec<Value> {
+    attempts
         .iter()
         .map(|a| {
             let mut a = a.clone();
@@ -36,8 +37,43 @@ fn masked_attempts_excerpt(attempts: &[Value]) -> String {
             }
             a
         })
-        .collect();
-    serde_json::to_string(&masked).unwrap_or_default()
+        .collect()
+}
+
+/// Build one report-safe [`ProbeCommand`] per attempted credential.
+///
+/// The probe records the *structure* of each auth attempt so a reviewer can
+/// reproduce it, but the password is scrubbed by exact value via
+/// [`ProbeCommand::from_exact_redacting_secret`] before it can reach the
+/// published `effective_command`. This matters most for the ssh
+/// `sshpass -p '<pw>'` shape, which no `redact()` regex matches (only
+/// `-u user:pass` / `--password` are caught) — see pick#52 / pick#317.
+fn default_cred_probes(
+    service: &str,
+    credentials: &[(&str, &str)],
+    host: &str,
+    port: u16,
+) -> Vec<ProbeCommand> {
+    let template = match service.to_lowercase().as_str() {
+        "ssh" => |u: &str, p: &str, h: &str, port: u16| {
+            format!("sshpass -p '{p}' ssh -o StrictHostKeyChecking=no -p {port} {u}@{h} exit")
+        },
+        "ftp" => |u: &str, p: &str, h: &str, port: u16| {
+            format!("curl --connect-timeout 5 -u {u}:{p} ftp://{h}:{port}/")
+        },
+        _ => |u: &str, p: &str, h: &str, port: u16| {
+            format!("curl -s -o /dev/null -w '%{{http_code}}' -u {u}:{p} http://{h}:{port}/")
+        },
+    };
+
+    credentials
+        .iter()
+        .map(|(u, p)| {
+            let full = template(u, p, host, port);
+            ProbeCommand::from_exact_redacting_secret(full, p)
+                .with_description("default credential probe")
+        })
+        .collect()
 }
 
 impl DefaultCredsTool {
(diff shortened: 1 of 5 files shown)


ngdevo:
Findings by severity
High — http_request Info-node flood blocks the report gate and can evict higher-severity evidence http_request is a primitive the agent calls constantly. evidence_from_http_request pushes one node per call (pr_http_request.rs:197), and the integration test confirms these nodes block gate_for_report as PendingNodes until validated. Consequences:
-- Every HTTP request in an engagement must be validated before any report can be generated; a busy engagement (hundreds–thousands of requests) creates a gate backlog that can stall report production.
-- push_evidence silently drops on MAX_EVIDENCE_NODES full (pick-review-evidence_producer.rs:93-99). If low-signal http_request Info nodes fill the buffer first, a real default_creds High-severity node pushed later can be silently dropped — the opposite of the PR's security intent. The let _ = push_evidence(...) swallow (all three wrappers) hides this. Recommendation: don't promote http_request by default (emit only on a flag / non-2xx / error), or cap/rate-limit, and at minimum warn (not ignore) when a High-severity push fails.
Medium — Severity::Info http_response nodes risk report noise Even when validated, every request URL (redacted) lands in the customer report manifest. The author accepts this trade-off, but given the gate impact above, the default of emitting-on-every-call should be reconsidered rather than relying on the Validator to prune.
Medium — global lock contention on a hot path push_evidence takes PENDING_EVIDENCE.write() (pick-review-evidence_producer.rs:91). Firing it from the high-frequency http_request primitive adds lock contention to a path that previously never touched the evidence buffer.
Low — mask_attempts only handles a top-level password key pr_default_creds.rs mask_attempts replaces exactly the "password" key. Fine for the current shape, but any future attempt schema with nested/password-like keys would bypass it (the truncate_excerpt JSON-quote gap this PR fixes suggests such shape drift happens). Consider masking by value or a deny-list of secret keys.
Low — secret scrubbing relies on exact password match from_exact_redacting_secret(full, p) redacts the literal p. A password that is a substring of the command/URL/username (e.g. exit, root) could over-redact or under-redact edge cases; acceptable for default-cred DB values but worth a comment.
Low/Nit — web_vuln_scan details/path are copied unmasked into the node evidence_from_web_vuln_scan puts details into description and path into metadata. These can contain response fragments or tokens. Mirrors existing nmap behavior, so not a regression, but the same redaction discipline applied to URLs/passwords here would be consistent.
Nit — base_url is reused as both scan target and raw data["url"] Fine, but confirm base_url is already validated/sanitized before it becomes a node target (it does flow through validate_url for http_request; verify web_vuln_scan does too).
Things done well
Shared evidence_from_generic removes per-builder boilerplate cleanly.
Guard tests are designed to go red on revert (default_cred_probes, URL-secret, full serialized-node assertion) — strong.
Security fixes are correctly applied to both sinks (provenance excerpt + wire data). default_creds correctly emits only on SUCCESS and withholds the password everywhere.
Verdict
Approve with requested changes (non-blocking on tests, blocking on the High item). The code is correct, well-factored, and well-tested; the secret-handling fixes are solid. Before merge I'd require addressing the http_request flooding concern (gate backlog + potential silent eviction of High-severity findings), since it undermines the PR's own security goal and can stall report generation. The web_vuln_scan/default_creds promotion itself is safe to ship.

jtomek-strike48:
Pushed `906e982`, which closes a third report-bound secret sink of the same class as the http_request URL fix in `ec7b129`.

**What:** `evidence_from_web_vuln_scan` (in `crates/tools/src/evidence_producer.rs`) copied the scanned base URL into the node `title`/`affected_target` and each finding's `details` into the node `description` verbatim, with no `redact()`. The sibling web_vuln_scan provenance curls for the same scan are already redacted, so the node fields were the weaker path.

**Impact:** a scan target carrying a secret in the URL - HTTP userinfo (`user:pass@host`) or a query token (`?api_key=...`) - would surface unredacted in published node fields, while its own provenance redacted it. Same asymmetry the earlier http_request fix addressed, in the other web wrapper.

**Fix:** `redact()` the target and each finding's `details` before they populate `title`/`description`/`affected_target`, plus a regression test (`test_evidence_from_web_vuln_scan_redacts_url_secret`) that fails on the pre-fix code and passes after.

**Verification:** `pentest-tools --lib` 408 pass, `evidence_pipeline` 5 pass, `clippy --all-targets -D warnings` clean, `fmt` clean (rustc 1.92, features `pentest-platform/desktop-pcap` + ui `connector,shell-ws`).

Note: the red checks on this PR are the pre-existing fork-PR ghcr package-write denial, not introduced by these commits.

cchao-devo (approved):
PR review: Strike48-public/pick#401 — feat(evidence): promote web-wrapper provenance into evidence graph (@jtomek-strike48, 5 files, +686/−32)

  Verdict: block — CI-only (rec concern, confidence 0.88). Code itself is merge-quality; see CI note.

  🤖 Automated review by ratchet review-pr — four dimensions scored; human judgment required.

  What it does — Wires web_vuln_scan, default_creds, http_request findings into the evidence graph via a shared GenericFinding/evidence_from_generic core, plus a folded-in secret-redaction
  hardening pass (pick#52 Gap A).

  Foundation — 5 pass · 1 nit · 2 n/a
  - nit — No linked issue: body writes Closes pick#52, which GitHub doesn't recognize. Fix: add Refs #52 / Refs #317 as proper keywords to the PR body.
  - Tests pass strongly: unit tests for all three builders, revert-red secret guards (default_creds.rs:420, evidence_producer.rs:1125/1154), and evidence_pipeline.rs integration proving tool →
    buffer → graph → gate → manifest.
    
  Architecture — 3 pass · 3 n/a — Builders land beside the existing evidence_from_* family; execute()-tail wiring mirrors the shipped service_banner.rs:301 pattern; no boundary/manifest changes
  needed; no config-in-code.

  Security — 3 pass · 3 n/a — Verified against main:
  - redact()'s regexes (provenance.rs:190-208) genuinely miss the sshpass -p '<pw>' shape, so the by-value scrub via from_exact_redacting_secret is the correct primitive, and its debug_assert
    canary catches silent no-ops.
  - mask_attempts masks once and feeds BOTH report-bound sinks (provenance excerpt + wire data).
  - URL userinfo / ?api_key= tokens are redact()ed before reaching published node fields; ProbeCommand.command is serde(skip_serializing) so exact plaintext never serializes.
  
  Adversarial — 5 pass · 3 n/a — No attacker leverage found. Target-supplied strings (base URL, finding details, final response URL) are all redacted before publication; probes are display-only,
  never executed; the http_request Info-node volume is bounded by the 10k buffer cap (acknowledged trade-off).

  ⚠️  The blocking item — infrastructure, not this diff:
  Build arm64 FAILURE · Android Build FAILURE · Build amd64 CANCELLED. Log inspection: Android fails on curl 404 fetching the proot package; arm64 fails at docker push ghcr.io/...: denied: 
  installation not allowed to Write organization package (compile succeeded). Both reproduce identically on main (2026-08-24/25 runs); Test/Clippy/Format are green on this PR.
  Fix: route through CI triage (triage-ci) for the proot fetch URL + ghcr org-package permission, then re-run — no author code change indicated. Merge state is BLOCKED until then regardless.

  Summary — Clean pass on all four review dimensions; the only asks are a one-line PR-body fix (Refs #52) and green Build CI, which is a repo-infra problem predating this branch. Recommend merging
  as soon as CI is fixed and re-run green.

Issue #362, opened 2026-08-12
docs(TOOLS.md): reconcile tool catalog with code + add a drift-preventing count assertion


jtomek-strike48:
## Problem

`docs/TOOLS.md` is out of sync with the code, in two ways:

1. **Missing tools.** ~16 shipped tools are absent from the catalog entirely, including high-visibility ones (metasploit, burpsuite, zap), core AD tooling (bloodhound, netexec, certipy, kerbrute), and high-blast-radius tools (autopwn_*, default_creds_test).
2. **Non-reconciling total.** The summary table's `24 native + 70 external = 94` matches neither its own itemized rows (~109) nor the code. Because the catalog is incomplete, no hand-picked total is correct.

This surfaced in the #357 review. As a stopgap, #357 replaced the falsifiable exact count with a code-verified `100+` floor across the docs, but did not reconcile the catalog itself (deliberately scoped out).

## Ground truth (reproducible)

```bash
# Total real tool implementations (excludes 3 #[cfg(test)] fixtures in tools.rs)
rg -oN "impl PentestTool for ([A-Za-z0-9_]+)" -r '$1' -g '*.rs' crates/ | sort -u
```

At time of filing: 123 `impl PentestTool` blocks minus 3 test-only fixtures (`UbiquitousTool`, `LinuxOnlyTool`, `NowhereTool`) = **120 real tools**.

Tools observed missing from `docs/TOOLS.md` (verify against the command above, which is the source of truth): bloodhound, burpsuite, certipy, kerbrute, netexec, zap, metasploit, webwright, arp_scan, autopwn_detect, autopwn_webapp, default_creds_test, export_session, ldapsearch_tool, safety_check, whois_tool.

## Proposed fix

1. Add the missing tools to `docs/TOOLS.md` under the right categories.
2. Reconcile the summary table (native/external/total) so its rows sum correctly, or derive it from code.
3. **Add a CI assertion** that counts `impl PentestTool` (excluding `#[cfg(test)]`) and fails if the documented total drifts. This closes the doc-code drift class rather than patching one number, so future tool additions can't silently desync the catalog again.

## Related

- Follow-up from #357 (docs accuracy).
- Related to the dynamic catalog EPIC #250, but distinct: this is near-term accuracy of the static `TOOLS.md` plus a drift guard; #250 is the live BlackArch-driven catalog.

jtomek-strike48:
[...]

PR #403, opened 2026-08-26
docs(tools): reconcile TOOLS.md with registry + drift guard

jtomek-strike48:
## Summary

- Rebuilds `docs/TOOLS.md` from the exact `name()` strings `create_tool_registry()` returns, fixing three kinds of drift found while investigating #362.
- Adds `crates/tools/tests/tools_doc_sync_test.rs` (`tools_doc_sync`), an integration test that fails on drift in either direction, so the catalog cannot silently desync again.
- Runs in the required `cargo test --workspace` Linux CI lane (plain `#[test]`, no feature gate, no `integration` tag), so it is an actual gate, not a green-but-unrun check.

## What was wrong (verified against the registry, not the diff)

The static catalog had desynced from `create_tool_registry()` three ways at once:

1. **12 registered tools had no row:** `bloodhound`, `burpsuite`, `zap`, `metasploit`, `certipy`, `netexec`, `kerbrute`, `webwright`, `safety_check`, `cyberchef`, `http_request`, `autopwn_webapp`.
2. **11 rows used a name the connector never registers** (so calling the documented name 404s): `arpscan`->`arp_scan`, `ldapsearch`->`ldapsearch_tool`, `whois`->`whois_tool`, `session_export`->`export_session`, `default_creds`->`default_creds_test`, `impacket_*`->`impacket-*` (psexec/wmiexec/secretsdump/getuserspns), `evilwinrm`->`evil-winrm`, `aircrackng`->`aircrack-ng`.
3. **4 rows for tools that are implemented but never registered** (removed from the catalog): `credential_harvest`, `lateral_movement`, `autopwn_orchestrator`, `autopwn_network_plan`.

The summary table (`24 native + 70 external = 94`) matched neither its rows nor the code. It is replaced with a `Totals` section that points at the registry and the ground-truth command instead of a hand-kept number.

## Why the guard keys off the registry, not `impl PentestTool`

The issue proposed counting `impl PentestTool` (excluding `#[cfg(test)]`). That is the wrong source of truth: it yields 121, which over-counts the shipped catalog by 4 (the implemented-but-unregistered tools above) and misses that "shipped" depends on a build feature (`inject-test-evidence`) and a runtime capability (pcap). The issue's own reconciliation (`123 - 3 fixtures = 120`) had itself drifted - it missed a 4th `#[cfg(test)]` fixture, `StubTool`. The test asserts the doc's name set equals `create_tool_registry().names()`, with a small allowlist for the two conditionally-registered tools.

## Follow-up (not in this PR)

`credential_harvest`, `lateral_movement`, `autopwn_orchestrator`, and `autopwn_network_plan` have full `impl PentestTool` blocks but are never added to `create_tool_registry()`, so they are unreachable at runtime. Removing their doc rows is correct for the catalog, but the implementations being dead is a separate question (wire them up or delete them) worth its own issue.

## Test plan

- [ ] `cargo test -p pentest-tools --test tools_doc_sync_test` passes.
- [ ] Neutered check: reverting `docs/TOOLS.md` to its prior content turns both `tools_doc_sync` tests red (the guard actually guards).
- [ ] `cargo fmt --all --check` and `cargo clippy -p pentest-tools --all-targets -- -D warnings` clean.
- [ ] Confirm the test executes in the required Linux CI lane (it is not `integration`-tagged and not feature-gated).

### DIFF
--- a/crates/tools/tests/tools_doc_sync_test.rs
+++ b/crates/tools/tests/tools_doc_sync_test.rs
@@ -1,0 +1,89 @@
+//! Guard against drift between `docs/TOOLS.md` and the live tool registry.
+//!
+//! Regression guard for pick#362: the static catalog silently desynced from
+//! `create_tool_registry()` in three ways at once - registered tools with no
+//! catalog row, catalog rows whose name did not match the registered `name()`,
+//! and catalog rows for tools that were never registered. A hand-maintained
+//! count could not catch any of that (and its own arithmetic had drifted too).
+//!
+//! These tests derive the source of truth from the registry - not from a
+//! hand-kept total - and fail on drift in either direction. The tool name a
+//! catalog row must use is the exact string returned by that tool's `name()`,
+//! which is what an operator or the model invokes; a row whose first backticked
+//! token is not a real `name()` is a landmine, not documentation.
+
+use std::collections::BTreeSet;
+use std::path::PathBuf;
+
+use pentest_tools::create_tool_registry;
+
+/// Tools whose registration is conditional on a build feature or a runtime
+/// capability. They may be absent from the registry when this test runs, yet
+/// are still legitimately listed in the catalog, so they are exempt from the
+/// "documented but not registered" direction only:
+///
+/// - `traffic_capture` is registered only when packet capture (libpcap on
+///   Linux, Npcap on Windows) is available at runtime
+///   (`pentest_platform::is_pcap_available()`).
+/// - `inject_test_evidence` is gated behind the opt-in `inject-test-evidence`
+///   feature and is never registered in release builds (pick#184).
+const CONDITIONAL_TOOLS: &[&str] = &["traffic_capture", "inject_test_evidence"];
+
+/// The first backticked token of every Markdown table row in `docs/TOOLS.md`.
+/// Tool rows look like ``| `nmap` | External | ... |``; header, separator, and
+/// category-summary rows do not start with `` | ` `` and are ignored.
+fn documented_tool_names() -> BTreeSet<String> {
+    let path = PathBuf::from(env!("CARGO_MANIFEST_DIR")).join("../../docs/TOOLS.md");
+    let doc = std::fs::read_to_string(&path)
+        .unwrap_or_else(|e| panic!("failed to read {}: {e}", path.display()));
+
+    doc.lines()
+        .filter_map(|line| {
+            let rest = line.trim_start().strip_prefix("| `")?;
+            let end = rest.find('`')?;
+            Some(rest[..end].to_string())
+        })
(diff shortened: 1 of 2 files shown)


cchao-devo (approved):
PR review: Strike48-public/pick#403 — docs(tools): reconcile TOOLS.md with registry + drift guard (@jtomek-strike48, 2 files, +183/−40)

  Verdict: block — CI-only (rec concern, confidence 0.77). Code + docs are publishable; the merge-block is the pre-existing Android CI failure.

  🤖 Automated review by ratchet review-pr — four dimensions scored; human judgment required.

  What it does — Rebuilds docs/TOOLS.md from create_tool_registry().names() (picking up 12 missing tools, fixing 11 wrong names under which tools were documented but unreachable, removing 4 dead rows
  for tools that lack registry entries), plus an integration test (tools_doc_sync_test.rs) that catches drift in either direction against the registry.

  Foundation — 4 pass · 1 nit · 3 n/a
  - nit — No Fixes #362 / Refs #362 keyword in body (prose-only "found while investigating #362"), and all 4 test-plan checkboxes are - [ ] though CI shows the claims holding. Fix: add Fixes #362 and
    tick the boxes.
  - Tests are excellent — the guard asserts both directions (every_registered_tool_is_documented + every_documented_tool_is_real), correctly excludes the two conditional tools (traffic_capture,
    inject_test_evidence), and runs in the required Linux CI lane.

  Architecture — 3 pass · 3 n/a — Test under crates/tools/tests/next to the crate it guards; consumes only the already-pub create_tool_registry(); no manifest or export changes.

  Security — 3 pass · 3 n/a — Secrets/binary/injection surfaces all clean (it's docs + a test with no user input).

  Adversarial — 2 pass · 6 n/a — The parsed input is the repo-tracked markdown file (CI, not attacker-controllable); assert messages echo the set but no shell/SQL/template interpolation; no product
  runtime path changed.

  ⚠️  The blocking item — infrastructure, not this diff:
  Android Build FAILURE — nix develop --command just fetch-proot → curl: (22) 404. Same root cause as pick#401 (and on main 2026-08-24/25/26). All 23 other checks green.
  Fix: triage the proot fetch URL; no author code change needed.

  One additional nit:
  The new header says "115 registered in a default build" — a hand-kept number the drift guard won't protect (it checks name sets, not counts), contradicting the new Totals section's self-professed "no
  hand-maintained total here on purpose." Fix: drop the number or date-stamp it ("115 as of 2026-08-26 — see Totals").

  Summary — Clean, well-explained PR that fixes three real classes of doc-vs-code drift and installs a genuine gate. The only blocking item is the repo-wide Android CI failure predating this branch.
  Recommend merging once Android is fixed.

Issue #405, opened 2026-08-27
Decide fate of unregistered native credential_harvest + lateral_movement tools


jtomek-strike48:
## Problem

Two native post-exploitation tools are fully implemented but **not registered** in `create_tool_registry()`, so they are unreachable at runtime. Unlike the autopwn pair (safe re-wire, tracked separately), these two are a **security decision**, not a merge-fix - each needs a call on whether to rework-and-wire or delete.

- `credential_harvest` (`CredentialHarvestTool`, `crates/tools/src/credential_harvest.rs`, 404 lines)
- `lateral_movement` (`LateralMovementTool`, `crates/tools/src/lateral_movement.rs`, 522 lines)

Both are complete implementations. Both were registered when introduced (`a454b4c`, "feat: add credential harvesting and lateral movement tools") and lost their `registry.register(...)` line in a later silent merge resolution, never restored. Only their `pub use` imports remain.

## Why each needs a decision, not a blind re-wire

**`credential_harvest` - operator self-harvest footgun.** Its `execute()` harvests from the **local host it runs on**: `~/.ssh` private keys, WiFi passwords, environment secrets, and config files, on a "compromised host" assumption. Pick runs on the *operator's* machine, so registering it unchanged would harvest the operator's own credentials into the evidence graph and the report - the same "reflects Pick's own host" class as the safety_check self-reflection issue. It must be re-scoped to an explicit remote/compromised target (or hard-gated) before it is safe to register. Relates to #313 (flag discovered credentials as sensitive-but-logged) and #162 (differential-authz identities).

**`lateral_movement` - high blast radius.** It is an auto-orchestrator for SSH-key reuse, credential reuse, Pass-the-Hash, and tunneling across hosts. It is not redundant with the registered impacket / evil-winrm / crackmapexec tools (it is a layer above them), but re-enabling auto-PtH deserves explicit sign-off and tests. Its PtH path was already OS-gated in #204. It is coupled by design to `credential_harvest` (consumes "harvested credentials").

## Decision needed

For each tool, pick one:

1. **Rework + wire up** - credential_harvest: re-scope away from self-harvest; lateral_movement: security review + regression tests, then register. Re-add `docs/TOOLS.md` rows (the `tools_doc_sync` guard requires it once registered).
2. **Delete** - remove the impls and `pub use` lines if the harvest->move workflow no longer fits the current architecture (which now leans on the external impacket/crackmapexec tools + fact-first discovery).

[...]

PR #407, opened 2026-08-27
fix(tools): re-register autopwn_detect and autopwn_network_plan

### LINKED ISSUE #406: Re-register accidentally un-wired autopwn tools (autopwn_detect, autopwn_network_plan)
## Problem

Two native autopwn tools are fully implemented but are **not registered** in `create_tool_registry()` (`crates/tools/src/lib.rs`), so they are unreachable at runtime:

- `autopwn_detect` (`AutoPwnOrchestratorTool`, `crates/tools/src/autopwn/orchestrator.rs`) - detects hardware and recommends WiFi-vs-Network strategy.
- `autopwn_network_plan` (`AutoPwnNetworkPlanTool`, `crates/tools/src/autopwn/network_plan.rs`, 485 lines, 7 unit tests) - builds a phased discovery/scan/exploitation plan.

Both are complete implementations, and neither is superseded by any registered tool. Only their `pub use` imports remain; the `registry.register(...)` calls are gone.

## Live symptom

`crates/ui/src/components/dashboard.rs:120` ships an "AutoPwn" quick-action whose prompt tells the agent to *"Plan a full network penetration test (autopwn_network_plan)"*. Because the tool is not registered, the agent is instructed to call a tool that does not exist and gets an unknown-tool error. `autopwn_detect`'s own `next_step` output also points at `autopwn_network_plan`.

## Root cause

Both were registered when introduced (`autopwn_detect` in `9285b02`, `autopwn_network_plan` in `369dc11`, 2026-03-12). Their `registry.register(...)` lines were silently dropped in the 3-way merge `f280c01` ("Merge main into toolchain branch - resolve conflicts by keeping all tools", 2026-03-19), which unioned the `pub use` imports but lost the register calls with no conflict marker - directly contradicting its own commit message. This is accidental merge loss, not deprecation. (`git log -S` on `lib.rs` misses it because pickaxe skips merge diffs.)

## Fix

1. Re-register `AutoPwnOrchestratorTool` and `AutoPwnNetworkPlanTool` in `create_tool_registry()`.
2. Re-add their rows to `docs/TOOLS.md` (name `autopwn_detect`, not the old `autopwn_orchestrator`) - the `tools_doc_sync` guard from #362/#403 requires this once they are registered.
3. Add a registration-presence test so a future merge cannot silently drop them again (root-cause closure). Note the existing `all_tools_registered` test (see #392) checks registry->capabilities, not impl->registry, so it does not cover this class.

## Related

- Discovered during #362 / PR #403 (tool-catalog reconciliation + `tools_doc_sync` guard).
- #392 (registration guard does not run on the macOS lane).
- Sibling decision for the other two dropped-and-never-restored native tools (`credential_harvest`, `lateral_movement`) is tracked in #405 - those are security-gated, this one is a safe re-wire.

jtomek-strike48:
## Summary

- Re-registers `autopwn_detect` (`AutoPwnOrchestratorTool`) and `autopwn_network_plan` (`AutoPwnNetworkPlanTool`) in `create_tool_registry()`. Both are complete implementations whose `register()` lines were silently dropped in merge `f280c01`; only their `pub use` imports survived.
- Re-adds their two `docs/TOOLS.md` rows, which the `tools_doc_sync` guard from #403 now requires once they are registered.
- Adds `crates/tools/tests/tools_registration_test.rs` to guard against this class of silent un-registration recurring.

Stacked on #403 (`docs/tools-catalog-drift-362`); base should be retargeted to `main` once #403 merges. Closes #406.

## Why (root cause + live symptom)

`autopwn_detect` (`9285b02`) and `autopwn_network_plan` (`369dc11`) were registered when introduced, then lost their `registry.register(...)` lines in the 3-way merge `f280c01` ("Merge main into toolchain branch - resolve conflicts by keeping all tools"), which unioned the imports but dropped the register calls with no conflict marker. `git log -S` on `lib.rs` misses it because pickaxe skips merge diffs.

Live symptom (Lens 5 consumer trace): `crates/ui/src/components/dashboard.rs:120` ships an "AutoPwn" quick-action whose prompt tells the agent to *"Plan a full network penetration test (autopwn_network_plan)"*. With the tool unregistered, the agent is told to call a tool that does not exist. Re-registering resolves it.

## Guard design

`tools_registration_test` pins the two known cases the existing guards miss:

- `all_tools_registered` (crates/core) checks registry -> capabilities; `tools_doc_sync` (#403) checks registry -> docs. Neither catches an `impl PentestTool` whose `register()` line was dropped.
- `autopwn_orchestration_tools_are_registered` asserts both autopwn tools are present (regression guard).
- `security_gated_tools_remain_unregistered` asserts `credential_harvest` and `lateral_movement` stay unregistered, pending the security decision in #405 - a deliberate tripwire so they are not wired up without resolving that issue first.

Residual (honest): this pins named tools, not "every impl is registered." A general guard would need source parsing or a macro-based registry; that is the over-broad `impl PentestTool`-count approach #403 deliberately avoided. Out of scope here.

## Test plan

- [ ] `cargo test -p pentest-tools --test tools_registration_test` passes (autopwn registered; credential_harvest/lateral_movement still gated).
- [ ] `cargo test -p pentest-tools --test tools_doc_sync_test` still passes (rows added for the newly registered tools).
- [ ] Neutered check: deleting one `register()` line turns `autopwn_orchestration_tools_are_registered` red.
- [ ] `cargo fmt --all --check` and `cargo clippy -p pentest-tools --all-targets -- -D warnings` clean.
- [ ] Runs in the required `cargo test --workspace` Linux lane (plain `#[test]`, no feature gate).

### DIFF
--- a/crates/tools/src/lib.rs
+++ b/crates/tools/src/lib.rs
@@ -114,9 +114,11 @@ pub fn create_tool_registry() -> ToolRegistry {
     // WiFi tools
     registry.register(WifiScanTool);
     registry.register(WifiScanDetailedTool);
+    registry.register(AutoPwnOrchestratorTool); // Hardware detection + strategy (autopwn_detect)
     registry.register(AutoPwnPlanTool);
     registry.register(AutoPwnCaptureTool);
     registry.register(AutoPwnCrackTool);
+    registry.register(AutoPwnNetworkPlanTool); // Network-mode attack planning
 
     // Vulnerability assessment
     registry.register(ServiceBannerTool);
--- a/crates/tools/tests/tools_registration_test.rs
+++ b/crates/tools/tests/tools_registration_test.rs
@@ -1,0 +1,47 @@
+//! Registration guards: tools that must (or must not) be in the runtime registry.
+//!
+//! `all_tools_registered` (crates/core) checks registry -> capabilities, and the
+//! `tools_doc_sync` guard checks registry -> docs. Neither catches a tool whose
+//! `impl PentestTool` exists but whose `registry.register(...)` line was dropped,
+//! for example silently lost in a merge resolution. These tests pin the two
+//! known cases so the regression cannot recur unnoticed.
+
+use pentest_tools::create_tool_registry;
+
+/// `autopwn_detect` and `autopwn_network_plan` were registered when introduced,
+/// then silently lost their `register()` lines in the 3-way merge f280c01
+/// (pick#406). A shipped dashboard quick-action prompt directs the agent at
+/// `autopwn_network_plan`, so leaving them unregistered is a live unknown-tool
+/// bug, not just dead code. Guard against the regression recurring.
+#[test]
+fn autopwn_orchestration_tools_are_registered() {
+    let registry = create_tool_registry();
+    let names = registry.names();
+    for tool in ["autopwn_detect", "autopwn_network_plan"] {
+        assert!(
+            names.contains(&tool),
+            "tool '{tool}' is implemented but not registered in create_tool_registry() \
+             (pick#406) - its register() line was likely dropped in a merge again. \
+             Registered: {names:?}"
+        );
+    }
+}
+
+/// `credential_harvest` and `lateral_movement` are implemented but deliberately
+/// NOT registered, pending a security decision (pick#405): credential_harvest
+/// harvests the operator's own host credentials as written, and lateral_movement
+/// is an auto-Pass-the-Hash orchestrator. This is a tripwire, not an oversight -
+/// if you register either, resolve pick#405 first (rework/scope it) and remove
+/// the corresponding entry here.
+#[test]
+fn security_gated_tools_remain_unregistered() {
+    let registry = create_tool_registry();
+    let names = registry.names();
+    for tool in ["credential_harvest", "lateral_movement"] {
+        assert!(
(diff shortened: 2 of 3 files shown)


jtomek-strike48:
Local review notes (verified empirically before requesting approval):

Verified by running, not reading:

- `cargo test -p pentest-tools --test tools_registration_test` -> 2 passed.
- Neuter check: deleting one `registry.register(...)` line in `crates/tools/src/lib.rs` turns `autopwn_orchestration_tools_are_registered` red with the expected panic; restored after. The guard fails-closed as intended, which is the whole point of this PR.
- `security_gated_tools_remain_unregistered` asserts the correct direction: `credential_harvest` and `lateral_movement` stay unregistered pending #405.
- `tools_doc_sync_test` (#403 guard) still passes with the two new TOOLS.md rows.
- `cargo fmt --all --check` clean; `cargo clippy -p pentest-tools --all-targets -- -D warnings` clean.
- New test is a plain `#[test]` with no feature gate; confirmed present in the Linux `test` job invocation.

All test-plan boxes check out. One pre-existing note (not introduced here): the `check-macos` lane scopes to `pentest-platform`/`pentest-core` and does not run `pentest-tools` tests, so this guard executes only on the Linux lane.

Issue #397, opened 2026-08-25
fix(ci): pin proot/talloc to immutable URLs so Android fetch-proot stops 404ing


jtomek-strike48:
## Problem

The `Android Build` CI lane fails at `just fetch-proot`:

```
curl: (22) The requested URL returned error: 404
error: Recipe `fetch-proot` failed with exit code 22
```

`fetch-proot` (in `justfile`) pulls **version-pinned** `.deb`s from the Termux package repo:

```
curl -fSL --retry 3 "${termux_repo}/p/proot/proot_${proot_version}_${arch}.deb"
curl -fSL --retry 3 "${termux_repo}/libt/libtalloc/libtalloc_${talloc_version}_${arch}.deb"
```

The Termux main repo is a **rolling** repo: it only serves the *latest* build of each package. When upstream supersedes the pinned `proot_version` / `talloc_version`, the old URL 404s and the lane goes red - with no change on our side.

## Evidence (introduced by external drift, not a code change)

- Green at `498fc35` (2026-08-19), red at `db5ecdd` (2026-08-24) - the range in between touches nothing in the Android build or this recipe.
- Failing run: CI run `32763830027`, `Android Build` job `97548646425`.
- Confirmed non-blocking for releases: the release workflow's publish job excludes `build-android`/`build-ios` from its `needs`, so v0.1.10 still published desktop/headless/web. This is CI hygiene, not a release blocker.

## Impact

- `Android Build` will 404 on every `main` push until the pin is refreshed or the source is changed.
- Advisory lane (no required status check on `main`), so it does not gate merges - but it's a permanent red that masks real Android regressions.

## Proposed fix

Stop sourcing release inputs from a mutable rolling repo:

1. Mirror the pinned proot/talloc `.deb`s (and any other Termux inputs) to an **immutable** location we control - a GitHub release asset or the same hosted-artifact pattern used for the desktop openat2 proot (#252).
2. Fetch from that mirror with `curl --fail` + a **checksum** verification step (aligns with #372).
3. Stopgap if a mirror isn't ready: bump `proot_version` / `talloc_version` to the current Termux builds to get the lane green now.

## Related

- #372 - sandbox/package downloads lack integrity checks (this adds the *availability/pinning* half to that *integrity* work).
- #252 - productionize openat2 proot as a CI build + checksummed hosted artifact (same "host our own artifact" model, applied to the Android/Termux proot).
- #387 - mobile Easy Mode experimental; stable Android build before CI gating.

PR #408, opened 2026-08-27
feat: surface sandbox backend fallback to the operator (#242)

### LINKED ISSUE #242: Surface sandbox backend fallback (bwrap<->proot) to the operator, not just logs
## Summary

When the interactive PTY shell falls back from one sandbox backend to its sibling on an EACCES spawn (#239's symmetric `fallback_backend_for`: bwrap <-> proot), the swap is only recorded via `tracing::warn!`. There is no operator-facing signal that the sandbox backend changed under them.

## Problem

The sandbox backend is security-relevant (bwrap = kernel namespaces vs proot = ptrace-mediated syscall interception; different isolation properties). An operator who believes they are in one backend but is silently running in the other has an inaccurate mental model of their isolation boundary. Today the only evidence is a log line.

## Proposed change

Surface the effective backend to the session/UI when a fallback occurs. Options (pick during design):

- Emit a one-line notice into the shell session banner (e.g. "sandbox backend: bwrap (fell back from proot)").
- Expose the effective backend in the connector/session status the UI already renders.
- Both.

## Related gap (from #239 review)

The `drop(pair) -> re-openpty -> spawn(sibling)` fallback control flow in `PtyShell::spawn_sandboxed` currently has no behavioral test (the triggering EACCES process-context is not reproducible in a bare test binary; only the pure decision function `fallback_backend_for` is unit-tested). If this work adds a seam that makes the fallback observable, add coverage for the swap path here too, or document the manual verification procedure.

## Context

Follow-up from #239 (requested by cchao-devo as an optional follow-up). Not a blocker for that PR.

## Acceptance criteria

- [ ] When the shell falls back to the sibling backend, the effective backend is visible to the operator (not just in logs).
- [ ] Fallback swap path has a behavioral test, or a documented manual-verification step.

ngdevo:
## Problem

Before this change, when the shell fell back from one sandbox backend to its sibling (bwrap ↔ proot) due to EACCES or early init exit, the swap was only recorded via \`tracing::warn!\`. The operator had no visible indication that the sandbox backend changed under them. This is security-relevant because bwrap (kernel namespaces) and proot (ptrace-mediated syscall interception) have different isolation properties.

## Solution

- Add \`primary_backend\` and \`effective_backend\` fields to \`PtyShell\` to track the selected and actually used sandbox backend
- Expose \`effective_backend()\` and \`primary_backend()\` accessor methods
- Set \`effective_backend\` on successful sandboxed spawn and fallback spawn paths
- Push a terminal notice ("Sandbox backend: X (fell back from Y)") when the shell falls back to a sibling backend
- Add unit test verifying native shells have no backend
- Document manual verification procedure for fallback terminal notice

## Files changed

- \`crates/platform/src/desktop/pty_shell.rs\` — add backend tracking fields and accessors
- \`crates/ui/src/shell_ws.rs\` — push terminal notice on backend fallback

## Verification

- \`cargo fmt --all -- --check\` passes
- \`cargo test -p pentest-platform\` passes (92 passed, 6 ignored)
- New unit test \`native_shell_has_no_backend\` passes
- Fallback survival invariant already covered by the existing \`sandboxed_shell_is_alive_after_backend_fallback\` integration test (ignored by default; requires provisioned rootfs)

## Related

- Closes #242
- Follow-up to #239 (symmetric bwrap ↔ proot fallback)

### DIFF
--- a/crates/platform/src/desktop/pty_shell.rs
+++ b/crates/platform/src/desktop/pty_shell.rs
@@ -15,6 +15,8 @@ use std::path::Path;
 pub struct PtyShell {
     pair: portable_pty::PtyPair,
     child: Box<dyn portable_pty::Child + Send + Sync>,
+    primary_backend: Option<SandboxBackend>,
+    effective_backend: Option<SandboxBackend>,
 }
 
 /// Return true if a `portable_pty` spawn error was caused by `EACCES`
@@ -65,6 +67,17 @@ fn fallback_backend_for(backend: SandboxBackend, err: &anyhow::Error) -> Option<
 }
 
 impl PtyShell {
+    /// Return the sandbox backend that was actually used to spawn this shell,
+    /// if any. `None` for native (non-sandboxed) shells.
+    pub fn effective_backend(&self) -> Option<SandboxBackend> {
+        self.effective_backend
+    }
+
+    /// Return the sandbox backend that was selected before any fallback, if any.
+    pub fn primary_backend(&self) -> Option<SandboxBackend> {
+        self.primary_backend
+    }
+
     /// Spawn a new interactive shell using the system default program.
     /// If `cwd` is provided, the shell will start in that directory.
     /// When `shell_mode` is `Proot`, spawns inside a sandboxed BlackArch rootfs
@@ -107,7 +120,12 @@ impl PtyShell {
             .spawn_command(cmd)
             .map_err(|e| Error::ToolExecution(format!("Failed to spawn shell: {e}")))?;
 
-        Ok(Self { pair, child })
+        Ok(Self {
+            pair,
+            child,
+            primary_backend: None,
+            effective_backend: None,
+        })
     }
 
     /// Spawn a sandboxed interactive shell using bwrap, proot, or WSL.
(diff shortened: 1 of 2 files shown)


jtomek-strike48 (approved):
Reviewed at head 5a83b48 in an isolated worktree; ran the gates rather than reasoning about the diff. Approving - clean, tightly-scoped, both ACs met.

Verified sound:
- `cargo check -p pentest-ui --features connector,shell-ws,desktop` compiles the notice path (the shell-ws/liveview-gated code the Docker Build jobs and --workspace lanes exercise; `SandboxBackend: Display` backs the formatting).
- Guard-check: neutering the native path (`primary_backend: None -> Some(Proot)`) turns `native_shell_has_no_backend` red; restore -> green. Genuine guard.
- Notice fires only when `primary != effective`, i.e. exactly the bwrap<->proot fallback case; `shell_ws.rs:99` is the sole production `PtyShell::spawn` caller, so no fallback path misses it.

Two non-blocking notes:
1. The fallback branch (shell_ws.rs:107) has no automated coverage - only the native-None case is tested; the notice relies on the documented manual step (permitted by AC#2). If you want to close it cheaply later, the decision is a pure fn of two `Option<SandboxBackend>` - extracting `fn fallback_notice(primary, effective) -> Option<String>` makes it unit-testable without a rootfs.
2. The notice lands in the Logs/activity page, not inline in the shell PTY stream - matches the issue's "session status the UI renders" option, just confirming that's the intended surface.

The red Android Build is the known external fetch-proot 404 drift (#397) - advisory, not this PR.

Issue #243, opened 2026-07-16
SandboxManager reports stale backend after PTY shell EACCES fallback (config drift)


jtomek-strike48:
## Summary

The global `SANDBOX_MANAGER` (`crates/platform/src/desktop/sandbox/mod.rs`) is initialized once with a single selected `backend` and a single corresponding executor (`bwrap_executor` / `proot_executor`, the others `None`). When the interactive PTY shell falls back to the sibling backend on an EACCES spawn (#239), that fallback is **local to the shell** - the shared manager still reports the originally selected backend.

## Problem

Consequences of the drift:

- `manager.backend()` lies about what the fallback shell actually runs (misleading for logging/metrics/dispatch).
- The **non-interactive** `execute_command` path (`SandboxManager::execute`) keeps using the originally selected backend's executor. If the primary backend (e.g. bwrap) passes `is_available()` but fails at runtime, the interactive shell recovers via the sibling but tool execution via `execute_command` does not - so a shell and a tool run in the same session can end up on different backends, and the tool path can still hit the failure the shell just worked around.
- Any future proot-only pre-spawn work added to `spawn_sandboxed` would run for a bwrap fallback shell too (currently benign because the pacman.conf fixups are bwrap-compatible).

Not a bug in current behavior, but a foot-gun: manager config can silently diverge from what a session is actually using.

## Proposed change

Options (pick during design):

- Push the EACCES fallback down into `SandboxManager` so the first successful backend selection persists globally (both interactive and non-interactive paths agree).
- Or, make backend selection re-validate a candidate by actually attempting a spawn during `detect_backend`, so a present-but-unspawnable backend is rejected at selection time and never becomes the manager's backend.
- At minimum, document the local-only nature of the shell fallback in the manager docs so the divergence is intentional and visible.

## Context

Follow-up from #239 (finding M3 from cchao-devo's review). Not a blocker for that PR.

## Acceptance criteria

- [ ] Interactive shell and non-interactive `execute_command` agree on the effective backend within a session, OR the divergence is explicitly documented as intended.
- [ ] `manager.backend()` reflects the backend actually in use, or its scope is documented.

jtomek-strike48:
[...]

PR #409, opened 2026-08-27
fix: propagate PTY shell EACCES fallback to global SandboxManager (#243)

ngdevo:
## Problem

The global \`SANDBOX_MANAGER\` (\`crates/platform/src/desktop/sandbox/mod.rs\`) is initialized once with a single selected \`backend\` and a single corresponding executor. When the interactive PTY shell falls back to the sibling backend on an EACCES spawn (#238), that fallback is **local to the shell** — the shared manager still reports the originally selected backend.

### Consequences of the drift

- \`manager.backend()\` lies about what the fallback shell actually runs (misleading for logging/metrics/dispatch).
- The **non-interactive** \`execute_command\` path (\`SandboxManager::execute\`) keeps using the originally selected backend's executor. If the primary backend (e.g. bwrap) passes \`is_available()\` but fails at runtime, the interactive shell recovers via the sibling but tool execution via \`execute_command\` does not — so a shell and a tool run in the same session can end up on different backends, and the tool path can still hit the failure the shell just worked around.
- Any future proot-only pre-spawn work added to \`spawn_sandboxed\` would run for a bwrap fallback shell too (currently benign because the pacman.conf fixups are bwrap-compatible).

## Root Cause

The fallback logic lives entirely in \`PtyShell::try_fallback_spawn\` (\`pty_shell.rs\`) and updates only the shell-local state (\`effective_backend\`). The \`SandboxManager\` is immutable after construction (fields are plain \`Option\` values with no interior mutability), so there is no mechanism to propagate the fallback globally.

## Solution

Push the EACCES fallback into \`SandboxManager\` with interior mutability:

1. Wrapped mutable \`SandboxManager\` state (\`backend\`) in \`std::sync::Mutex<ManagerState>\` so the selected backend can be updated after construction.
2. Added \`SandboxManager::try_fallback_to()\` — updates the stored backend atomically. Executors are constructed on-demand in \`execute()\` and \`ensure_ready()\` from the current backend, avoiding stale cached state.
3. In \`PtyShell::spawn_sandboxed\`, after a successful \`try_fallback_spawn\`, call \`manager.try_fallback_to(fallback)\` so the global state is consistent.
4. \`execute_command\` (\`command.rs\`) automatically benefits — it reads \`manager.backend()\` and dispatches at call time, so it now uses the fallback backend.

### Design decisions

- **Why not store executors in \`ManagerState\`?** The initial implementation stored executors, but review showed they were never read: both \`execute()\` and \`ensure_ready()\` constructed fresh instances. Keeping them would be dead store and add unnecessary \`Clone\` derives. The current design keeps only the backend in mutex-protected state and constructs the executor inline — simple, correct, and no stale-cache risk.
- **Why not re-validate in \`detect_backend\`?** Re-validating by actually attempting a spawn during \`detect_backend\` would require a real rootfs and bwrap invocation at manager-initialization time, which is expensive, fragile, and duplicates the existing EACCES recovery. The runtime fallback is the right place because:
  - The rootfs may not be ready yet during \`detect_backend\`
  - EACCES is a per-process-context issue, not a per-backend availability issue
  - The shell already has the symmetric fallback logic; this change simply makes the global manager aware of it
- **Poison safety:** All mutex access goes through \`lock_state()\`, which converts a \`PoisonError\` into a \`SandboxError\` instead of panicking. This prevents a single thread panic from killing the entire process's ability to dispatch sandbox commands.

## Testing

- \`cargo fmt --all\` — clean
- \`cargo clippy --all-targets -- -D warnings\` — clean
- \`cargo test --lib -p pentest-platform\` — 91 passed, 8 ignored

## Files Changed

- \`crates/platform/src/desktop/sandbox/mod.rs\` — Mutex-wrapped state, \`try_fallback_to()\`, \`lock_state()\` poison-safe helper, updated \`backend()\`/\`ensure_ready()\`/\`execute()\`
- \`crates/platform/src/desktop/pty_shell.rs\` — call \`try_fallback_to()\` on both fallback paths (spawn-error + early-exit)
- \`crates/platform/src/desktop/sandbox/{bwrap,proot,wsl,docker}.rs\` — \`#[derive(Clone)]\` for executor structs (useful for future cached-executor patterns)

### DIFF
--- a/crates/platform/src/desktop/pty_shell.rs
+++ b/crates/platform/src/desktop/pty_shell.rs
@@ -269,17 +269,17 @@ impl PtyShell {
                     )
                     .await?
                     {
-                        manager
-                            .try_fallback_to(fallback)
-                            .await
-                            .map_err(|e| {
-                                tracing::error!(
-                                    "[PtyShell::spawn_sandboxed] failed to update sandbox manager to {fallback:?}: {e}"
-                                );
-                                Error::ToolExecution(format!(
-                                    "Sandboxed shell fell back to {fallback:?} but failed to update the sandbox manager: {e}"
-                                ))
-                            })?;
+                        // Best-effort global sync: the shell already spawned and
+                        // verified healthy on `fallback`, so a bookkeeping failure
+                        // here (only reachable on lock poison) must not tear it
+                        // down. Log and hand back the working shell; the worst case
+                        // is `execute_command` dispatching on the pre-fallback
+                        // backend, which the manager's own EACCES recovery covers.
+                        if let Err(e) = manager.try_fallback_to(fallback) {
+                            tracing::error!(
+                                "[PtyShell::spawn_sandboxed] shell fell back to {fallback:?} but failed to update the sandbox manager: {e}; continuing with the working shell"
+                            );
+                        }
                         return Ok(shell);
                     }
                     // fallback unavailable: fall through to the hard error below.
@@ -321,17 +321,17 @@ impl PtyShell {
                     )
                     .await?
                     {
-                        manager
-                            .try_fallback_to(fallback)
-                            .await
-                            .map_err(|e| {
-                                tracing::error!(
-                                    "[PtyShell::spawn_sandboxed] failed to update sandbox manager to {fallback:?}: {e}"
-                                );
-                                Error::ToolExecution(format!(
-                                    "Sandboxed shell fell back to {fallback:?} but failed to update the sandbox manager: {e}"
-                                ))
-                            })?;
+                        // Best-effort global sync: the shell already spawned and
+                        // verified healthy on `fallback`, so a bookkeeping failure
+                        // here (only reachable on lock poison) must not tear it
+                        // down. Log and hand back the working shell; the worst case
+                        // is `execute_command` dispatching on the pre-fallback
+                        // backend, which the manager's own EACCES recovery covers.
+                        if let Err(e) = manager.try_fallback_to(fallback) {
+                            tracing::error!(
+                                "[PtyShell::spawn_sandboxed] shell fell back to {fallback:?} but failed to update the sandbox manager: {e}; continuing with the working shell"
+                            );
+                        }
                         return Ok(shell);
                     }
                 }
--- a/crates/platform/src/desktop/pty_shell.rs
+++ b/crates/platform/src/desktop/pty_shell.rs
@@ -269,6 +269,17 @@ impl PtyShell {
                     )
                     .await?
                     {
+                        manager
+                            .try_fallback_to(fallback)
+                            .await
+                            .map_err(|e| {
+                                tracing::error!(
+                                    "[PtyShell::spawn_sandboxed] failed to update sandbox manager to {fallback:?}: {e}"
+                                );
+                                Error::ToolExecution(format!(
+                                    "Sandboxed shell fell back to {fallback:?} but failed to update the sandbox manager: {e}"
+                                ))
+                            })?;
                         return Ok(shell);
                     }
                     // fallback unavailable: fall through to the hard error below.
@@ -310,6 +321,17 @@ impl PtyShell {
                     )
                     .await?
                     {
+                        manager
+                            .try_fallback_to(fallback)
+                            .await
+                            .map_err(|e| {
+                                tracing::error!(
+                                    "[PtyShell::spawn_sandboxed] failed to update sandbox manager to {fallback:?}: {e}"
+                                );
+                                Error::ToolExecution(format!(
+                                    "Sandboxed shell fell back to {fallback:?} but failed to update the sandbox manager: {e}"
+                                ))
+                            })?;
                         return Ok(shell);
                     }
                 }
--- a/crates/platform/src/desktop/sandbox/bwrap.rs
+++ b/crates/platform/src/desktop/sandbox/bwrap.rs
@@ -18,7 +18,6 @@ use tokio::sync::OnceCell;
 static BWRAP_USABLE: OnceCell<bool> = OnceCell::const_new();
 
 /// Bubblewrap executor for Linux namespace-based sandboxing
-#[derive(Debug, Clone)]
 pub struct BwrapExecutor {
     config: SandboxConfig,
 }
--- a/crates/platform/src/desktop/sandbox/bwrap.rs
+++ b/crates/platform/src/desktop/sandbox/bwrap.rs
@@ -18,6 +18,7 @@ use tokio::sync::OnceCell;
 static BWRAP_USABLE: OnceCell<bool> = OnceCell::const_new();
 
 /// Bubblewrap executor for Linux namespace-based sandboxing
+#[derive(Debug, Clone)]
 pub struct BwrapExecutor {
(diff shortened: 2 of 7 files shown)


jtomek-strike48 (commented):
## Review summary

Correct, well-scoped fix that meets both acceptance criteria of #243: the shell's bwrap->proot EACCES fallback now propagates to the global `SandboxManager`, so non-interactive `execute_command` dispatch agrees with the interactive session. No blocking issues. One should-fix (the fix's own regression guard never runs in CI) plus a cluster of cheap low-severity robustness nits.

Verified empirically: fetched the head into an isolated worktree at merge-base `b9a43d92` (origin/main), built and ran tests with the 1.92 toolchain (MSRV 1.92), mutation-tested the new tests, and traced every consumer of the changed API across `crates/`.

### Verified sound
- Swap propagates: `pty_shell.rs:120` and `command.rs:161` both use the same global manager via `get_sandbox_manager()`; `try_fallback_to` mutates the shared `Arc`, and `execute()` reads `self.lock_state()?.backend` at dispatch time. AC1 + AC2 met.
- Dropping `ensure_proot` from `new()` is not a download regression. `execute()`'s Proot arm now uses `get_proot_path` (existence-only). Traced every path that sets `backend == Proot`: `detect_backend` downloads proot itself before returning `Proot` (mod.rs:240), and the runtime fallback is gated by `ProotExecutor::is_available` plus a prior `ensure_ready()`/rootfs setup (`rootfs.rs:644`) that already fetched it. So proot is always present before `get_proot_path` runs.
- `ensure_ready` refactor (early-returns -> `match`) is behavior-preserving; Bwrap and Proot share the `_` arm.
- Tests genuinely guard the fix: ran both new tests with `--ignored` (2 passed), then neutered the swap in `try_fallback_to` and `try_fallback_to_switches_backend` went red (`left: Proot, right: Bwrap`); the idempotent test stayed green (early-returns before the mutated line).
- init-dev headless is a no-op: `execute_command_in_dir` early-returns to `execute_command_direct` when `!is_sandbox_enabled()` (DISABLE_SANDBOX), never touching `SandboxManager`.
- `std::sync::Mutex` is never held across an `.await` in the diff (correct choice); the removed `.expect("executor not initialized")` is a strict improvement.
- Not stale (0 commits behind main). The failing Android Build is pre-existing and unrelated: it dies at step "Fetch Android proot runtime" with `curl (22) 404` on an external asset, and this diff touches zero Android files.

### SHOULD-FIX

**1. The #243 regression guard has zero CI coverage.** `sandbox/mod.rs` (the two `try_fallback_to_*` tests). Both are `#[ignore]`d because they build a real manager via `SandboxManager::new()` -> `detect_backend`, and the required CI job never passes `--ignored`, so the switch logic ships with no green signal (the "8 ignored" in the PR body includes these two). The swap itself is pure mutex state mutation. Suggest a `#[cfg(test)]` constructor that builds a `SandboxManager` with a fixed backend (no detection), then a non-ignored test asserts `try_fallback_to` swaps and is idempotent so the guard actually runs on `main`.

### LOW (cheap, recommended, non-blocking)

**2. A working fallback shell is discarded if the manager-sync fails.** `pty_shell.rs:254-266` and `:306-318`. After `try_fallback_spawn` returns an already-spawned, early-exit-verified live `shell`, `manager.try_fallback_to(fallback).map_err(...)?` turns a bookkeeping failure into a fatal `Error::ToolExecution`, tearing down the working shell. This contradicts the method's own "best-effort" doc comment. Only reachable on mutex poison, which is currently unreachable (no critical section can panic), hence low, but the fix is strictly better: log and `return Ok(shell)` regardless of sync outcome. Pairs with #3.

**3. `backend()` silently fabricates `Bwrap` on lock poison.** `sandbox/mod.rs:274-278`: `.unwrap_or(SandboxBackend::Bwrap)` with no log, while sibling `execute()`/`ensure_ready()` propagate the same poison via `?`. Consumers include `pty_shell.rs:135` (spawn dispatch). Same unreachable-poison precondition as #2, so low today, but a latent trap (the infallible-looking getter is the natural one to reach for). Return a `Result`, or at minimum `tracing::error!` before defaulting; do not hardcode `Bwrap`.

**4. Speculative `#[derive(Clone)]` on all four executors.** `bwrap/proot/wsl/docker.rs`. None of them are ever `.clone()`d (grep-confirmed); all are built via `::new()`. The added `Clone` contradicts the PR's own decision not to store executors. Suggest dropping it.

**5. `which proot` subprocess per command (perf).** `execute()` -> `get_proot_path` -> `system_proot_exists()` spawns `which proot` on every proot-backed command; the old cached executor resolved it once. Proot-backend only, minor next to the sandboxed command.

**6. Non-atomic triple lock-read of `state.backend` in `execute()`.** Three separate lock acquisitions per call; a concurrent `try_fallback_to` could flip between them. Benign today (the fallback set is only the Bwrap/Proot sibling pair, which share the `ensure_ready` `_` arm); would bite if the pairing expands.

**7. Polish.** `try_fallback_to` is `async` with no `.await`; it trusts the caller to have pre-verified availability; and its poison error reuses `SandboxError::RootfsSetupFailed` (semantically off).

Net: gate merge only on #1 (make the guard run in CI). #2 and #3 are cheap and worth folding into the same round; #4-#7 are optional.

jtomek-strike48:
Addressed the review in a follow-up commit (rebased on current `main`).

**SHOULD-FIX**

1. **#243 guard now runs in CI.** Added a detection-free `#[cfg(test)]` constructor `SandboxManager::with_backend_for_test(backend)` and rewrote both `try_fallback_to_*` tests as non-ignored `#[test]`s. The swap is pure mutex-state mutation, so they no longer need a real backend and give a green signal on `main`. Verified: both run and pass in the default (non-`--ignored`) set.

**LOW (folded in)**

2. **Working fallback shell is no longer discarded on sync failure.** Both `pty_shell.rs` fallback paths now log and `return Ok(shell)` on a `try_fallback_to` error instead of turning a live, early-exit-verified shell into a fatal `ToolExecution` error. Matches the method's best-effort contract.
3. **`backend()` no longer fabricates `Bwrap` silently.** It `tracing::error!`s the poison before defaulting; signature stays infallible so spawn-dispatch/logging callers are unaffected.
4. **Dropped the speculative `#[derive(Clone)]`** on all four executors (never cloned; contradicted the on-demand-executor decision).
6. **`execute()` reads the backend once** and dispatches on that snapshot, closing the non-atomic multi-read window.
7. **Polish:** `try_fallback_to` is now synchronous (did no async work; callers updated), and the poison path has its own `SandboxError::StateLockPoisoned` variant instead of reusing `RootfsSetupFailed`.

**Not changed**

5. `which proot` per proot-backed command: left as-is. Caching the resolved path would reintroduce the cached-executor state the PR deliberately removed; the cost is proot-only and negligible next to the sandboxed command.

Validation: `cargo fmt --all` clean, `cargo clippy -p pentest-platform --all-targets -- -D warnings` clean, sandbox tests pass (incl. the two now-non-ignored guards), `cargo check --all-targets` clean across the workspace. The Android Build lane is still the pre-existing external `fetch-proot` 404, unrelated to this diff.

edbonddevo (approved):

PR #410, opened 2026-08-27
fix: reconcile sandbox switches, gate custom tool installers on sandbox state, surface sandbox chip in UI (#244)

### LINKED ISSUE #244: Signal sandbox-disabled state in Tools UI; reconcile DISABLE_SANDBOX vs shell_mode

ngdevo:
## Description

Fixes #244.

When `DISABLE_SANDBOX=true`, users currently see a wall of per-tool installation errors with no indication of why, and no way to tell which sandbox mode is active. This PR addresses three root causes.

---

## Problem 1: Two unsynchronized sandbox switches

`DISABLE_SANDBOX` (env var) and `shell_mode` (UI setting in `Settings` → `ShellMode`) both write to the same `USE_SANDBOX` atomic bool with no coordination. `init_sandbox_from_env()` runs lazily via `std::sync::Once`, while the UI calls `set_use_sandbox()` directly from `workspace_app.rs` and `router.rs`. Depending on call order, one silently overrides the other — sessions can be in a contradictory state.

### Resolution

**`crates/platform/src/desktop/command.rs`**
- Added `DISABLED_BY_ENV: AtomicBool` — set to `true` when `DISABLE_SANDBOX` is detected at startup.
- `set_use_sandbox()` checks this flag first and returns immediately (with a `warn!` log) when the env has locked the sandbox off. The env setting **always wins**.
- `is_sandbox_enabled()` short-circuits to `false` when `DISABLED_BY_ENV` is set, regardless of what the UI shell-mode toggle may have written to `USE_SANDBOX`.

---

## Problem 2: Custom installers advertise as auto-installable in native mode

`CatalogEntry::is_auto_installable()` returned `true` unconditionally for all `InstallMethod::Custom` entries. The custom installers themselves (bloodhound, zap, metasploit, certipy, netexec, kerbrute) check `!sandbox_enabled()` in their `install()` method and return an error with manual instructions when the sandbox is off. So the UI shows an "Install" button, the user clicks it, and gets a per-tool failure — instead of seeing the manual instructions up front.

### Resolution

**`crates/tools/src/installers/mod.rs`**
- Added `fn sandbox_required(&self) -> bool { true }` to the `ToolInstaller` trait. Returns `true` by default (safest assumption).

**`crates/tools/src/installers/webwright.rs`**
- `WebwrightInstaller` overrides `sandbox_required() -> false` — it installs via pip/uv on the host and works natively.

**`crates/tools/src/catalog.rs`**
- `CatalogEntry::is_auto_installable()` for `Custom { id }` entries now checks:
  - If sandbox is enabled → auto-installable (can run `pacman` inside).
  - If sandbox is disabled → auto-installable only when `!installer.sandbox_required()`.
  - This means bloodhound, zap, metasploit, certipy, netexec, kerbrute correctly show manual instructions instead of an "Install" button when the sandbox is off. Webwright remains installable in both modes.

---

## Problem 3: No sandbox state visible in the UI

The settings page had no indicator showing whether the sandbox is enabled or which mode is active. Users had no way to understand why some tools showed manual instructions or why "Install all recommended" produced no visible result.

### Resolution

**`crates/ui/src/components/settings_page.rs`**
- Added a **sandbox status chip** in the Tools card header:
  - 🟢 **"Sandboxed"** — shown when `is_sandbox_enabled()` is true. This is the correct signal because it reflects the `DISABLE_SANDBOX` override, not just the `shell_mode`/`sandbox_available` inputs.
  - 🟡 **"Native (host)"** — shown when the sandbox is disabled (DISABLE_SANDBOX, Native mode, or no backend available).
- The **"Install all recommended"** button is now disabled when there are no recommended, auto-installable, not-yet-installed tools. Previously it only checked `installing().is_some()`, so clicking it when sandbox was off would silently do nothing (no error, no install, no feedback).

**`crates/ui/src/components/css/settings_page.css`**
- Added `.sandbox-badge`, `.sandbox-enabled`, and `.sandbox-disabled` styles.

---

## Changes by file

| File | Change |
|------|--------|
| `crates/platform/src/desktop/command.rs` | +33/-2 — `DISABLED_BY_ENV` atomic, `set_use_sandbox` no-op when env disabled, `is_sandbox_enabled` short-circuit |
| `crates/tools/src/installers/mod.rs` | +12/-0 — `sandbox_required()` trait method (default `true`) |
| `crates/tools/src/installers/webwright.rs` | +7/-0 — override `sandbox_required() -> false` |
| `crates/tools/src/catalog.rs` | +15/-1 — `is_auto_installable()` checks `!installer.sandbox_required()` when sandbox off |
| `crates/ui/src/components/settings_page.rs` | +30/-1 — sandbox status chip, disable button when nothing auto-installable |
| `crates/ui/src/components/css/settings_page.css` | +22/-0 — `.sandbox-badge` styles |

## Testing

- `cargo test -p pentest-platform --lib` — 91 passed (including 7 command tests)
- `cargo test -p pentest-tools --lib -- "catalog::tests"` — 20 passed
- `cargo test -p pentest-tools --lib -- "installers::tests"` — 4 passed

All pre-existing test suites unaffected.

### DIFF
--- a/crates/platform/src/desktop/command.rs
+++ b/crates/platform/src/desktop/command.rs
@@ -53,6 +53,9 @@ fn init_sandbox_from_env() {
 /// variable permanently disabled the sandbox — the env setting always wins.
 pub fn set_use_sandbox(use_sandbox: bool) {
     // Ensure the env has been read before we check DISABLED_BY_ENV.
+    // `init_sandbox_from_env` is backed by `std::sync::Once`, so after the
+    // first call this is a single atomic load — cheap enough for every
+    // UI-shell-mode toggle.
     init_sandbox_from_env();
 
     if DISABLED_BY_ENV.load(Ordering::SeqCst) {
--- a/crates/platform/src/desktop/command.rs
+++ b/crates/platform/src/desktop/command.rs
@@ -19,7 +19,16 @@ use tokio::time::timeout;
 /// Can be disabled via DISABLE_SANDBOX=true environment variable.
 static USE_SANDBOX: AtomicBool = AtomicBool::new(true);
 
-/// Initialize sandbox state from environment on first access
+/// Whether the sandbox was *permanently* disabled by the `DISABLE_SANDBOX`
+/// environment variable. When `true`, [`set_use_sandbox`] becomes a no-op so
+/// the env setting cannot be overridden by the UI shell-mode toggle.
+static DISABLED_BY_ENV: AtomicBool = AtomicBool::new(false);
+
+/// Initialize sandbox state from environment on first access.
+///
+/// If `DISABLE_SANDBOX=true` (or `=1`) is set, both `USE_SANDBOX` and
+/// `DISABLED_BY_ENV` are set so that later [`set_use_sandbox`] calls are
+/// silently ignored.
 fn init_sandbox_from_env() {
     use std::sync::Once;
     static INIT: Once = Once::new();
@@ -28,7 +37,10 @@ fn init_sandbox_from_env() {
         if let Ok(val) = std::env::var("DISABLE_SANDBOX") {
             if val.to_lowercase() == "true" || val == "1" {
                 USE_SANDBOX.store(false, Ordering::SeqCst);
-                tracing::info!("Sandbox disabled via DISABLE_SANDBOX environment variable");
+                DISABLED_BY_ENV.store(true, Ordering::SeqCst);
+                tracing::info!(
+                    "Sandbox disabled via DISABLE_SANDBOX environment variable; UI toggle will be ignored"
+                );
             }
         }
     });
@@ -36,7 +48,22 @@ fn init_sandbox_from_env() {
 
 /// Set whether to use sandbox for command execution.
 /// Call this based on the user's ShellMode setting.
+///
+/// This is a **no-op** (with a warning) when the `DISABLE_SANDBOX` environment
+/// variable permanently disabled the sandbox — the env setting always wins.
 pub fn set_use_sandbox(use_sandbox: bool) {
+    // Ensure the env has been read before we check DISABLED_BY_ENV.
+    init_sandbox_from_env();
+
+    if DISABLED_BY_ENV.load(Ordering::SeqCst) {
+        if use_sandbox {
+            tracing::warn!(
+                "Ignoring set_use_sandbox(true): sandbox was disabled via DISABLE_SANDBOX env var"
+            );
+        }
+        return;
+    }
+
     USE_SANDBOX.store(use_sandbox, Ordering::SeqCst);
     tracing::info!(
         "Sandbox execution: {}",
@@ -47,6 +74,13 @@ pub fn set_use_sandbox(use_sandbox: bool) {
 /// Check if sandbox is enabled
 pub fn is_sandbox_enabled() -> bool {
     init_sandbox_from_env();
+
+    // When the env disabled the sandbox, always return false regardless of
+    // what the UI may have subsequently tried to set.
+    if DISABLED_BY_ENV.load(Ordering::SeqCst) {
+        return false;
+    }
+
     USE_SANDBOX.load(Ordering::SeqCst)
 }
 
--- a/crates/platform/tests/disable_sandbox_precedence.rs
+++ b/crates/platform/tests/disable_sandbox_precedence.rs
@@ -1,0 +1,48 @@
+//! Regression guard for #244 Problem 1: `DISABLE_SANDBOX` must take precedence
+//! over the UI shell-mode toggle. Once the environment has disabled the
+//! sandbox, `set_use_sandbox(true)` (driven by the desktop/web shell-mode
+//! setting) is a no-op and `is_sandbox_enabled()` stays `false`.
+//!
+//! This is pure in-memory logic (an env var plus process-global atomics): it
+//! provisions no rootfs, spawns no backend, and touches no network. Unlike the
+//! backend harnesses in this directory it is therefore NOT `#[ignore]`d and
+//! runs in the required CI job (`cargo test --workspace ... --features
+//! pentest-platform/desktop-pcap`, which also sets `DISABLE_SANDBOX=true`).
+//!
+//! It lives in its own integration binary on purpose: `init_sandbox_from_env`
+//! reads `DISABLE_SANDBOX` exactly once via `std::sync::Once`, so the env var
+//! must be set before anything else in the process touches the sandbox API.
+//! A dedicated single-test binary guarantees that ordering.
+//!
+//! Desktop-gated: the real `set_use_sandbox` / `is_sandbox_enabled` live in
+//! `pentest_platform::desktop`; on non-desktop feature builds the crate root
+//! exposes no-op stubs, so there is no precedence to guard there.
+#![cfg(feature = "desktop")]
+
+#[test]
+fn env_disable_wins_over_ui_shell_mode_toggle() {
+    // Set before the first sandbox call so the `init_sandbox_from_env` `Once`
+    // latches the env-disabled state. Safe on edition 2021.
+    std::env::set_var("DISABLE_SANDBOX", "true");
+
+    // The env has disabled the sandbox: it reports off.
+    assert!(
+        !pentest_platform::is_sandbox_enabled(),
+        "DISABLE_SANDBOX=true must report the sandbox disabled"
+    );
+
+    // The UI shell-mode toggle tries to switch it back on. The env wins, so
+    // this is a no-op and the sandbox stays off.
+    pentest_platform::set_use_sandbox(true);
+    assert!(
+        !pentest_platform::is_sandbox_enabled(),
+        "set_use_sandbox(true) must not override DISABLE_SANDBOX=true"
+    );
+
+    // An explicit shell-mode "off" is consistent and also leaves it off.
+    pentest_platform::set_use_sandbox(false);
+    assert!(
+        !pentest_platform::is_sandbox_enabled(),
+        "sandbox must remain disabled after set_use_sandbox(false)"
+    );
+}
--- a/crates/tools/src/catalog.rs
+++ b/crates/tools/src/catalog.rs
@@ -1250,4 +1250,65 @@ mod tests {
             );
         }
     }
+
+    #[test]
+    fn custom_installer_gating_respects_sandbox_state_when_disabled() {
+        // Regression guard for #244 Problem 2. With the sandbox disabled, a
+        // custom installer that can ONLY install inside the sandbox
+        // (bloodhound/zap/metasploit/certipy/netexec/kerbrute) must not be
+        // auto-installable, so the UI shows manual instructions instead of an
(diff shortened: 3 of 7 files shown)


jtomek-strike48:
Reviewed by checking out the head (`e73d6a4`) against base `b9a43d9` in an isolated worktree and running the real gates, not by reading the diff. The implementation is sound: the atomics have no TOCTOU and `set_use_sandbox` now correctly calls `init_sandbox_from_env()` (closing a real pre-existing hole where the UI could silently clobber `DISABLE_SANDBOX`); `install_all_recommended` filters by `is_auto_installable()` so the bulk install no longer produces a per-tool error wall; the non-auto-installable path falls through to the manual-instructions card. Three of the four acceptance criteria (#2, #3, #4) are met.

One gap worth closing before merge: the two behaviors this PR exists to deliver have no regression guard.

## SHOULD-FIX: missing regression coverage for both fixes

**Problem 2 gating is unguarded.** Reverting `is_auto_installable()`'s Custom branch (`crates/tools/src/catalog.rs:119-133`) back to the old `Custom { .. } => true` leaves the entire `pentest-tools` suite green. The rewritten `catalog_item_maps_state_and_category_to_strings` uses `webwright`, which is auto-installable in either sandbox state by construction, so it cannot distinguish the fix from the bug.

**Problem 1 precedence is unguarded.** `command.rs` has 7 tests, none referencing `DISABLED_BY_ENV` / `set_use_sandbox` / `is_sandbox_enabled`. The "env always wins" precedence has no test.

Both guards below were verified empirically: each passes on the current head, and each goes red under the exact pre-fix it guards (Custom `=> true` for the first; removing the `DISABLED_BY_ENV` short-circuits for the second). Both clear `cargo fmt --check` and `cargo clippy -- -D warnings`, and both run in the required CI job (`cargo test --workspace ... --features pentest-platform/desktop-pcap`, which sets `DISABLE_SANDBOX=true`).

### 1. Append to `crates/tools/src/catalog.rs` `mod tests`

```rust
    #[test]
    fn custom_installer_gating_respects_sandbox_state_when_disabled() {
        // Regression guard for #244 Problem 2. With the sandbox disabled, a
        // custom installer that can ONLY install inside the sandbox
        // (bloodhound/zap/metasploit/certipy/netexec/kerbrute) must not be
        // auto-installable, so the UI shows manual instructions instead of an
        // "Install" button that fails per-tool. A custom installer that works
        // natively (webwright, pip/uv on the host) stays auto-installable.
        //
        // Without this, reverting is_auto_installable()'s Custom branch to the
        // old `Custom { .. } => true` leaves the whole suite green.
        //
        // The required CI test job runs with DISABLE_SANDBOX=true, so the
        // sandbox-off branch is what actually executes there; set_use_sandbox
        // is called anyway so the guard also holds when run locally.
        pentest_platform::set_use_sandbox(false);

        // Precondition: bloodhound is a registered, sandbox-required installer,
        // so the negative assertion below exercises the sandbox_required gate
        // rather than merely an unregistered id (which also yields false). Also
        // fails loudly if the installer is ever renamed or its default flips.
        assert!(
            get_installer("bloodhound").is_some_and(|i| i.sandbox_required()),
            "bloodhound must be a registered, sandbox-required installer"
        );

        let sandbox_only = CatalogEntry {
            binary_name: "bloodhound-python".into(),
            display_name: "BloodHound".into(),
            description: "AD collection".into(),
            category: ToolCategory::ActiveDirectory,
            install_method: InstallMethod::Custom {
                id: "bloodhound".into(),
            },
            recommended: true,
            used_by: vec![],
            state: InstallState::Missing,
        };
        assert!(
            !sandbox_only.is_auto_installable(),
            "a sandbox-required custom installer must NOT be auto-installable when the sandbox is off"
        );

        let native = CatalogEntry {
            binary_name: "webwright".into(),
            display_name: "Webwright".into(),
            description: "browser automation".into(),
            category: ToolCategory::Web,
            install_method: InstallMethod::Custom {
                id: "webwright".into(),
            },
            recommended: true,
            used_by: vec![],
            state: InstallState::Missing,
        };
        assert!(
            native.is_auto_installable(),
            "a natively-installing custom tool (webwright) must stay auto-installable when the sandbox is off"
        );
    }
```

### 2. New file `crates/platform/tests/disable_sandbox_precedence.rs`

Its own integration binary on purpose: `init_sandbox_from_env` reads `DISABLE_SANDBOX` once via `std::sync::Once`, so the env var must be set before anything else in the process touches the sandbox API. It is pure in-memory (no rootfs, no backend, no network), so unlike the backend harnesses in that directory it is NOT `#[ignore]`d and runs in CI. Note this guards the observable precedence contract (`is_sandbox_enabled()` stays false); it reddens under the full pre-fix, not if only one of the two short-circuits is removed.

```rust
//! Regression guard for #244 Problem 1: `DISABLE_SANDBOX` must take precedence
//! over the UI shell-mode toggle. Once the environment has disabled the
//! sandbox, `set_use_sandbox(true)` (driven by the desktop/web shell-mode
//! setting) is a no-op and `is_sandbox_enabled()` stays `false`.
//!
//! This is pure in-memory logic (an env var plus process-global atomics): it
//! provisions no rootfs, spawns no backend, and touches no network. Unlike the
//! backend harnesses in this directory it is therefore NOT `#[ignore]`d and
//! runs in the required CI job (`cargo test --workspace ... --features
//! pentest-platform/desktop-pcap`, which also sets `DISABLE_SANDBOX=true`).
//!
//! It lives in its own integration binary on purpose: `init_sandbox_from_env`
//! reads `DISABLE_SANDBOX` exactly once via `std::sync::Once`, so the env var
//! must be set before anything else in the process touches the sandbox API.
//! A dedicated single-test binary guarantees that ordering.
//!
//! Desktop-gated: the real `set_use_sandbox` / `is_sandbox_enabled` live in
//! `pentest_platform::desktop`; on non-desktop feature builds the crate root
//! exposes no-op stubs, so there is no precedence to guard there.
#![cfg(feature = "desktop")]

#[test]
fn env_disable_wins_over_ui_shell_mode_toggle() {
    // Set before the first sandbox call so the `init_sandbox_from_env` `Once`
    // latches the env-disabled state. Safe on edition 2021.
    std::env::set_var("DISABLE_SANDBOX", "true");

    // The env has disabled the sandbox: it reports off.
    assert!(
        !pentest_platform::is_sandbox_enabled(),
        "DISABLE_SANDBOX=true must report the sandbox disabled"
    );

    // The UI shell-mode toggle tries to switch it back on. The env wins, so
    // this is a no-op and the sandbox stays off.
    pentest_platform::set_use_sandbox(true);
    assert!(
        !pentest_platform::is_sandbox_enabled(),
        "set_use_sandbox(true) must not override DISABLE_SANDBOX=true"
    );

    // An explicit shell-mode "off" is consistent and also leaves it off.
    pentest_platform::set_use_sandbox(false);
    assert!(
        !pentest_platform::is_sandbox_enabled(),
        "sandbox must remain disabled after set_use_sandbox(false)"
    );
}
```

## Minor notes (non-blocking)

- **AC #1 partial:** the Tools chip shows enabled/disabled but not which backend (bwrap/proot/none) as the issue asks. A proot/native indicator already exists elsewhere in the settings UI, so this is a polish gap. Fine to defer if #244 is updated to reflect it.
- **PR description drift:** the description says the chip is shown "when `shell_mode == Proot && sandbox_available`", but the code keys on `is_sandbox_enabled()` (`settings_page.rs:280`), which is the more correct signal (it reflects the `DISABLE_SANDBOX` override). Worth reconciling the description so it matches the code.
- **`STATE_INSTALLED` scope:** the constant's doc-comment says all call sites share one copy, but `catalog.rs:492` still hardcodes `"installed"` independently. True within `settings_page.rs`, not across the crate.
- **Android Build CI red is pre-existing infra**, not this PR: the failing step is `Fetch Android proot runtime` (curl exit 22, an external download), also red on the sibling `fix/pick/stale-fallback-err-pr-243-v1` branch. Every code gate (Clippy, Format, Check, Test, CodeQL, Gitleaks) passes.

edbonddevo (approved):

PR #422, opened 2026-09-01
fix(seed): create resource base dir before validating wordlist downloads

josh-devo:
## Problem

On a fresh install, **Settings → Seed Resources → Download wordlists** fails for every resource. Reproduced on a native Kali box, but the bug is host/distro independent.

Root cause is in `crates/core/src/seed.rs`:

- `SeedManager::new()` sets `base_dir = ~/.pick/resources` but never creates it.
- `seed_resource_with_options` called `self.base_dir.canonicalize()` for its path-traversal check **before** anything created that directory.
- `canonicalize()` requires the path to exist, so on a fresh system it returns `ENOENT` and every resource aborts with `Failed to validate base directory: No such file or directory` — before a single byte is downloaded.

This is **not** related to bwrap / the sandbox / the Docker image. Seeding is a pure in-process `reqwest` HTTP download to the host filesystem; it never routes through the command sandbox. The fix belongs in `pick`, not in `demo-workloads`.

## Fix

Extract `resolve_destination()`, which:

1. `create_dir_all(base_dir)` first, so `canonicalize()` succeeds on a fresh install.
2. Keeps the `..`-traversal and containment checks, but resolves the destination against the **canonical** base so a symlinked `HOME` does not cause a spurious "outside base directory" rejection now that the directory is created.

## Testing

- Offline reproduction confirmed the pre-fix failure (every resource → `Failed to validate base directory`, base dir never created).
- New offline regression tests: `resolve_destination_creates_missing_base_dir`, `resolve_destination_rejects_paths_outside_base`.
- Ignored end-to-end network test `e2e_seed_real_wordlist_on_fresh_install` downloads a real SecLists wordlist into a fresh base dir (run with `--ignored`).
- `cargo fmt`, `cargo clippy -p pentest-core --all-targets -- -D warnings`, and `cargo test -p pentest-core --lib seed::` all green.

### DIFF
--- a/crates/core/src/seed.rs
+++ b/crates/core/src/seed.rs
@@ -456,6 +456,59 @@ impl SeedManager {
             .await
     }
 
+    /// Ensure the seed base directory exists and resolve `dest` into a path that
+    /// is guaranteed to live under it. Returns an error if the destination would
+    /// escape the base directory (path-traversal / SSRF-write protection).
+    ///
+    /// Creating the base directory here — rather than assuming it already exists —
+    /// is what lets seeding work on a fresh install. `~/.pick/resources` is not
+    /// created anywhere else, so without this `canonicalize()` below failed with
+    /// ENOENT and every wordlist/resource download aborted at validation before a
+    /// single byte was fetched.
+    async fn resolve_destination(&self, dest: &Path) -> Result<PathBuf> {
+        // Create the base directory first so canonicalize() can resolve it.
+        fs::create_dir_all(&self.base_dir)
+            .await
+            .map_err(|e| Error::ToolExecution(format!("Failed to create base directory: {}", e)))?;
+
+        // Canonicalize resolves symlinks (e.g. a symlinked HOME) so the
+        // containment check below compares like-for-like.
+        let canonical_base = self.base_dir.canonicalize().map_err(|e| {
+            Error::ToolExecution(format!("Failed to validate base directory: {}", e))
+        })?;
+
+        // Reject any ".." components up front to prevent traversal.
+        if dest
+            .components()
+            .any(|c| matches!(c, std::path::Component::ParentDir))
+        {
+            return Err(Error::InvalidParams(format!(
+                "Security: Path traversal detected in destination: {}",
+                dest.display()
+            )));
+        }
+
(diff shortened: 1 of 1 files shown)


jtomek-strike48 (approved):
Approve. Verified empirically in an isolated worktree at the PR head (045e446), not by reading.

## Verified sound

- Fix works and the regression is real. Mutation test: deleting `create_dir_all(&self.base_dir)` in `resolve_destination` makes `resolve_destination_creates_missing_base_dir` go RED with the exact production error (`Failed to validate base directory: No such file or directory (os error 2)`). Restored -> green. Genuine guard that reproduces the described bug.
- Containment guard is live. Neutering the `!resolved.starts_with(&canonical_base)` check makes `resolve_destination_rejects_paths_outside_base` go RED ("must reject a destination outside the base dir"). The `..`-traversal and out-of-base rejections both hold.
- CI actually runs these. Both new tests are plain `#[tokio::test]` in the `pentest-core` lib (no feature gate, no `#[ignore]`), so the required Linux `Test` lane (`cargo test --workspace`) executes them. The network `e2e_seed_real_wordlist_on_fresh_install` is correctly `#[ignore]`d.
- Claim audit passes. `SeedManager::new()` sets `base_dir` and never creates it; seeding is pure in-process reqwest (sandbox-independent). The symlinked-HOME rebasing onto `canonical_base` is necessary, not cosmetic: the old lexical `starts_with` would spuriously reject a symlinked HOME.
- Blast radius clean. Single write chokepoint (`File::create`), guarded by `resolve_destination`; all consumers are the Settings UI via `seed_tier*` with absolute in-base destinations.

## Low / advisory (non-blocking)

`resolve_destination` computes and returns the validated `resolved` path, but `seed_resource_with_options` discards it and writes to the raw `resource.destination`. Equivalent for every shipped input (all destinations are absolute and inside base). Latent only: a relative destination would pass the containment check (rebased onto the canonical base) yet `File::create(&resource.destination)` would write it relative to the process CWD. Not reachable today (no consumer passes a relative/custom destination). Cheap hardening: bind `let dest = self.resolve_destination(...).await?;` and use `dest` for the parent-dir creation and `File::create`, so the write always lands on the path that was validated.

PR #433, opened 2026-09-03
fix(demo): pass STRIKE48_API_URL through to the pick service (OTT callback pin)

cchao-devo:
## Problem

Since #394, the vendored SDK resolves the post-approval OTT callback in the order `STRIKE48_API_URL` env var > derived from dialed host > server-advertised. The DVWA demo unit dials an insecure `grpc://…:80` front, so the derived-from-dialed-host step yields an `http://` callback base. On platforms whose edge terminates plain HTTP with a 301 (envoy http-to-https `RequestRedirect`, for example), the `register-with-ott` POST gets the empty redirect body and the exchange dies with:

    Failed to complete OTT registration: Serialization error: Failed to parse response: error decoding response body

The connector log names the override explicitly:

    OTT callback base overridden: server advertised "https://studio.strike48.test", using "http://studio.strike48.test"
    STRIKE48_API_URL is not configured; … set STRIKE48_API_URL to pin the origin explicitly.

That is exactly what the SDK's own log message recommends — but the demo compose's pick service declares a fixed `environment:` allowlist, so **`STRIKE48_API_URL` can never reach the container even when set**. There is no supported way for a deployer of the compose unit to make the pin the SDK recommends.

## Fix

Pass the variable through. The `:-` default is the empty string: with nothing set, SDK resolution behavior is **byte-identical to today**. Nothing changes for anyone whose platform front is https/`wss://` (the derivation is already correct there); only deployments fronted by an http→https-redirecting edge (like a local newdev envoy) opt in by setting the pin.

## Evidence

- Observed on a four-connector fleet (studio built from 2026-09-02 develop): two tenants failed the OTT exchange with the exact log lines above; after adding this pass-through and setting the pin in the env file, all four completed — `Sent JWT re-registration on existing stream`, zero `ERROR` lines, engagements ran tools end-to-end.
- The same pin is already the required pattern in other pick deployment paths: our EC2 demo stacks declare `STRIKE48_API_URL` **required** (`:?`) in their generated compose, and the k8s connector chart sets it on the deployment.

## Test plan

- No unit test: pure compose allowlist pass-through, zero Rust code changes.
- Manual proof (platform with http→https edge redirect):
  1. `just dvwa-up` with the demo env (insecure `grpc://` host) **without** the pin → approve in Gateways → observe `OTT callback base overridden` + serialization error; connector never becomes authenticated.
  2. Set `STRIKE48_API_URL=<tenant API base>` in the env file → recreate → approve → `Sent JWT re-registration on existing stream`.

## Reference

  Found during testing of Strike48/init-dev#423

### DIFF
--- a/docker-compose.dvwa.yml
+++ b/docker-compose.dvwa.yml
@@ -107,6 +107,14 @@ services:
       # the container + the internal scan-net, not the in-container sandbox. (The
       # full sandbox lives in the prod Kali image built from Dockerfile.scratch.)
       DISABLE_SANDBOX: "${DISABLE_SANDBOX:-true}"
+      # Same-origin base for the OTT post-approval callback. The vendored SDK
+      # resolves the register URL as: STRIKE48_API_URL > derived-from-dialed-
+      # host > server-advertised. When the connector dials an insecure
+      # grpc:// fronting that 301-redirects http→https (envoy http-to-https),
+      # the dialed-host derivation yields http:// and the token exchange dies
+      # on the redirect's empty body. Pin the tenant API base to override.
+      # Unset = SDK's built-in resolution (behavior unchanged).
+      STRIKE48_API_URL: "${STRIKE48_API_URL:-}"
       # STRIKE48_TOKEN intentionally unset → Pick registers PENDING approval
       # (OTT flow). Operator approves in Studio → Gateways. The minted JWT is
       # persisted to HOME below so the approval survives restarts.


jtomek-strike48:
## Review: merge

Confidence: high. This is an 8-line, one-file compose passthrough. I read the vendored SDK resolution code and traced the empty-string default through all seven `STRIKE48_API_URL` consumers, and confirmed the compose-allowlist premise directly against the file. One advisory-only PR-body inaccuracy; nothing blocking.

### Verified sound

**Core premise is true - the `environment:` block is the sole allowlist.** `docker-compose.dvwa.yml` has no `env_file` directive, so an unlisted `STRIKE48_API_URL` set on the host genuinely cannot reach the container. Right mechanism, and the new line matches the file's existing `${VAR:-default}` idiom.

**"Byte-identical to today when unset" - verified, with one non-observable caveat.** `${STRIKE48_API_URL:-}` sets the container var to the empty string, not unset, and every consumer reads it with raw `std::env::var(...).ok()` -> `Some("")`, not `None`. It does not break here because every resolution/validation path filters empty:
- `resolve_register_base` - `clean` closure drops empty after trim (`ott_provider.rs:237-242`)
- `resolve_register_base_verbose` - delegates to the above + `configured.is_none_or(|c| c.trim().is_empty())` (`:315`)
- `validate_register_origin` - `Some(c) if !c.trim().is_empty()` else skip (`:465-466`)
- the three OTT callers (`connector.rs:2279`, `registration_runner.rs:937`, `ws_multiplex.rs:1460`) all feed the empty-filtered verbose resolver.

The one raw consumer that does not filter is `http_target` at `ott_provider.rs:127-129` (`api_url.unwrap_or_else(|| "https://studio")`): `Some("")` yields `""` vs `"https://studio"` when unset. But that value only feeds `build_http_client` -> `resolve_proxy` for NO_PROXY host matching. With no proxy configured (`STUDIO_PROXY`/`HTTP_PROXY` unset - the dvwa demo case), `resolve_proxy` returns `None` and the client is built identically. Not observable in the demo; not a defect.

**Security posture neutral-to-positive.** `STRIKE48_API_URL` is an origin allowlist pin (`enforce_same_origin`); setting it tightens the OTT callback, and the empty default neither weakens nor enables anything. No secret introduced.

**Scope clean.** Confined to `docker-compose.dvwa.yml`; no reach into the headless connector, desktop app, or prod (prod sandbox is `Dockerfile.scratch`). CI fully green incl. Compose Lint; "zero Rust code changes" is accurate, so "no unit test" is the correct call.

### Advisory (non-blocking)

- **Claim audit:** the body says "the k8s connector chart sets it on the deployment." The chart actually sets `MATRIX_API_URL`, not `STRIKE48_API_URL` (`chart/pick-connector/templates/deployment.yaml:71`) - a distinct variable. (That `MATRIX_API_URL` is already required in this same compose file yet the bug still reproduces is itself proof the SDK's OTT callback reads `STRIKE48_API_URL` specifically.) Does not affect the merged artifact; worth a one-word correction if you touch the description.
- **Unable to verify:** the "EC2 demo stacks declare it required (`:?`)" claim is in externally-generated compose not in this repo. Low stakes.

jtomek-strike48 (approved):
Approve. Detailed empirical review posted above (verdict: merge). Safe, correctly-scoped compose passthrough; empty-string default is byte-identical to unset across all SDK consumers. Only advisory nit: the k8s chart sets MATRIX_API_URL, not STRIKE48_API_URL - optional PR-body wording fix.

PR #440, opened 2026-09-14
test(core): fix flaky secure_store backend tests (global-state race)

### LINKED ISSUE #439: Flaky: secure_store backend tests race on shared global BACKEND under parallel test load
## Summary

Two unit tests in `crates/core/src/secure_store.rs` share the process-global `static BACKEND` and mutate it, so they race when the test binary runs them in parallel (cargo's default). This produces intermittent failures of the required `Test` CI job that are unrelated to the change under test.

## Affected tests

- `secure_store::tests::registered_backend_round_trips` (asserts `is_available()`)
- `secure_store::tests::unavailable_without_backend_by_default` (asserts `!is_available()`)

Both read/write the same global:

```
crates/core/src/secure_store.rs:25
static BACKEND: std::sync::Mutex<Option<Backend>> = std::sync::Mutex::new(None);
```

`registered_backend_round_trips` calls `set_backend(...)` then asserts `is_available()`, while `unavailable_without_backend_by_default` clears the global to `None` and asserts `!is_available()`. Running concurrently, either can observe the other's write, so either assertion can fail.

## Evidence (reproduced empirically)

Observed in CI on an unrelated PR (#438), required `Test` job:

```
test secure_store::tests::registered_backend_round_trips ... FAILED
thread '...registered_backend_round_trips' panicked at crates/core/src/secure_store.rs:119:9:
assertion failed: is_available()
```

Reproduced locally on the base SHA `5fb18c8` (i.e. without #438), 3 consecutive runs of `cargo +stable test -p pentest-core --lib secure_store::tests`:

- run 1: FAILED (`unavailable_without_backend_by_default`, `assertion failed: !is_available()`)
- run 2: ok
- run 3: FAILED (`unavailable_without_backend_by_default`)

The failing test differs run to run, which is the signature of a shared-state race rather than a logic bug. It is pre-existing and independent of any single PR.

## Impact

Intermittent red on the required `Test` job (`cargo test --workspace --locked --features pentest-platform/desktop-pcap`), causing false failures on unrelated PRs and wasted reruns.

## Suggested fix

Serialize these tests, or remove the shared-global dependency. Options:

- Gate the backend-touching tests with a test-local mutex (or the `serial_test` crate's `#[serial]`), so they never run concurrently with each other.
- Have each test save and restore the previous `BACKEND` value, and hold the backend lock across the set-and-assert.
- Collapse the two into a single sequential test that steps through the states.

## Related

- #269 (closed) - same class: shared-global flakiness under full-workspace parallelism (evidence buffer).
- #318 (open) - another parallel-load test flake.

jtomek-strike48:
## Summary

- `secure_store::tests::registered_backend_round_trips` and `unavailable_without_backend_by_default` both mutate the process-global `BACKEND`. Rust's default multi-threaded test runner let them race, so one test could set/clear the backend between the other's mutation and its `is_available()` assertion - failing intermittently in the required `Test` CI job.
- Serialize the two on a module-local `TEST_GUARD` mutex (recovering from poisoning so one panic does not cascade). This mirrors the existing guard in `clipboard.rs`, which fixed the same class of flake in this crate.
- No production code changed - test-only.

Fixes #439

## Test plan

- [x] `cargo +stable test -p pentest-core --lib secure_store::tests` run 30 consecutive times: all green (previously ~2 of 3 runs failed, on either test).
- [x] `cargo +stable clippy -p pentest-core --all-targets -- -D warnings` clean
- [x] `cargo +stable fmt -p pentest-core` clean (no changes)
- [x] Reproduced the pre-fix flake on base SHA to confirm root cause (shared global, not logic).


cchao-devo (approved):
===BEGIN REVIEW — Strike48-public/pick#440===                                                                                                                                    
   Strike48-public/pick#440 — test(core): fix flaky secure_store backend tests (global-state race)                                                                                  
   Author: @jtomek-strike48 · main ← fix/pick/secure-store-test-race · 1 file, +10/-0                                                                                               
                                                                                                                                                                                    
   Verdict: merge (rec approve, conf 1.0, HR=no). CI: all 25 green.                                                                                                                 
                                                                                                                                                                                    
   Composition: 100% feat (1 file, 10 lines) — test-only.                                                                                                                           
                                                                                                                                                                                    
   Foundation (8 axes) — 5 pass, 3 n/a. Fixes #439 and nothing else; the guard lives in the module that owns the process-global and mirrors the crate's existing clipboard.rs guard 
 pattern (static Mutex<()> + poisoning recovery via into_inner()).                                                                                                                  
   Architecture (A1–A6) — 3 pass, 3 n/a (no schema, migration, or view files). No manifest, workspace, or boundary file changed; Cargo.lock untouched by this diff.                 
   Security (S1–S6) — 1 pass, 5 n/a. secure_store is secrets-adjacent but the change is test-only: no production code, no storage behaviour, no boundary touched.                   
   Adversarial (V1–V8) — 1 pass, 7 n/a. V4 pass — the diff IS the race fix, and the closing is sound: each test takes TEST_GUARD exactly once at its top (no nesting, so no         
 self-deadlock), poisoning is recovered so one panic cannot cascade, and a module sweep confirms both BACKEND mutators are the two now-guarded tests.                               
                                                                                                                                                                                    
   Findings: None.                                                                                                                                                                  
                                                                                                                                                                                    
   Verification (executed, not assumed): on the base SHA (5fb18c84) the suite fails 10/10 runs — panic at crates/core/src/secure_store.rs:98/99, the !is_available() assert in      
 unavailable_without_backend_by_default, even more reliably than the body's "~2 of 3 runs"; on this PR's head the same suite passes 30/30 consecutive runs using the body's own     
 protocol. A module sweep confirms no BACKEND user remains unguarded, and the pattern exactly matches the crate's clipboard.rs precedent.                                           
                                                                                                                                                                                    
   Summary: Serializes the two secure_store backend tests on a module-local mutex because they mutate the process-global BACKEND under Rust's multi-threaded test runner. Verified  
 end-to-end rather than by argument: base fails 10/10 at exactly the assert the issue describes, fixed head passes 30/30. Test-only, mirrors the crate's own precedent, no          
 production code touched, no BACKEND user left unguarded. Nothing to change.                                                                                                        
   ===END REVIEW — Strike48-public/pick#440===

Issue #386, opened 2026-08-18
connector_name used as OTT connector_type breaks StrikeHub pre-approved registration


jtomek-strike48:
## Summary

Pick passes `self.config.connector_name` as the OTT **`connector_type`** when resolving and redeeming a pre-approved OTT. These are semantically different: `connector_name` is the connector's display/persona identity, while `connector_type` must be the SDK connector type (`pentest-connector`). When they differ, OTT registration fails with a misleading error.

## Where

- `crates/ui/src/liveview_connector/mod.rs:916` calls `resolve_ott_tenant(&self.config.connector_name, &self.config.instance_id)`, passing `connector_name` into the `connector_type` parameter (fn signature at `mod.rs:697`).
- That value flows into `OttProvider::new(Some(connector_type...))` and `register_with_ott(connector_type, ...)` (`mod.rs:735`, `mod.rs:761`), and the SDK POSTs it as `connector_type` to `/api/connectors/register-with-ott`.

## Impact

A host (e.g. StrikeHub) mints a pre-approved OTT for `connector_type = "pentest-connector"`, then launches Pick without setting `CONNECTOR_NAME`. Pick falls back to a persisted `connector_name` in `~/.config/pentest-connector/settings.json`. If that value is anything other than `pentest-connector` (for example a custom persona set to avoid the shared-agent collision across multiple connectors), Pick redeems the otherwise-valid OTT with the wrong `connector_type`. The server rejects it (type mismatch / not found) and the SDK surfaces the generic `Invalid configuration: Invalid or expired OTT`. The connector then exits and never becomes ready (host marks it Offline).

This is easily misdiagnosed as an expired/invalid token or a server problem, because the error text points at the token rather than the type.

## Repro

1. Set `connector_name` to a custom value (not `pentest-connector`) in `settings.json` (this is exactly what the agent-persona-uniqueness workaround does).
2. Launch Pick via a host that pre-approves an OTT for `pentest-connector` and passes it via `STRIKE48_REGISTRATION_TOKEN` (StrikeHub does this).
3. Pick logs `Starting <connector_name> connector`, then `Invalid or expired OTT`, and exits.
4. Set `connector_name = "pentest-connector"` and relaunch: registers successfully with the identical token flow.

## Proposed fix

[...]

PR #454, opened 2026-09-17
fix(deps): bump rustls to 0.23.45 to clear RUSTSEC-2026-0285

jtomek-strike48:
## Summary

- Bumps `rustls` from 0.23.37 to 0.23.45 in `Cargo.lock` to clear RUSTSEC-2026-0285 (TLS 1.3 handshake messages accepted across encryption-level boundaries, severity 5.3 medium).
- Advisory was published 2026-09-14, after main's last green audit. The Dependency Audit job began failing on every open branch once it landed in the advisory database, not from any code change. 0.23.45 is the first patched release and is semver-compatible, so this is a lock-only update.
- cargo pulls the matching `aws-lc-rs`, `aws-lc-sys`, and `rustls-webpki` point releases along with it. No source or manifest changes.

## Test plan

- [x] `cargo update -p rustls --precise 0.23.45` produces a lock-only diff (4 version bumps)
- [x] `cargo audit` exits 0 after the bump (was exit 1, "1 vulnerability found")
- [ ] CI Dependency Audit job passes on this branch
- [ ] CI build/test lanes pass with the updated lock

PR #455, opened 2026-09-18
fix: send SDK connector_type in OTT registration and fix persona connector bindings

jtomek-strike48:
## Summary

- `BaseConnector::connector_type()` now returns the SDK wire type (`CONNECTOR_TYPE`) instead of the persona `connector_name`, so OTT registration, credential loading, and headless runs identify the connector correctly
- Agent default connector keys now derive from the wire type (`{tenant_id}.{CONNECTOR_TYPE}.*`) at all three key sites, replacing the stale `{tenant}.{connector_name}.*` construction; `connector_name` is retained only for display/report identity
- Adds regression guards for persona-vs-wire-type at the config, `ToolConnector`, and `PickConnector` layers

Refs #386

## Test plan

- [x] `cargo fmt --check` clean
- [x] `cargo check --workspace --features pentest-platform/desktop-pcap` clean
- [x] `cargo clippy --workspace --features pentest-platform/desktop-pcap -- -D warnings` clean
- [x] `cargo test --workspace --features pentest-platform/desktop-pcap` — all 35 binaries pass, 0 failures (pre-existing `ignored` entries only)
- [x] New regression tests pass: `connector_type_is_sdk_type_not_persona_name`, config persona-leak guard, `pick_connector` persona guard, `default_pentest_agent_input` wire-typing assertion


cchao-devo (changes requested):
===BEGIN REVIEW — Strike48-public/pick#455===                                                                                                                                    
   Strike48-public/pick#455 — fix: send SDK connector_type in OTT registration and fix persona connector bindings                                                                   
   Author: @jtomek-strike48 · main ← fix/pick/ott-connector-type · 9 files, +226/-41                                                                                                
                                                                                                                                                                                    
   Verdict: request-changes (rec nit, conf 0.78, HR=yes). CI: 1 failing / 25 total — Dependency Audit, attributed below; not this PR's diff.                                        
                                                                                                                                                                                    
   Composition: 95% bug fix (8 files) · 5% UI (1 file).                                                                                                                             
                                                                                                                                                                                    
   Foundation (8 axes) — 7 pass, 1 n/a. Refs #386 and the linked issue is an exact match; the SDK wire type becomes one constant in config.rs and every consumer routes through it; 
 four new regression guards each use a fixture whose persona deliberately differs from CONNECTOR_TYPE so they cannot pass vacuously. The test-only PentestConnector keeping its     
 "dioxus-pentest" sentinel as a documented tripwire is a nice touch.                                                                                                                
   Architecture (A1–A6) — 3 pass, 3 n/a (no schema, migration, or view files). The SDK wire contract (OTT pre-approve, register-with-ott, saved-credential file algebra) is stated  
 once in the constant's doc and honoured at every site.                                                                                                                             
   Security (S1–S6) — 4 pass, 2 n/a. S1 pass: the OTT redemption path now sends the server-minted type, so registration cannot be steered onto the wrong identity by the persona    
 env var. S3 pass: no migration gap exists — under the old code a custom persona made the helpers look for a file the SDK never wrote (effective no-ops), and for the default       
 persona the file name is byte-identical, so silent reconnect keeps working. S4 pass: an operator-set persona string no longer flows into the registration request the server       
 matches against minted tokens.                                                                                                                                                     
   Adversarial (V1–V8) — 2 pass, 6 n/a. V1 pass: the persona string can no longer select or shift a registration identity; the wire type is a compile-time constant and instance_id 
 remains the per-instance discriminator. V2 pass: OTT redemption now presents exactly the type the host minted the token for.                                                       
                                                                                                                                                                                    
   Findings:                                                                                                                                                                        
   - nit: the diff's own file still carries the doc line this PR invalidates — agent_defaults.rs:106-108 reads "connector_name controls the gateway identity. Instances sharing the 
 same name are round-robin'd; use a unique name (e.g. pentest-connector-<hostname>)". After this PR, connector_name controls nothing on the wire — the registration identity is the 
 constant CONNECTOR_TYPE plus instance_id — so that paragraph now teaches the exact misconception #386 fixed.                                                                       
     Fix: rewrite that doc paragraph to the new model (persona = display/agent name; identity = type + instance id).                                                                
   - nit: an operational behaviour change for multi-instance deployments is not called out in the body. The old docs told operators to set a unique CONNECTOR_NAME (e.g.            
 pentest-connector-<hostname>) for a dedicated agent view; with the agent connector key now {tenant}.{CONNECTOR_TYPE}.*, EVERY pick instance in a tenant matches the same agent key 
 and joins one round-robin pool, and a unique name changes only the display name. That is the correct reading of #386, but hosts relying on unique names for per-instance agent     
 isolation will silently change behaviour.                                                                                                                                          
     Fix: note the multi-instance semantics in the body/commit and confirm with the host side that per-type round-robining is the intended model.                                   
   - nit: the Dependency Audit job is red on this head, and it is NOT this PR's diff: the run scanned rustls 0.23.37 and failed on exactly RUSTSEC-2026-0285, the advisory #454     
 already fixed on main (4ee23a55f). This branch was cut before that merge.                                                                                                          
     Fix: rebase onto current main so the audit goes green (the branch is fast-forwardable — main is 0 commits ahead of the base).                                                  
   - unable-to-verify: the new pentest-ui regression guard could not be executed locally: on macOS, cargo test -p pentest-ui --lib --features pentest-platform/desktop-pcap (CI's   
 exact command) fails to compile pentest-ui with 5 cfg-gated errors (shell_ws, liveview_server::{get_workspace_path,set_matrix_credentials} "configured out"). I verified this is   
 pre-existing — origin/main fails identically at the same command — so it is not introduced here, and CI's Linux test lane (green at this head) is what exercises that test.        
     Fix: none required for this PR; a separate fix for the macOS test build (gate the callers consistently with the modules, or enable the missing features) would let reviewers   
 on Apple Silicon run the UI tests instead of deferring to CI.                                                                                                                      
                                                                                                                                                                                    
   Summary: Separates the SDK wire identity from the persona identity everywhere it matters — connector_type sent to OTT pre-approve/register-with-ott and the WebSocket            
 registration is now the fixed CONNECTOR_TYPE constant, the saved-credential file algebra matches what the SDK's runner actually writes, and all three agent connector keys bind to 
 the type the connector registered as — with the persona retained for display, agent naming, and logs. The migration risk does not exist: the old persona-scoped lookups were       
 no-ops for custom personas and byte-identical for the default one. Four regression guards with honest fixtures; three run green locally, the fourth via CI's Linux lane.           
 Remaining: one stale doc line inside the diff, an unannounced multi-instance behaviour change, the pre-existing rustls audit red, and a pre-existing macOS test-build gap.         
   ===END REVIEW — Strike48-public/pick#455===

PR #459, opened 2026-09-18
fix(ci): checksum-pin Termux fetch-proot debs against upstream drift

jtomek-strike48:
## Summary

- Adds sha256 pins for all 8 proot/talloc/busybox/libandroid-shmem debs, verified on every download
- Documents the Termux rolling-pool 404 mode, the version-bump procedure, and the durable #252 mirror-to-release-asset fix
- Pinned URLs and checksums were verified live against packages.termux.dev

## Test plan

- [x] `fetch-proot --dry-run` parses
- [x] shellcheck clean
- [x] tamper test on a downloaded deb fails loudly
- [x] all 8 URLs return 200

Refs #397


cchao-devo (changes requested):
===BEGIN REVIEW — Strike48-public/pick#459===                                                                                                                                    
   Strike48-public/pick#459 — fix(ci): checksum-pin Termux fetch-proot debs against upstream drift                                                                                  
   Author: @jtomek-strike48 · main ← fix/pick/pin-proot-urls · 1 file, +58/-0                                                                                                       
                                                                                                                                                                                    
   Verdict: request-changes (rec nit, conf 0.80, HR=yes). CI: all 25 green — but the Android Build lane does not run on PRs, so the four test-plan items (all left unchecked in the 
 body) were executed for this review and pass.                                                                                                                                      
                                                                                                                                                                                    
   Composition: 100% bug fix (1 file, 58 lines).                                                                                                                                    
                                                                                                                                                                                    
   Foundation (8 axes) — 5 pass, 3 n/a. Refs #397 and the CAUTION block documents the failure mode, the version-bump procedure, and honestly scopes the fix: "the sha256 pins below 
 catch silent in-place replacement of a pinned .deb; they do NOT fix 404s. The durable fix is mirroring ... (#252)". The download filenames were renamed to the full                
 pkg_version_arch form so the sha256sum -c manifest matches unambiguously (the old talloc_${arch}.deb short name would have been a suffix-collision risk against libtalloc_...).    
   Architecture (A1–A6) — 3 pass, 3 n/a (no schema, migration, or view files). Manifest, verify helper, and version pins all live in the recipe that downloads the artifacts.       
   Security (S1–S6) — 1 pass, 5 n/a. The change hardens the build's dependency ingestion: a replaced or superseded pinned .deb now fails the recipe loudly instead of silently      
 shipping different binaries inside the APK.                                                                                                                                        
   Adversarial (V1–V8) — 2 pass, 6 n/a. V1 pass: this is the change's purpose — a registry-side replacement of a pinned .deb can no longer silently ship different binaries into    
 the APK; the tamper test was executed and the failure is loud, attributed, and actionable. V7 pass: verified against the LIVE upstream — all 8 pinned URLs returned 200 and every  
 sha256 matched the downloaded bytes today; the manifest as materialized by just (extracted from --dry-run, trailing newline intact) is well-formed and sha256sum -c passes 8/8.    
                                                                                                                                                                                    
   Findings:                                                                                                                                                                        
   - nit: the body's test-plan checkboxes are all UNCHECKED while the Summary asserts the pins were "verified live against packages.termux.dev" — both are now true (this review    
 re-executed all four items), so the boxes can simply be ticked. Ticking them matters here because the Android Build lane does not run on PRs, so the checkboxes are the only       
 recorded signal that the pins were validated against the live pool.                                                                                                                
     Fix: tick the four boxes.                                                                                                                                                      
   - nit: "shellcheck clean" is true only for the ADDED lines. Shellchecking the evaluated recipe (just --dry-run output) reports 2 pre-existing warnings on lines this PR did not  
 touch: SC2064 at line 4 — trap "rm -rf $TMP" EXIT expands $TMP at set-time rather than signal-time (harmless today because TMP is assigned once immediately before and never       
 reassigned, but it is the exact bug class SC2064 exists for), and SC2010 at line 79 — ls ... | grep -v '\.so$' | head -1 for the busybox stub lookup. Both are two-line fixes      
 inside a recipe this PR already edits.                                                                                                                                             
     Fix: trap 'rm -rf "$TMP"' EXIT and replace the ls | grep | head with a glob, then tick the shellcheck box.                                                                     
   - Observation (not a finding): the no-match path of verify_sha relies on set -o pipefail — on this platform sha256sum -c with empty input exits 0, so if pipefail were ever      
 dropped, an unlisted filename would silently pass. Unreachable in practice (download names are constructed from the same pins as the manifest) and the recipe has pipefail, but    
 worth knowing if the recipe is ever refactored.                                                                                                                                    
                                                                                                                                                                                    
   Summary: Checksum-pins all eight Termux .debs (proot, libtalloc, busybox, libandroid-shmem x x86_64/aarch64) and verifies every download with a loud, actionable failure that    
 names the bump procedure — closing the "silently ship different binaries" half of #397. The rolling-pool 404 half is explicitly not fixed and honestly documented as deferred to   
 the #252 mirror. Executed here rather than trusted: all 8 URLs returned 200 and every pin matched the live bytes; the tamper test fails with the intended message; the manifest as 
 materialized by just verifies 8/8; and the recipe parses. The only actionable items are checkbox hygiene and two pre-existing shellcheck warnings in a recipe the PR already       
 edits.                                                                                                                                                                             
   ===END REVIEW — Strike48-public/pick#459===

Issue #309, opened 2026-07-25
SSDP/mDNS discovery returns success when the probe never ran


jtomek-strike48:
## Problem

Network discovery helpers return a successful, empty result when the probe could not be sent at all, making "nothing is on the network" indistinguishable from "we never looked."

In `crates/platform/src/common/ssdp.rs`:

```rust
let socket = match UdpSocket::bind("0.0.0.0:0") {
    Ok(s) => s,
    Err(_) => return Ok(vec![]),        // :25
};
...
if socket.send_to(...).is_err() {
    return Ok(vec![]);                  // :41
}
```

`crates/platform/src/common/mdns.rs` does the same at `:38` and `:49`, and additionally discards three setup calls with `let _ =`. One of those matters on its own: the dropped `set_read_timeout` at `:45` means that if the `setsockopt` fails, the socket stays blocking and the deadline loop at `:57-67` can park in `recv_from` indefinitely. Nothing upstream enforces the caller's `timeout_ms`.

The doc comment states the empty-result behaviour is deliberate ("a blocked sandbox degrades gracefully instead of failing the whole scan"), which is a reasonable instinct. The issue is that graceful degradation and a genuine negative finding currently produce byte-identical output.

## Impact

Under a sandbox that denies UDP, or with no multicast route, discovery returns `Ok([])`. The agent reads zero devices and writes a report stating no UPnP or mDNS devices were found. The operator may hand that to a client as evidence of a clean network when in fact no packet ever left the host.

This is the same failure class as #306 and the reason it is worth fixing at the pattern level rather than per call site: several discovery paths independently chose "succeed with nothing" as their failure mode.

## Reproduction

```bash
rg -n 'return Ok\(vec!\[\]\)' crates/platform/src/common/
rg -n 'let _ = socket' crates/platform/src/common/mdns.rs
```

A socket whose bind or send is denied yields `Ok([])` with `is_ok() == true`, indistinguishable from a clean sweep.

## Suggested fix

Two separable changes.

1. **Make the outcome explicit.** Return a struct that records whether the probe actually ran, rather than a bare `Vec`. Something like:

```rust
pub struct DiscoveryResult {
    pub devices: Vec<SsdpDevice>,
    pub probe_status: ProbeStatus,   // Completed | Skipped(reason) | Failed(reason)
}
```

[...]

PR #460, opened 2026-09-18
fix(tools): report skipped SSDP/mDNS probes as skipped, not as an empty successful sweep

jtomek-strike48:
## Summary

- ProbeOutcome becomes a tagged enum with ran/skipped+reason and flows through new NetworkOps outcome-aware methods
- ssdp_discover and network_discover expose data.probe in tool JSON
- A skipped probe now yields ToolOutcome::Skipped with success=false per the #184 house pattern, so a never-ran probe can no longer masquerade as a clean-network finding

## Test plan

- [x] cargo check/test/clippy clean for pentest-platform and pentest-tools
- [x] Regression tests assert skipped vs ran-empty distinction and JSON shape

Refs #309


cchao-devo (changes requested):
## 🔍 Code Review — Strike48-public/pick#460                                                                                                                                     
   **PR:** fix(tools): report skipped SSDP/mDNS probes as skipped, not as an empty successful sweep · @jtomek-strike48 · 9 files, +315/−0 · Refs #309 · CI: all 25 green            
                                                                                                                                                                                    
   ---                                                                                                                                                                              
                                                                                                                                                                                    
   ## 🚫 Verdict: BLOCK — the fix misses one platform path that still has the bug                                                                                                   
                                                                                                                                                                                    
   | Field | Value |                                                                                                                                                                
   |---|---|                                                                                                                                                                        
   | **Action** | `block` — one or more concerns present |                                                                                                                          
   | **Recommendation** | `concern` |                                                                                                                                               
   | **Confidence** | 0.42 |                                                                                                                                                        
   | **Human review** | required |                                                                                                                                                  
                                                                                                                                                                                    
   **Why:** The desktop/tool-layer fix is clean and well-tested — a tagged `ProbeOutcome` rides in the tool payload, skipped runs downgrade to                                      
 `ToolOutcome::Skipped`/`success=false` per the #184 house pattern, the trait default's contract is verified correct for the impls that use it, and all 8 new tests pass (executed  
 locally; full suites 96+423 green). **But on Android mDNS the bug survives**: the Kotlin bridge swallows NSD start failure into a silent empty result, so the trait default labels 
 a never-ran probe `Ran` and the tool still reports a clean network — the exact #309 failure mode, verified in `ConnectorBridge.kt:87-89`.                                          
                                                                                                                                                                                    
   **Blocker to resolve (A2):**                                                                                                                                                     
   - Propagate the failure through the bridge: in `ConnectorBridge.mdnsDiscover`, record `onStartDiscoveryFailed` (it carries `errorCode`) and emit `{"error": "nsd start failed:   
 <code>"}` (or an explicit skipped marker); then let the existing Rust error path handle it, or add an Android `mdns_discover_with_outcome` override mapping the marker to          
 `ProbeOutcome::Skipped { reason }` (mirroring ios/mod.rs:90-96); add a Kotlin-side test for the failure shape.                                                                     
                                                                                                                                                                                    
   **Cheap hardening (nits worth taking while here):**                                                                                                                              
   1. **Serde-seam test** — the classifier matches the literal `"skipped"` against serde output; a `rename_all` change in probe.rs resurrects #309 with *every test in the diff     
 staying green* (executed counterexample). One test that serializes the real `ProbeOutcome::Skipped{..}` through the classifier pins it.                                            
   2. **Hoist the duplicated classifier** — `classify_probe_outcome`/`probe_status` are byte-identical in ssdp_discover.rs:82-100 and network_discover.rs:97-110, string-matching   
 JSON the tool just serialized; move to a shared home (`crates/tools/src/util.rs` or beside `with_outcome`) and pass the typed `ProbeOutcome` (it's `Clone`).                       
   3. **Trait-default mock test** — pin the documented Err→Err / Ok→Ran mapping; the override contract is currently documentation-only for future implementors.                     
   4. **Body scoping line** — "Closes the SSDP/mDNS scope of #309; remaining house-style instances (ARP reader, other scanners) stay under the issue" (or flip to `Closes` if the   
 team reads it as title-scoped).                                                                                                                                                    
                                                                                                                                                                                    
   ---                                                                                                                                                                              
                                                                                                                                                                                    
   ### Foundation checks — 1 pass · 5 nit · 2 n/a                                                                                                                                   
                                                                                                                                                                                    
   1. **Intent & linked issue** — nit — both in-scope asks of #309 delivered and tested; scoping statement missing (see nit 4 above).                                               
   2. **Architecture drift** — nit — trait default-evolution is the right pattern; the `Ok → Ran` precondition is doc-only for future implementors (see hardening 3).               
   3. **Security boundary** — nit — the #184 property is delivered at the tool layer; the platform→tool seam is untested and drifts silently (see hardening 1).                     
   4. **Reimplementation** — nit — verbatim classifier duplication (see hardening 2).                                                                                               
   5. **Backend changes** — pass — non-breaking default methods; verified compiling (platform 96/96); default correct for Desktop (all failures `Err`) and Android mDNS Rust-side;  
 zero old-variant constructors remain. (Android's Kotlin-side caveat → A2.)                                                                                                         
   6. **UI changes** — n/a. **7. Screen recording** — n/a — payload JSON only.                                                                                                      
   8. **Tests** — nit — all 8 new tests verified green locally; gaps: the seam test (1) and a trait-default mock test (3); platform-impl integration is JNI-bound and not           
 realistically runnable here.                                                                                                                                                       
                                                                                                                                                                                    
   ### Architecture review — 1 nit · 1 concern · 4 n/a                                                                                                                              
                                                                                                                                                                                    
   - **A1 Module placement** — nit — enum/trait homes correct; the tools-crate classifier duplication is the drift (see hardening 2).                                               
   - **A2 Boundary spec drift** — concern — the documented default-override contract is violated across the JNI boundary: Android mDNS NSD start failure → silent `Ok(vec![])` →    
 labeled `Ran` → zero-finding success. Verified in `ConnectorBridge.kt:87-89`. **Fix:** per the blocker.                                                                            
   - **A3/A4/A5/A6** — n/a — no schemas/migrations/views/constants in the diff.                                                                                                     
                                                                                                                                                                                    
   ### Security review — 1 pass · 5 n/a                                                                                                                                             
                                                                                                                                                                                    
   - **S1/S2/S3/S4/S5** — n/a — probe code behavior-identical; reason strings are errno-class only ("Permission denied (os error 13)"), holding the "short, non-PII" contract; no   
 sinks; single-tenant agent.                                                                                                                                                        
   - **S6 Suspicious content** — pass — 9 text hunks, no binaries.                                                                                                                  
                                                                                                                                                                                    
   ### Adversarial review — 3 pass · 5 n/a                                                                                                                                          
                                                                                                                                                                                    
   - **V1/V2/V3** — n/a — no user input, auth surfaces, or sinks.                                                                                                                   
   - **V4 Race / TOCTOU** — pass — probe code behavior-identical; classification is a pure function of the produced payload.                                                        
   - **V5 Data exfiltration** — pass — reason strings errno-class only; the change strictly reduces false reporting.                                                                
   - **V6 DoS / resource** — pass — timeouts/caps unchanged; O(1) classification.                                                                                                   
   - **V7 Supply chain** — n/a — no new deps.                                                                                                                                       
   - **V8 Insider misuse / audit gap** — n/a — the silent-failure risk is scored as A2.                                                                                             
                                                                                                                                                                                    
   ### Summary                                                                                                                                                                      
                                                                                                                                                                                    
   The SSDP/mDNS scope of #309 is delivered cleanly at the desktop/tool layer with correct trait-evolution design and good tests — but the PR's headline claim ("a skipped probe    
 can no longer masquerade as a clean network") does not hold on Android mDNS, where the Kotlin bridge's silent NSD-failure handling keeps the old behavior alive across the JNI     
 boundary. Fix the bridge/override, add the serde-seam test, and land it.

Issue #374, opened 2026-08-13
Saved endpoint/tenant clobbered by stale-signal settings save (regresses #223)


jtomek-strike48:
## Summary

Surfaced during review of #249. The saved connector endpoint (`last_config`), `auto_connect`, and the self-healed canonical `tenant_id` can be silently clobbered, wiping a fresh user's saved endpoint and regressing the #223 tenant-UUID persistence.

## Root cause
Two sites persist by cloning the in-memory settings snapshot and saving it to disk, but never update the `AppSettings` signal itself:
- `crates/ui/src/connector_app.rs:685` (on_connect, "remember"): `let mut s = settings.peek().clone(); s.last_config = ...; s.auto_connect = true; save_settings(&s);`
- `crates/ui/src/connector_app.rs:772` (OTT tenant self-heal): same pattern for `last_config.tenant_id`.

Because the signal is never updated, any later `settings.write()`-path save (e.g. `connector_app.rs:365`, theme/density/telemetry handlers) serializes the stale signal and overwrites the just-saved fields. The self-heal at :772 also reads the stale `settings.peek()`, so it operates on out-of-date data.

## Proposed fix
Update the authoritative signal (`settings.write().last_config = ...` / `auto_connect` / `tenant_id`) at these sites, matching the fix already applied to `easy_mode` and `matrix_api_url`.

Source: #249 review (SHOULD-FIX).

Issue #223, opened 2026-07-13
Improve auth flow for standalone pick


taylor-wood-devo:
User entered the url `https://discoball.strike48.engineering/`
<img width="1007" height="753" alt="Image" src="https://github.com/user-attachments/assets/8f83f753-502b-4ec5-9840-f08cef97e578" />
- we can see the `:443` is appended and the resulting URL is not found

<img width="796" height="64" alt="Image" src="https://github.com/user-attachments/assets/7f348269-0a5b-43a3-9c96-92422787829f" />

If you hit the disconnect button after this happens, you get the `cleared invalid expired/invalid auth token from settings`
<img width="713" height="682" alt="Image" src="https://github.com/user-attachments/assets/8ba78a3f-7486-4abb-bb5e-d57b30ec33e5" />
- resulting in end-user having to find and edit file value or relaunch app and enter corrected value
  - `cat ~/Library/Application\ Support/pentest-connector/settings.json`

And after correct said URL, we go to gateways and any token being accepted becomes a pre-approved token and doesn't properly get accepted and applied:
<img width="1776" height="945" alt="Image" src="https://github.com/user-attachments/assets/1134521b-f17b-4208-b37c-498885d0b15b" />

TODO:
- [ ] incorrect url entry should essentially be denied otherwise the pick app is opened with no ability to run cmds, etc due to failed auth to studio instance
- [ ] make sure end-user can easily edit studio url instance in between sessions of starting the app i.e) target env A, close and reopen app, enter target env B
- [ ] the OTT token needs to properly be accepted

taylor-wood-devo:
one note on this reported issue

> incorrect url entry should essentially be denied otherwise the pick app is opened with no ability to run cmds, etc due to failed auth to studio instance

the pick app does post approval flow, so it is okay if the pick app is running and pending end-user to accept OTT token in studio
- but we should still fix / handle bad entered urls (ie extra trailing `/`) 
- and also still fix OTT token post approval flow so token is properly accepted

Fixed locally, prepping PR

taylor-wood-devo:
Folded the tenant slug ↔ UUID handling into #225 rather than filing a separate follow-up.

[...]

PR #458, opened 2026-09-18
fix(ui): persist easy-mode telemetry opt-out through the settings signal

jtomek-strike48:
## Summary

- Easy-mode telemetry toggle no longer persists via a detached load/save that a later signal-based save could clobber
- It now emits `on_telemetry_change` to the parent, which writes the authoritative `AppSettings` signal, saves, and updates the telemetry core - the same pattern as the expert-mode toggle

## Test plan

- [x] `cargo check` / `cargo test` / `cargo clippy` pass on pentest-ui with the CI feature set under the pinned 1.92.0 toolchain
- [x] Opt-out survives a subsequent unrelated settings save
- [x] Default-on behavior unchanged

Closes #373

PR #462, opened 2026-09-18
fix(ui): persist settings through the authoritative signal in remaining detached-clone sites

jtomek-strike48:
## Summary

- Fixes the last two sites that persisted a detached settings snapshot without updating the authoritative signal (the `on_connect` remember path and the first-run OTT tenant self-heal), which let later signal-based saves clobber the saved endpoint/tenant, regressing #223
- Fixes the same pattern in the `CredentialsUpdated` JWT handler so a minted token can no longer be clobbered off disk
- Tenant self-heal now re-checks inside the write guard to close the peek-then-write window
- Known follow-up: disconnect/logout `auto_connect` inverse-clobber at connector_app.rs ~1055/~1116, same bug family, deferred

## Test plan

- [x] `cargo clippy --workspace --locked --features pentest-platform/desktop-pcap -- -D warnings` passes
- [x] `cargo test --workspace` with the same features passes, 0 failures
- [x] `cargo fmt --check` clean

Refs #374


cchao-devo (changes requested):
## 🔍 Code Review — Strike48-public/pick#462                                                                                                                                     
   **PR:** fix(ui): persist settings through the authoritative signal in remaining detached-clone sites · @jtomek-strike48 · 2 files, +71/−16 · Refs #374 · CI: all 25 green        
                                                                                                                                                                                    
   ---                                                                                                                                                                              
                                                                                                                                                                                    
   ## ✏️ Verdict: REQUEST-CHANGES — core fix is precise and verified; small owed hygiene                                                                                            
                                                                                                                                                                                    
   | Field | Value |                                                                                                                                                                
   |---|---|                                                                                                                                                                        
   | **Action** | `request-changes` — actionable nits the author owes |                                                                                                             
   | **Recommendation** | `nit` |                                                                                                                                                   
   | **Confidence** | 0.77 |                                                                                                                                                        
   | **Human review** | required |                                                                                                                                                  
                                                                                                                                                                                    
   **Why:** The PR implements #374's prescribed remedy verbatim at exactly the two sites the issue names — the on_connect remember path and the OTT tenant self-heal now persist    
 through the authoritative `AppSettings` signal (with the self-heal's peek-then-write window closed under the write guard) — and the `CredentialsUpdated` handler syncs **both**    
 signals so a freshly minted JWT can no longer be clobbered off disk. All three test-plan claims were **independently verified locally at head `0ff5023`**: fmt clean, clippy `-D   
 warnings` clean, `cargo test --workspace` = 1313 passed / 0 failed. What's owed is hygiene, not substance.                                                                         
                                                                                                                                                                                    
   **Owed changes:**                                                                                                                                                                
   1. **Tick the three test-plan checkboxes** — they're `- [ ]` but each claim is true (verified above); also correct the follow-up line numbers to ~1059/~1121.                    
   2. **Log failed `save_settings` writes** instead of `let _ =` swallowing them (connector_app.rs:691/:797, lib.rs:126) — a silent failure means the UI reports                    
 endpoint/tenant/JWT saved while disk holds stale values, masking an auth-loss-on-relaunch. One-line `tracing::warn!` (same ask as the #458 review; precedent in settings.rs).      
   3. **File the disclosed `auto_connect` follow-up — with the full mechanism**: the detached clears at :1060/:1121 can be clobbered back by *any* full-signal rewrite, including   
 the self-heal spawn's own up-to-2-minute post-connect polling loop (unexpected auto-connect on next launch). The stamp-decision extraction + round-trip test (F8) can ride the     
 same issue.                                                                                                                                                                        
                                                                                                                                                                                    
   ---                                                                                                                                                                              
                                                                                                                                                                                    
   ### Foundation checks — 4 pass · 2 nit · 2 n/a                                                                                                                                   
                                                                                                                                                                                    
   1. **Intent & linked issue** — nit — #374's proposed fix implemented verbatim at both named sites; `Refs` (not `Closes`) is defensible since the same-family `auto_connect` path 
 is honestly disclosed as deferred. Gap: unticked test-plan boxes (claims verified true). **Fix:** tick + correct line numbers.                                                     
   2. **Architecture drift** — pass — *reduces* drift: completes the signal migration for the #374 sites; guard-across-save is safe by construction (atomic tmp+rename, no          
 re-entrancy); the self-heal window is closed and documented in-code.                                                                                                               
   3. **Security boundary** — pass — net-positive: pre-fix, the minted JWT left `signals.config.auth_token` stale, so any later `last_config = Some(config.peek().clone())` save    
 (:1345) would overwrite the fresh token on disk — silent auth loss after restart. The fix narrows the persisted write to a token stamp, strictly shrinking the clobber surface.    
   4. **Reimplementation** — pass — reuses `save_settings` + the authoritative signals; matches the `easy_mode`/`matrix_api_url` pattern and the `MatrixTokenObtained` handler.     
   5. **Backend changes** — n/a — `crates/ui` only.                                                                                                                                 
   6. **UI changes** — pass — wiring only; the self-heal's terminal line now fires only on the stamping arm after the guard drops, preserving old semantics.                        
   7. **Screen recording** — n/a — persistence correctness, observable only across restart.                                                                                         
   8. **Tests** — nit — no tests pin the regression invariant; the class already regressed once (#223 → #374) and the still-detached sites are a live trap. **Fix:** extract the    
 self-heal stamp decision into a pure function and unit-test its three arms + one round-trip test (or ride the follow-up issue).                                                    
                                                                                                                                                                                    
   ### Architecture review — 2 pass · 4 n/a                                                                                                                                         
                                                                                                                                                                                    
   - **A1 Module placement** — pass — JWT persistence stays in `run_event_loop`, which owns both signals and already persists `MatrixTokenObtained` identically.                    
   - **A2 Boundary spec drift** — n/a — single-crate; no pub-surface change.                                                                                                        
   - **A3/A4** — n/a — serde JSON settings; no migrations.                                                                                                                          
   - **A5 View/component patterns** — pass — short bounded guard scopes (repo convention :365-367/:1345, session.rs:136-141); the peek at :782 is only a fast path with the real    
 decision re-validated under the guard; the JWT handler introduces **no new dual-signal divergence** (identical token written to both; the `None` branch fully converges); no       
 `.await` between lock acquisitions. The one residual drift channel (self-heal stamps `last_config.tenant_id` without updating `config`) is pre-existing and self-converges.        
   - **A6 Config in code** — n/a.                                                                                                                                                   
                                                                                                                                                                                    
   ### Security review — 3 pass · 3 n/a                                                                                                                                             
                                                                                                                                                                                    
   - **S1 Auth pipeline** — n/a — no auth chain surface; the JWT chain (CredentialsUpdated → config → settings → disk) is now synchronous and in-sync — no path saves an empty or   
 stale token.                                                                                                                                                                       
   - **S2 Network posture** — n/a — no listeners/manifests.                                                                                                                         
   - **S3 Secret handling** — pass — no stale/cleared-token window: both branches write the fresh token argument; the producer emits only non-empty tokens                          
 (liveview_connector/mod.rs:1077-1082); no `.await` interleaving in the single-threaded event loop; the clobber vector is closed; boot-time JWT validation and atomic saves intact. 
 Benign subtlety noted: the `Some(lc)` stamp keeps `lc`'s host/tenant under a token minted for a possibly different endpoint — worst case a loud auth failure, never cross-tenant   
 use. (Pre-existing, out of scope: JWT stored plaintext in settings.json.)                                                                                                          
   - **S4 Injection** — n/a — no sinks; pre-existing local-UUID terminal line only.                                                                                                 
   - **S5 Tenant isolation** — pass — the peek-then-write window is closed (fast-path peek + guarded re-check + `_ => return`); `canonical` comes from the operator's own local     
 credentials file — same trust boundary.                                                                                                                                            
   - **S6 Suspicious content** — pass — pure text Rust.                                                                                                                             
                                                                                                                                                                                    
   ### Adversarial review — 1 pass · 1 nit · 6 n/a                                                                                                                                  
                                                                                                                                                                                    
   - **V1 Input tampering** — n/a — local signals only.                                                                                                                             
   - **V2 Auth bypass** — n/a — JWT persistence correctness verified under S3.                                                                                                      
   - **V3 Injection** — n/a — no sinks.                                                                                                                                             
   - **V4 Race / TOCTOU** — pass — the PR *closes* a peek-then-write window; residual same-family race (auto_connect clobber, including the self-heal spawn's full-signal rewrites) 
 is pre-existing and disclosed → the follow-up.                                                                                                                                     
   - **V5 Data exfiltration** — n/a — nothing leaves the device.                                                                                                                    
   - **V6 DoS / V7 Supply chain** — n/a.                                                                                                                                            
   - **V8 Insider misuse / audit gap** — nit — swallowed persistence failure on auth-state writes (owed change 2).                                                                  
                                                                                                                                                                                    
   ### Summary                                                                                                                                                                      
                                                                                                                                                                                    
   A precise completion of the authoritative-signal migration #374 prescribed, with the self-heal race closed under the guard and the JWT handler syncing both signals; every       
 test-plan claim independently verified (fmt/clippy clean, 1313/0, CI 25/25). Owed: tick the boxes, log persistence failures, and file the `auto_connect` follow-up with the full   
 self-heal-spawn mechanism.

Issue #335, opened 2026-07-27
execute_command logs and exposes injected credentials (tracing args + process table)


jtomek-strike48:
## Problem

Follow-up from the #317 review (differential-authz identity injection). PR #317 (which now carries #319/#314's `execute_command` identity injection) fixed the CRITICAL — the injected credential no longer serializes to the agent via `ProbeCommand.command`. But two **residual HIGH** exposures cchao flagged remain, both in the platform command layer, and both were made secret-bearing by the identity feature:

### 1. Plaintext credential in tracing logs

`crates/platform/src/desktop/command.rs` logs the full argv, which now includes `-H "Authorization: Bearer <secret>"` (or `Cookie:` / `-u user:pass`) on any `identity_ref` run:

- `command.rs:71` (debug): `"Sandbox disabled, executing directly: {} {:?}", cmd, args`
- `command.rs:84` (info): `"[execute_command] cmd={:?} args={:?} full_cmd={:?}"`

Any log sink (journald, Quickwit, file) captures the credential in plaintext. The logging predates #317, but the identity feature makes these lines secret-carrying.

### 2. Plaintext credential in the process table

`-H "Authorization: Bearer <secret>"` passed as argv is visible in `ps` / `/proc/<pid>/cmdline` to any local user for the lifetime of the subprocess. This is inherent to passing headers via argv.

## Impact

On an `identity_ref` run, the operator's test credential is exposed to (a) every log sink and (b) any local process observer — even though the agent-facing wire path is now clean. Directly relevant to the init-dev headless deployment once operators provision `identities.json`.

## Suggested fix

1. **Logs (small):** redact the injected auth flag before the `tracing::info!`/`debug!` at `command.rs:71,84` — reuse `pentest_core::provenance::redact` on `args`/`full_cmd`, or suppress argv logging when an identity was injected. Add a regression test asserting no secret substring appears in the emitted log record.
2. **Process table (design):** document the exposure as an inherent limitation of header-via-argv, and consider `curl -H @file` (header from a file) / `--config` indirection for the supported binaries so the secret never lands in argv. Larger change; can be split.

## Context

- Confirmed leak chain and CRITICAL fix: PR #317 review + commit b0b4768
- The agent-facing serialization leak (`ProbeCommand.command`) is already fixed; this issue is the logging + process-table surface only.
- Feature is inert until the cross-repo keystone (matrix#3354) lands, so there is runway to fix before it goes live.

[...]

PR #463, opened 2026-09-18
fix(security): redact injected credentials from command-layer logs (by value)

jtomek-strike48:
## Summary

- Adds structured by-value redaction of injected identity credentials to all command-layer log sites: host-direct, sandboxed, bwrap inner-command, and SandboxManager::execute
- New redact_arg/redact_args in provenance scrub the exact known secret first, pattern redactor second, non-secret args verbatim
- New execute_command_with_secret trait method threads the known secret through SandboxManager to all four sandbox executors, with a delegating default so mobile behavior is unchanged
- Process-table argv exposure is deferred and documented in-code: the four header flags are argv-only and a naive -H @file workaround would defeat the pick#162 agent-never-sees-secret invariant
- Feature is inert until matrix#3354

## Test plan

- [x] cargo check --workspace clean
- [x] cargo fmt --all -- --check clean
- [x] clippy -D warnings clean on core/platform/tools
- [x] cargo test -p pentest-core -p pentest-platform passes, 0 failures, including 6 new redaction regression tests
- [x] rust-review and security-review both APPROVE

Refs #335


cchao-devo (changes requested):
===BEGIN REVIEW — Strike48-public/pick#463===                                                                                                                                    
   Strike48-public/pick#463 — fix(security): redact injected credentials from command-layer logs (by value)                                                                         
   Author: @jtomek-strike48 · main ← fix/pick/credential-log-leak · 11 files, +373/-41                                                                                              
                                                                                                                                                                                    
   Verdict: request-changes (rec nit, conf 0.86, HR=yes). CI: all 25 green. All claimed tests re-run locally: pentest-core + pentest-platform, 0 failures, redaction filters 25/25. 
                                                                                                                                                                                    
   Composition: 99% bug fix (10 files) · 1% tests (1 file).                                                                                                                         
                                                                                                                                                                                    
   Foundation (8 axes) — 6 pass, 2 n/a. Refs #335; the tracing-args half is closed by value and the process-table residual is documented honestly with the pick#162 constraint in   
 two places; the feature is inert until matrix#3354. The new tests are directional, not tautological: exotic_shape_credential_is_redacted_when_secret_is_known first asserts the    
 precondition that the pattern redactor misses X-Api-Id: ab12cd, then asserts the by-value scrub removes it.                                                                        
   Architecture (A1–A6) — 3 pass, 3 n/a (no schema, migration, or view files). The trait gains a DEFAULTED method so platforms without secret-aware logging keep working unchanged; 
 SandboxManager::execute's new parameter threads to all four backends; execution behaviour is identical — the secret is threaded for logging only.                                  
   Security (S1–S6) — 3 pass, 3 n/a. S3 pass: the injected credential is scrubbed by value from every command-layer tracing line before it reaches persistent log sinks, with       
 known-secret-first ordering immune to header shapes the pattern redactor misses. S4 pass: log redaction strengthened; no new sink.                                                 
   Adversarial (V1–V8) — 2 pass, 1 nit, 5 n/a. V3 pass: the change shrinks the log-injection/leak surface and keeps non-secret arguments discrete and parseable rather than         
 mangling the whole argv.                                                                                                                                                           
                                                                                                                                                                                    
   Findings:                                                                                                                                                                        
   - nit (V5): one adjacent sink the fix does not cover — bwrap.rs:217 logs the child's stdout/stderr VERBATIM at info, and on an identity run curl -v/--verbose reflects the       
 injected Authorization header into stderr, landing in the same journald/Quickwit sinks this PR just cleaned, one line below the fixed site. Out of #335's stated scope (argv logs  
 + process table), but the same leak class.                                                                                                                                         
     Fix: route the captured output through redact_arg/redact with the known secret (or rely on and verify the evidence-layer sanitization for these internal log lines).           
   - nit: the by-value scrub of the ESCAPED command strings is verbatim-substring, and shell_escape (command.rs:398-410) transforms secrets containing ', \ or $ (single-quote      
 wrapping with '\'' splicing). A credential like pass'word therefore no longer appears verbatim inside the shell-escaped full_cmd (or the bwrap inner-cmd string, which is the same 
 escaped form), so redact_arg(&full_cmd, known_secret) misses it and the escaped fragments still spell the secret in the log. The per-argument args={:?} site is safe — redact_args 
 runs on the raw args BEFORE formatting.                                                                                                                                            
     Fix: build the loggable command by shell-escaping the REDACTED arguments (redact -> escape, not escape -> redact), or redact each escaped element individually.                
                                                                                                                                                                                    
   Also verified: a sweep of every tracing site in crates/platform found no remaining unredacted argv/command log lines (the PTY-shell sites log static text only); the 6 new       
 regression tests are covered by the 25-test redaction filter, all green; SandboxManager::execute's signature change updated all four backends and both test call sites; and mobile 
 keeps the delegating default exactly as the body states.                                                                                                                           
                                                                                                                                                                                    
   Summary: Closes the tracing-args half of #335 properly: the known injected secret is threaded from the tool layer through a defaulted trait method and SandboxManager into every 
 command-layer log site, where it is scrubbed by value (immune to the exotic header shapes that defeat the regex redactor) with the pattern redactor as defense in depth and        
 non-secret arguments kept fully visible. The bwrap inner-command site — which predated command.rs's redaction and was missed by it — is fixed too, the process-table residual is   
 documented honestly, and the feature is inert until matrix#3354. Two adjacent items remain: the bwrap executor still logs child stdout/stderr verbatim (a curl -v run reflects the 
 injected header), and the by-value scrub of shell-escaped command strings misses secrets containing quote characters.                                                              
   ===END REVIEW — Strike48-public/pick#463===

Issue #292, opened 2026-07-21
Pending-approval UX: gate Registered on JWT presence + proper callout (was PR #83)


jtomek-strike48:
## Summary

Salvaged from the now-closed draft PR #83. That PR bundled a browser-OAuth
disable (since superseded by the compile-time `browser-auth` cargo feature and
#225/#234) with a genuine, still-open desktop UX bug. This issue tracks only the
still-unmet UX part.

## The bug (verified on `main`)

During post-registration **admin approval** — the connector has connected over
wss but no JWT has been issued yet — the desktop flips to a non-functional
Dashboard instead of staying on the connecting screen.

Root cause, traced on `main`:

- `crates/ui/src/liveview_connector/mod.rs:822` emits
  `ConnectorEvent::StatusChanged(ConnectorStatus::Registered)` **unconditionally
  on a 3-second timer**, gated only on transport being up (`running &&
  ever_connected`) — **not** on a JWT actually being present.
- `crates/ui/src/lib.rs:54` maps `ConnectorStatus::Registered =>
  AppScreen::Connected(...)`, so that timer flips the UI to the connected
  Dashboard even while authorization is still pending.

Net effect: the operator sees an empty/non-functional Dashboard with "pending
approval" buried, rather than a clear "waiting for approval" state.

Secondarily, the approval affordance is minimal: `connecting_screen.rs:86`
renders a flat `.approval-instruction` line + a plain hint ("Awaiting
approval" / "Open the Strike48 web UI and accept this connector.").

## Proposed fix

1. **Gate the `Registered` transition on JWT presence.** Only emit
   `StatusChanged(Registered)` when an auth token actually exists; otherwise
   keep the status `Connecting` with `ConnectingStep::WaitingForApproval` so the
   connecting screen stays visible until approval completes.
2. **Improve the approval callout** on the connecting screen (structured card:
   icon + "Pending approval" title + instruction). Ready-made CSS + markup exist
   on the closed PR #83's branch `fix/desktop-no-auto-browser-oauth`
   (`@719084059`): `connecting_screen.rs` + `css/connecting_screen.css`
   (`.approval-callout*` + `pulse-callout` keyframes). Cherry-pick just those UI
   hunks; drop the auth.rs / config.rs changes (superseded).

## Acceptance criteria

- [ ] With no JWT yet (pending approval), the desktop stays on the connecting
      screen; it does not flip to the Dashboard on the 3s timer.
[...]

PR #464, opened 2026-09-18
fix(ui): gate Registered on JWT presence and show pending-approval callout

jtomek-strike48:
## Summary

- transport-up no longer implies Registered: the status transition is gated on JWT presence via a pure `registration_outcome` decision fn
- while approval is pending, the UI shows a structured pending-approval callout and polls every 2s until the JWT lands, including for a transport that establishes late
- adds a read-only `ConnectorRunner::has_auth_token()` accessor to the vendored strike48-connector fork, documented in PATCHES.md as drop-when-upstream
- callout markup adapted from closed PR #83

## Test plan

- [x] `cargo check`/`test`/`clippy` pass on pentest-ui with `--features connector,shell-ws,desktop`
- [x] `registration_outcome` truth table now asserted for all 8 combinations
- [x] loop bounds verified: shutdown flag, runner-stopped, JWT-issued

Note: a fresh-approval JWT is not persisted to app settings — `last_config.auth_token` stays empty until the next connect; restart relies on the SDK saved-credentials file.

Refs #292


cchao-devo (changes requested):
===BEGIN REVIEW — Strike48-public/pick#464===                                                                                                                                    
   Strike48-public/pick#464 — fix(ui): gate Registered on JWT presence and show pending-approval callout                                                                            
   Author: @jtomek-strike48 · main ← fix/pick/registered-jwt-gate · 5 files, +208/-25                                                                                               
                                                                                                                                                                                    
   Verdict: request-changes (rec nit, conf 0.79, HR=yes). CI: 1 failing / 25 total — Check (macOS), attributed below; NOT this PR's diff.                                           
                                                                                                                                                                                    
   Composition: 62% bug fix (2 files) · 29% UI (2 files) · 9% docs (1 file).                                                                                                        
                                                                                                                                                                                    
   Foundation (8 axes) — 7 pass, 1 nit. Refs #292 and the issue's verified-on-main analysis matches the fix; the decision is factored into a pure registration_outcome function so  
 the choice is unit-testable; the truth table is asserted for all 8 input combinations across 3 tests, re-run locally at this head with the body's exact feature set (--features    
 connector,shell-ws,desktop): 3/3 green.                                                                                                                                            
   Architecture (A1–A6) — 4 pass, 3 n/a (no schema, migration, or view-pattern issues). The vendored-fork modification is documented in PATCHES.md per the fork's convention        
 (problem/change/tracking/drop-when), with an upstream-candidate note.                                                                                                              
   Security (S1–S6) — 1 pass, 5 n/a. The accessor is read-only over the runner's private config — returns a bool, never exposes the token; the UI no longer claims an authorized    
 state without a JWT.                                                                                                                                                               
   Adversarial (V1–V8) — 2 pass, 6 n/a. V2 pass: the change CLOSES a false-authorized display — the desktop previously flipped to the Dashboard on a timer with no JWT, presenting  
 an unusable session as connected; now the authorized claim requires the actual credential.                                                                                         
                                                                                                                                                                                    
   Findings:                                                                                                                                                                        
   - nit: the AwaitingApproval and None (transport-down) branches contain ~20 duplicated lines — identical sleep/shutdown-check/stats-check/JWT-check/emit logic differing only in  
 the tracing::info! text. A future change to the poll cadence, exit conditions, or the JWT check must be made twice, and the two loops can silently diverge.                        
     Fix: factor the bounded wait into one helper (e.g. async fn wait_for_jwt(shutdown, runner) -> bool) called from both arms, or restructure so the initial outcome only picks    
 the step label and a single loop handles the wait.                                                                                                                                 
   - nit: color-mix(in oklch, ...) is the first oklch usage in the codebase (the other 45 color-mix uses are all in srgb). oklch mixing requires a newer webview engine (WebKit     
 16.2+, Chromium 111+); on an older webview the declarations are invalid and the callout renders with no background, border, or warning colour — degrading precisely the attention  
 signal the callout exists to provide.                                                                                                                                              
     Fix: use in srgb like every other color-mix in the crate, unless the wider-gamut blend is deliberate — in which case a fallback declaration ahead of it would keep older       
 webviews styled.                                                                                                                                                                   
                                                                                                                                                                                    
   CI attribution (executed): the Check (macOS) failure is test secure_store::tests::registered_backend_round_trips — the exact #439/#440 global-state flake. Verified three ways:  
 the failing test name matches, pr-464's secure_store.rs contains no TEST_GUARD (pre-fix), and #440 is still OPEN. Landing #440 clears this lane.                                   
                                                                                                                                                                                    
   Also verified: the vendored accessor has no unit test — acceptable for a one-line poll exercised end-to-end by the loop; the poll loop's bounds (shutdown flag, runner-stopped,  
 JWT-issued) are documented at both exit points; and the test build compiles on macOS with the body's feature set (without them it hits the pre-existing cfg-gated breakage also    
 present on main).                                                                                                                                                                  
                                                                                                                                                                                    
   Summary: Closes the #292 UX bug properly: the 3-second timer no longer flips the desktop to a non-functional Dashboard while admin approval is pending. A pure                   
 registration_outcome decision function (full 8-combination truth table, re-run locally — 3/3 green with the body's exact feature set) gates the Registered transition on actual    
 JWT presence, a structured amber callout replaces the bare instruction while approval is pending, and a bounded 2-second poll catches late transports and post-approval JWT        
 issuance. The vendored-fork accessor is a minimal, read-only, documented addition. Two nits: the two poll branches are near-verbatim duplicates, and the callout's color-mix(in    
 oklch, ...) is the codebase's first oklch usage with no fallback. The red Check (macOS) lane is the known #439/#440 flake, not this PR.                                            
   ===END REVIEW — Strike48-public/pick#464===

